Issue #95, opened 2014-07-30
time problems with some events


ghost:
% khal -v
Using configuration:
    calendars:
        0:
            color: 
            name: hths172
            path: /home/mortzu/.vdirsyncer/calendars/hths172/default/
            readonly: False
        1:
            color: 
            name: uni
            path: /home/mortzu/.vdirsyncer/calendars/hths172/uni/
            readonly: False
        2:
            color: 
            name: mw
            path: /home/mortzu/.vdirsyncer/calendars/mw/default/
            readonly: False
        3:
            color: 
            name: planetcyborg
            path: /home/mortzu/.vdirsyncer/calendars/planetcyborg/default/
            readonly: False
        4:
            color: 
            name: vertraege
            path: /home/mortzu/.vdirsyncer/calendars/planetcyborg/vertraege/
            readonly: False
        5:
            color: dark blue
            name: privat
            path: /home/mortzu/.vdirsyncer/calendars/mortzu/default/
            readonly: False
    default:
        debug: False
        default_calendar: privat
        default_command: calendar
    locale:
        dateformat: %d.%m.
        datetimeformat: %d.%m. %H:%M
        default_timezone: Europe/Berlin
        encoding: utf-8
        firstweekday: 0
        local_timezone: Europe/Berlin
        longdateformat: %d.%m.%Y
        longdatetimeformat: %d.%m.%Y %H:%M
        timeformat: %H:%M
        unicode_symbols: True
    sqlite:
        path: /home/mortzu/.config/khal/khal.db
created version table
created accounts table
calculating recurrence dates for abe304dc-f7b1-4f2b-997d-5c8bcfaa7e0b.ics, this might take some time.
Failed to parse vcard abe304dc-f7b1-4f2b-997d-5c8bcfaa7e0b.ics during update_vevent in collection hths172
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/khal/khalendar/khalendar.py", line 183, in update_vevent
    ignore_invalid_items=True)
  File "/usr/lib/python2.7/site-packages/khal/khalendar/backend.py", line 297, in update
    href)
  File "/usr/lib/python2.7/site-packages/khal/khalendar/datetimehelper.py", line 103, in expand
    dtstartl = [start for start in dtstartl if start not in exdates]
TypeError: can't compare offset-naive and offset-aware datetimes

created version table
created accounts table
[...]

Issue #68, opened 2014-05-05
Better support for recurrent events


geier:
- [x] EXDATE
- [x] RDATE with DATE and DATE-TIME values
- [ ] RDATE with PERIOD
- [ ] RECURRENCE-ID (see #95 for example)
- [ ] RECURRENCE-ID;RANGE=THISANDFUTURE:19980401T133000Z

untitaker:
This is mostly done, the rest is documented in https://github.com/geier/khal/issues/152

whodevil:
Awesome.

geier:
@whodevil please report any error you might encounter

PR #147, opened 2015-01-26
[WIP] new db layout, better recurring events support

### LINKED ISSUE #118: new database layout
was `blacklist VTODOs in khal's db`

```
 khal tries to parse them every time the vdir updates, which is kind of unperformant. Could we blacklist these items in the DB until they actually update?
```

I'm currently rethinking the db layout to deal better with complex recurrence patterns (see branch for the current iteration `rrrule_exceptions`), this won't happen before that is done.

geier:
this is mainly a todo list:

- [x]  support empty ics (i.e. no vevents inside, but VTODOs)
- [x] test for empty ics
- [x] EXDATE
- [x] RDATE with DATE and DATE-TIME values
- [x] deal with RDATE with PERIOD
    -  implementation currently not possible, [need to support this in icalendar first](https://github.com/collective/icalendar/issues/156)
  - [x] print warning instead
- [x] RECURRENCE-ID (see #95 for example)
- [x] RECURRENCE-ID;RANGE=THISANDFUTURE:19980401T133000Z
- [x] explicit non-suppot for RECURRENCE-ID;RANGE=THISANDPRIOR
- [x] support for events with same uid in different calendars (while not needed for standard conformity this is probably nice to have)
- [x] check if more than one RANGE=THISANDFUTURE is allowed and if so implement, otherwise at least warn if this happes

once this is ready it should close #118 and  #68


untitaker:
I pushed a commit that skips the locale test if the DE-locales are not generated. Just FYI, so please don't ``push -f`` :)

tests/backend_test.py
```diff
@@ -38,4 +52,155 @@ def test_new_db_version():
...
+def test_event_rrule_this_and_future():
+    dbi = backend.SQLiteDb('home', ':memory:', locale=locale)
+    dbi.update(event_rrule_this_and_future, href='12345.ics', etag='abcd')
+    assert dbi.list() == [('12345.ics', 'abcd')]
+    events = dbi.get_time_range(datetime(2014, 4, 30, 0, 0), datetime(2014, 9, 26, 0, 0))
+    events.sort(key=lambda x: x.start)
+    assert len(events) == 6
+
+    assert events[0].start == berlin.localize(datetime(2014, 6, 30, 7, 0))
+    assert events[1].start == berlin.localize(datetime(2014, 7, 7, 9, 0))
```
untitaker: This currently fails. Instead of the expected value, it gives me:

    berlin.localize(datetime(2014, 7, 7, 7, 0))

untitaker: None of the recurring events should start at 9AM?

geier: this is not implemented yet...

geier:
thanks for properly fixing that locale test

tests/backend_test.py
untitaker: I don't understand... the output of the tested code looks correct to me. Note that I am not well-versed with dates and times though.

geier:
```RECURRENCE-ID;RANGE=THISANDFUTURE:20140707T050000Z
DTSTART;TZID=Europe/Berlin:20140707T090000```

means all events after 20140707T050000Z should start at 9:00.

untitaker: Oh, I missed those lines... That's horrifying to see, that the order of DTSTART/RECURRENCE-ID items seems to matter?

geier: no, the order doesn't matter, but all events with the same UID are updated. Actually the order can be arbitrary anywhere (that's why the events are not directly inserted into the db from the icalendar object, but sorted before).

untitaker: Screw me... I oversaw that these are two separate events.

untitaker:
BTW how does this behave if there is a VTODO? Does it save the VTODO in the db, or create a stub entry for it?

geier:
atm VTODOS **should** be inserted into the `events` table, but not in
recurring date(time) tables. This could of course be changed to include
e.g. their due date or reminder time.

Quoting Markus Unterwaditzer (2015-01-29 12:17:40)

untitaker:
I added a new testcase and changed some logging messages. The new behavior is:

* Warn if a file can't be parsed
* Just skip silently if no vevent could be found

PR #161, opened 2015-02-15
Programatically enforce stylechecking

untitaker:


geier:
I'm fine with this (when it works)

PR #168, opened 2015-02-21
Strictly separate new and update

untitaker:


untitaker:
I'd like to remove the ``href`` and ``etag`` properties from Event... I think there should be a differentiation between storage-specific information for an event and the event data itself. See e.g. ``change_collection`` for a place where things get ugly.

geier:
Do you care to elaborate? Where would we store the storage-specific information than?

Also, is this ready to merge?

untitaker:
Storing stuff like the calendar on the event is basically a duplication of data. E.g. at the moment ``event.calendar`` isn't being updated inside ``change_collection`` when it totally should be.

Yeah, this is ready to merge, the comment was unrelated. This PR addresses something different: There is no explicit distinction between uploading and updating (unlike in vdirsyncer), while I do feel that's important.

untitaker:
But I can't give objective reasons for this... I suppose it's kind of a gut-feeling thing. I fear that bugs are bound to happen, or at least more likely, if we don't make that distinction.

untitaker:
Perhaps it's not so much where this data is stored, but actually the fact that Event is a mutable object.

Issue #146, opened 2014-12-27
Creation of VTIMEZONE components from cache


geier:
Since creation of VTIMEZONE components from pyzt timezones will always result in annoying RDATE lists (becaus we don't know the RRULE), we should cache at least the commonly used timezone RRULES.

A possible solution would be to use the VTIMEZONEs available from http://tzurl.org/zoneinfo/ . Ideas to make them available to khal:
- downloaded at package building time and included in the distribution
- downloading with a new command from khal (and warning when they are not yet downloaded)

geier:
We could use [vzic](https://github.com/libical/vzic/) to create vtimezones. Probably best to package them as a separate package which we can update frequently.

PR #183, opened 2015-03-30
Fix tests

untitaker:


untitaker:
@geier Please verify that the tests are now not too liberal.

geier:
I think the tests are too loose now, we can produce pretty broken VTIMEZONE components without failing the tests now. 

I was thinking about giving freezegun a try (just learned about it this weekend).

untitaker:
>I think the tests are too loose now, we can produce pretty broken VTIMEZONE components without failing the tests now.

Which ones, for example?

Syntactically invalid properties are unlikely to occur because icalendar takes care of that for us. The order of the properties doesn't matter, too, and it's also unlikely that we're putting a property into the wrong kind of component.

We could also parse either of them and compare their contents.

geier:
The way it is now we cannot detect if standard or DST time is +0100 or +0200 (initially this was in the wrong order and we would not detect it with these tests)

geier:
perhaps the best solution would be to use cached vtimezone components #146

untitaker:
@geier Please review.

geier:
I'm much happier with this, please merge if it does indeed work.

untitaker:
Done.

PR #185, opened 2015-04-21
Add a line about khal being in Fedora 22 onwards

mbaldessari:
khal will be present as an rpm in Fedora 22 and later.
Make a note of this in the install docs.


untitaker:
Thanks!

PR #186, opened 2015-05-01
Show Organizer & Fix BST

bradleyjones:
If Organizer is present on an event show it, stripping out mailto: in the
string. Also move location before description as often descriptions are much
longer than the location.

Added check for BST for when timezone reports British Summer Time not DST


untitaker:
Seems straightforward, thanks!

geier:
:frowning:  I knew that code was somewhat brittle (relying on `__repr__()` doesn't sound like the best idea ever), but all timezones I had tested seemed to work. I really should find the time to create timezones from some kind of cache (#146).

PR #187, opened 2015-05-02
python3 compatibility

hobarrera:
~~**Do not merge yet**~~ Update: This PR is ready for merging. See below for further details.

This is a work-in-progress PR, mostly to avoid anybody else duplicating the effort, and to get some preliminary feedback.

`khal` and `ikhal` work flawlessly. There are still nine tests failing, *however*, I'm inclined to think that it's due to how the tests are written rather than the code itself.

There's two types of failing tests with the tests right now:

 * Comparisons where the generated UID does not match the expected one. I haven't looked deeply into this yet.
 * Comparisons where python3 seems to be including shell escape code for bold, hence, outputs do not match: '\x1b[1mNo events\x1b[0m\n' == 'No events\n'. Fixing the test would break the python2 version, so I'll have to think of a better solution later on.

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -21,6 +21,8 @@
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
 
+from __future__ import unicode_literals
+
 from click import echo
 
 import datetime
@@ -50,9 +52,9 @@ def construct_daynames(daylist, longdateformat):
     """
     for date in daylist:
         if date == datetime.date.today():
-            yield (date, u'Today:')
+            yield (date, 'Today:')
         elif date == datetime.date.today() + datetime.timedelta(days=1):
-            yield (date, u'Tomorrow:')
+            yield (date, 'Tomorrow:')
         else:
             yield (date, date.strftime(longdateformat))
 
@@ -186,5 +188,5 @@ class Interactive(object):
                               description='do something')
         ui.start_pane(
             pane, pane.cleanup,
-            program_info=u'{0} v{1}'.format(__productname__, __version__)
+            program_info='{0} v{1}'.format(__productname__, __version__)
         )
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -32,6 +32,7 @@ import sys
 import textwrap
 
 from khal import aux, calendar_display
+from khal.compat import to_unicode
 from khal.khalendar.exceptions import ReadOnlyCalendarError
 from khal.exceptions import FatalError
 from khal.khalendar.event import Event
@@ -146,7 +147,7 @@ class Agenda(object):
         term_width, _ = get_terminal_size()
         event_column = get_agenda(collection, dates=date, width=term_width,
                                   show_all_days=show_all_days, **kwargs)
-        echo('\n'.join(event_column).encode(encoding))
+        echo(to_unicode('\n'.join(event_column), encoding))
 
 
 class NewFromString(object):
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -137,6 +137,8 @@ class Calendar(object):
             firstweekday=firstweekday, weeknumber=weeknumber)
 
         rows = merge_columns(calendar_column, event_column)
+        # XXX: Generate this as a unicode in the first place, rather than
+        # casting it.
         echo('\n'.join(rows).encode(encoding))
 
 
@@ -147,6 +149,8 @@ class Agenda(object):
         term_width, _ = get_terminal_size()
         event_column = get_agenda(collection, dates=date, width=term_width,
                                   show_all_days=show_all_days, **kwargs)
+        # XXX: Generate this as a unicode in the first place, rather than
+        # casting it.
         echo(to_unicode('\n'.join(event_column), encoding))
 
 
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -23,7 +23,7 @@
 
 from __future__ import unicode_literals
 
-from click import echo
+from click import echo, style
 
 import datetime
 import itertools
@@ -37,7 +37,7 @@ from khal.exceptions import FatalError
 from khal.khalendar.event import Event
 from khal import __version__, __productname__
 from khal.log import logger
-from .terminal import bstring, colored, get_terminal_size, merge_columns
+from .terminal import colored, get_terminal_size, merge_columns
 
 
 def construct_daynames(daylist, longdateformat):
@@ -111,14 +111,14 @@ def get_agenda(collection, locale, dates=None,
         if len(events) == 0 and len(all_day_events) == 0 and not show_all_days:
             continue
 
-        event_column.append(bstring(dayname))
+        event_column.append(style(dayname, bold=True))
         events.sort(key=lambda e: e.start)
         for event in itertools.chain(all_day_events, events):
             desc = textwrap.wrap(event.compact(day), width)
             event_column.extend([colored(d, event.color) for d in desc])
 
     if event_column == []:
-        event_column = [bstring('No events')]
+        event_column = [style('No events', bold=True)]
     return event_column
 
 
@@ -172,10 +172,10 @@ class NewFromString(object):
                          'read-only'.format(collection.default_calendar_name))
             sys.exit(1)
         if conf['default']['print_new'] == 'event':
-            print(event.long())
+            echo(event.long())
         elif conf['default']['print_new'] == 'path':
             path = collection._calnames[event.calendar].path + event.href
-            print(path.encode(conf['locale']['encoding']))
+            echo(path.encode(conf['locale']['encoding']))
 
 
 class Interactive(object):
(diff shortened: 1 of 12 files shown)


hobarrera:
I've found a workaround for the second type of failing tests. See @8cb6fe8 for details. It's not the best solution possible, but it's the best thing I can do while keeping python2 and python3 supported at the same level.

hobarrera:
Right, I found the issue for the remaining test cases.

The vevents with the which the test outputs are compared have UIDs defined. These UIDs are the ones generated by python2 after seeing `random` with `1`, hence, they match the autogenerated ones.

python3's randomness algorithm differs from python2's one, so these tests fail. I've updated the tests to replace the value of the events with a pre-defined one, since UIDs generation isn't what we indend to test.

This fixes the last test when running with python3.

khal/controllers.py
```diff
@@ -21,6 +21,8 @@
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #
 
+from __future__ import unicode_literals
+
```
untitaker: **Don't do this!** Some Python 2 APIs exhibit weird behavior when fed unicode strings.

tests/cli_test.py
```diff
@@ -90,7 +97,7 @@ def test_simple(runner):
 
     result = runner.invoke(main_khal)
     assert not result.exception
-    assert result.output == 'No events\n'
+    assert strip_ansi_escapes(result.output) == 'No events\n'
```
untitaker: Shouldn't be necessary, Click does this for us.

untitaker: FWIW, Click works flawlessly in that sense for vdirsyncer.

tox.ini
```diff
@@ -1,10 +1,12 @@
 [tox]
-envlist = py27, pypy
+envlist = py27, pypy, py34
 
 [testenv]
 deps = 
     pytest
     pytest-capturelog
+    tzlocal
+    configobj
```
untitaker: Shouldn't be necessary, `khal` gets installed as pkg (including dependencies as specified in `setup.py`)

tests/cli_test.py
hobarrera: I mentioned this in the PR. Without this, the output mismatches.

Remove that function call, and run `pytest -vv`. You'll noticed that the compared string on the left includes the escape codes.

Maybe you have some setting on vdirsyncer for it to do this for you? If so, I'd very much prefer to use that rather than this hack.

khal/controllers.py
hobarrera: I wasn't aware of this. :(

Have you found any in khal's case? Some brief testing proved that everything worked, and tests pass, but I'm willing to look into this on a case-base-case basis.

tox.ini
hobarrera: Right, but you might want to run the tests *before* installing (which, for example, is the case when developing or packaging in a clean environment).

untitaker: I'm not sure what you mean -- tox installs the current pkg into the created virtualenv by default, AFAIK

khal/controllers.py
untitaker: Probably khal doesn't even use any APIs where this matters, I just wouldn't do it as a general rule.

tests/cli_test.py
untitaker: This is click's default behavior. I'll test this soonish.

tox.ini
hobarrera: You're right, I seem to have misunderstood something during my initial testing. I'll remove these.

khal/controllers.py
hobarrera: Great. So let's keep this around in khal's case. 😉

tests/cli_test.py
untitaker: Ahh, this is because Khal doesn't use `click.echo`, which takes care of this.

hobarrera: Please do. Maybe there was something odd in my environment, but merely redefining `strip_ansi_escapes` to `return value` will show this issue, though *only on python 3!*

untitaker: It is -- it strips colorcodes from the given string if its running under click's test runner.

khal/controllers.py
untitaker: I really wouldn't do it... to argue my points, I've found [this](http://python-future.org/unicode_literals.html#drawbacks), but it doesn't touch all of the issues we could experience.

untitaker: Maybe from a readability POV (although a very weak point IMO): Because only some modules have that future-import, you can't know whether a particular string literal produces a unicode or bytestring. So we should then do this import for *every* module, but that means we'd have to wrap 70% of string literals with `str(...)` (because "native strings" is what most stdlib APIs actually expect)

tests/cli_test.py
hobarrera: It seems to remove colour codes, but not bold codes (and it's bold ANSI escapes we're getting in this case).

untitaker: I think that's a bug in Click.

khal/controllers.py
hobarrera: The linked website only mentions a few issues in detail, and I don't see them applying to us.

> Because only some modules have that future-import, you can't know whether a particular string literal produces a unicode or bytestring.

Without this import, literals are `str`, *not* bytestrings. With it, they become unicode strings (`unicode`). And stdlib still works with unicode strings, eg:

    >>> with open(u'test', u'w') as f:
    ...   f.write("1")

It's not merely readability either, IMHO. When writing new code, you don't need to remember to add adding `u'` to new strings - and forgetting that may cause subtle unexpected error.

Personally, I don't see drawbacks with using `unicode_literals` on khal's codebase, nor any way this can come back and bite us in a forseeable future, so I'd much rather keep it.

tests/cli_test.py
hobarrera: I'm not 100% sure. The doc doesn't say that bold escape codes are removed, but it *is* expected, so I'd still report this.

The problem is the inconsistency between python2 and python3, which *is* a [bug](mitsuhiko/click#334) IMHO.

khal/controllers.py
untitaker: >Without this import, literals are str, not bytestrings. With it, they become unicode strings (unicode).

Somehow the rest of the sentence slipped: You first have to look at the top of the file whether `unicode_literals` are imported to make sense of the literals, which is IMO dangerous while reviewing PRs.

>and forgetting that may cause subtle unexpected error

The opposite does too. Try the following code, with `file` having nonascii characters in it:

    with open('file') as f:
        print(u'output: {}'.format(f.read()))

untitaker: So basically my point is that changing literals everywhere (by importing `unicode_literals`) might create regressions for current Python 2 users, which are arguably the majority right now.

tests/cli_test.py
untitaker: I think you meant to link https://github.com/mitsuhiko/click/issues/334

hobarrera: Oops. Yes.

untitaker: Hmm, ISTM we should switch to unicode strings.

khal/controllers.py
```diff
@@ -170,10 +172,10 @@ def __init__(self, collection, conf, date_list, location=None, repeat=None):
                          'read-only'.format(collection.default_calendar_name))
             sys.exit(1)
         if conf['default']['print_new'] == 'event':
-            print(event.long())
+            echo(event.long())
         elif conf['default']['print_new'] == 'path':
             path = collection._calnames[event.calendar].path + event.href
-            print(path.encode(conf['locale']['encoding']))
+            echo(path.encode(conf['locale']['encoding']))
```
untitaker: This is always bytes.

hobarrera:
I can squash some commits pre-merge if you like since one or two ended up being reversed (I avoided doing so thus far because it'd break the comments on existing commits).

untitaker:
I'm still really averse to the whole `unicode_literals` thing.

Regarding squashing, I think GitHub preserves the diff comments I've made, at least on this PR view.

hobarrera:
> I'm still really averse to the whole `unicode_literals` thing.

I don't see any downsides to it in our case.

Scenarios like wsgi or other network-oriented IO might have issues, but khal doesn't do this, and most likely never will.  
It also makes sure new code behaves consistently between python2/python3 in case any developer forgets about `u'` (which is rather common).

untitaker:
>It also makes sure new code behaves consistently between python2/python3 in case any developer forgets about u' (which is rather common).

I'd rather say that it tries to conceal the differences badly.

I feel like we could do this in a separate PR where we would switch Khal's internal APIs and representations to unicode strings, but personally I am not really happy with it. @geier might have different opinions though.

hobarrera:
Rebased and squashed a few commits. Thanks a lot for all the feedback, @untitaker!

untitaker:
Awesome, thanks @hobarrera!

geier:
Thanks hobarrera!

I've added you as a contributor, but I'd appreciate it, if you would keep making pull requests for anything bigger.

hobarrera:
Thanks for that. :)
Yes, I'll still make PRs so someone else reviews the changes before merging.

Cheers!

Issue #111, opened 2014-09-17
import .ics files


geier:
import complete calendar .ics files

this could also be implemented in vdirsyncer, what do you think @untitaker ?

blocked by #118

untitaker:
I think maybe such things should be implemented in a ``vdir-utils`` package, which would be a set of standalone CLI tools.

For this particular feature, a command with an interface similar to coreutils' ``split`` could be provided:

    vobj-split my-file.ics ~/.calendars/default/

The main advantage here is that it is neither specific to khal, vdirsyncer or even vdir, the interface is easy to use, regardless of whether you already use khal/vdirsyncer or not.

geier:
When implementing (parts of) RFC 5546 this will be more or less a by
product (and I think RFC 5546 is a better fit for khal, because one
probably wants to view the event(s) and the event's date before
accepting or declining.

Since implementing RFC 5546 is a lot of work, it probably makes sense to
implement this independently now.

Quoting Markus Unterwaditzer (2015-02-02 14:36:54)

geier:
this will hopefully come together with the new database layout, I'm
working on that in `rrrule_exceptions` (sadly not enough time atm)

Quoting James Mills (2014-11-26 16:19:16)

untitaker:
@geier It seems that your replies appear to be sent very delayed, also in #130 and #131.

geier:
ran msmtp-runqueue on a seldomly used laptop...

bepolymathe:
Hi... any news about this. Pipe file .ics like this don't work
`
| khal import
`
Any idea ?

geier:
@bepolymathe please open a new issue for this, thank you!

PR #202, opened 2015-05-25
Importing ICS files

libre-man:
I made a fix to import ics  files with one or more VEVENT. Everything seems alright. I have not made tests because I didn't have the time just yet.

The fix is working for me, but I didn't test with strange data.

edit: this would solve #111

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,9 +17,9 @@ Features
   example CalDAV_ servers.
 - fast and easy way to add new events
 - ikhal (interactive khal) lets you browse and edit calendars and events
-- support for recurring events is not complete yet, they cannot be deleted or
-  edited and some recursion patterns are not understood
+- only rudimentary support for creating and editing recursion rules
 - you cannot edit the timezones of events
+- works with python 2.7 and 3.4+
 - khal should run on all major operating systems [1]_
 
 .. [1] except for Microsoft Windows
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -52,7 +52,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'khal'
-copyright = u'2014, Christan Geier et al.'
+copyright = u'2014-2015, Christan Geier et al.'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
@@ -108,12 +108,16 @@ todo_include_todos = True
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
+html_theme = 'alabaster'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-#html_theme_options = {}
+html_theme_options = {
+    'github_user': 'geier',
+    'github_repo': 'khal',
+    'travis_button': 'true',
+}
 
 # Add any paths that contain custom themes here, relative to this directory.
 #html_theme_path = []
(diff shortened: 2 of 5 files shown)


khal/aux.py
```diff
@@ -139,6 +139,20 @@ def generate_random_uid():
     choice = string.ascii_uppercase + string.digits
     return ''.join([random.choice(choice) for _ in range(36)])
 
+def construct_event_list_from_string(ics_as_string):
+    """Construct a list with icalendar.event types with a random UID"""
+    try:
+        events = icalendar.Calendar.from_ical(ics_as_string)
+    except ValueError:
+        logger.fatal("Please specify a ics file")
```
untitaker: I don't think it's a good idea to catch all ValueErrors here. In addition it makes sense to reprint the error message to show the user what's wrong.

khal/controllers.py
```diff
@@ -154,6 +154,32 @@ def __init__(self, collection, date=None, firstweekday=0, encoding='utf-8',
         echo(to_unicode('\n'.join(event_column), encoding))
 
 
+class NewFromFile(object):
```
untitaker: @geier Why are the controllers not functions?

khal/aux.py
libre-man: I don't agree on the <code>ValueError</code>. The only way a <code>ValueError</code> could occur here is if the <code>ics_as_string</code> is in fact not a ics_as_string.
Printing the error might be useful with the verbose flag, but without I think it is not useful to the user.

Furthermore this is an error that might occur in normal use. Sometimes people select the wrong file and sometimes files are not correct. So throwing an Error is, in my opinion, too verbose.

untitaker: That exception may be raised if there is only a *slightly* malformed ICS file passed.

libre-man: But how would you recognize and/or correct a slightly malformed ICS file? And wouldn't that be the job of the icalendar module?

libre-man: I checked it with an ics file that had <code>END:VEVEN</code> instead of <code>END:VEVENT</code> and icalendar imported it without throwing any errors.

geier:
Hi libre-man and welcome on board! 
While your code will probably work in >90% of cases, it will lead to unexpected behavior in some cases (because the icalendar standard allows dome crazy stuff and is pretty lax in some regards). I don't have the time to explain the details at the moment, but probably tonight. 

@untitaker historic reasons, look at my import branch

libre-man:
Ok I'm looking forward to your reply. But after looking at your import branch (should have looked for that earlier...) I'm think I understand some of the difficulties. 

Somethings that are not working with this is recurring events. They show up in khal and on my phone but not in ikhal.

geier:
ok, this took a bit longer than I hoped for, but here is a short explanation of the problems with your code:

A single `.ics` can contain several VEVENTS.
While one would expect every VEVENT to have its own unique UID (for what it's
worth they are named *unique identifier*), there is a case where several
VEVENTS have the same UID, but do describe the same (recurring) event.  In
this case, one VEVENT, containing an RRULE or RDATE element would be the
*proto* event, from which all recurrence instances are derived.  All other
VEVENTS with the same UID would then have a RECURRENCE-ID element (I'll call
them *child* event from now on) and describe deviations of at least one
recurrence instance (RECURRENCE-ID elements can also have the added property
RANGE=THISANDFUTURE, meaning the deviations described by this child event also
apply to all further recurrence instances.

Because it is possible that an event already in the database consists of a
master event and at least one child event gets updated and than consists only
of a master event, we currenty *delete* all events with the same UID from the
database when inserting or updating a new event. But this means that we need
to update an event always at once (master and all child events) at the same
time (using `Calendar.update()` or `Calendar.new()` in this case)

As this wouldn't be bad enough, the standard looses no words on the ordering
on those VEVENTS in any given `.ics` file (at least I didn't find any). Not
only can the proto event be *behind* any or all RECURRENCE-ID events, but also
events with different UIDs can be in between.

We therefore currently first collect all events with the same UID and than
sort those by their type (proto or child), and the childs by the value of the
RECURRENCE-ID property.

I think I'll add it to the documentation as well.

If you want to contribute but need some assistance, please do ask.

PR #205, opened 2015-06-02
Add space and backspace as aliases for keybindings

pdav:
As with vi, space is an alias for 'k' and backspace is an alias for 'h'.


untitaker:
Would you actually use this alias?

pdav:
To answer you, I intuitively tried space and backspace before using l and k (and I always use space and backspace in vi). I first configured them in my private config file, but I then thought other may benefit from these key bindings.

geier:
Sounds reasonable. Thanks!

PR #209, opened 2015-06-18
Remove encodings

untitaker:
The test about `khal calendar`'s output fails because Click doesn't
strip colorcodes if the string is a bytestring.

https://github.com/mitsuhiko/click/issues/334

I don't see a need for a custom encoding since this should be inferrable
from the environment, so I removed it. We probably should check if that
config key is necessary at all.


hobarrera:
Less code and one test fixed? YAY! 😄

PR #211, opened 2015-06-22
Add zsh autocompletion for `khal import`.

hobarrera:
This is the first time I've written a zsh completion, so feedback welcome (though the changes are extremely simple, so I don't expect anything to be off).

### DIFF
--- a/misc/__khal
+++ b/misc/__khal
@@ -10,6 +10,9 @@ _calendars() {
   compadd "$a" ${(kv)=accounts}
 }
 
+_ics_files() {
+  _files -g "*.ics"
+}
 
 local ret=1 state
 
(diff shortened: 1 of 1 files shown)


untitaker:
Click is able to generate bash completion automatically. I think we should replace all of this sometime, and if necessary, fix click to support zsh.

hobarrera:
I don't think we're including the bash completions (or something is wrong with my setup), so if click can generate them automatically, it's worth including.

+1 to click-zsh-generation. :)

geier:
thanks!

completion-pattern generation for zsh would be awesome, but I actually don't know enough about that to give it try (and I found it pretty hard to find some comprehensive and comprehensive documentation on the matter, but I may not have looked too thoroughly).

PR #214, opened 2015-06-24
let user choose calendar on import

### LINKED ISSUE #212: Interactive calendar selection when importing
I tried configuring mutt to import ICS files (eg: event invitations) by using "khal import", but I realized that I need to pass the target calendar as an argument.

Interactive calendar selection would be useful. Or maybe, if no calendar is selected, ask for each one until there's a yes:

```
Import ___ into calendar1? [y/N] N
Import ___ into calendar2? [y/N] y
```

geier:
this should close #212

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -99,6 +99,22 @@ containing the first given date. If today is included, it is highlighted.
 Have a look at ``khal agenda`` for a descrption of the options.
 
 
+import
+******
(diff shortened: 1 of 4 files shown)


hobarrera:
Yup, this works great for usage via mutt. :D
In case anybody's interested:

    $ grep khal .config/mutt/mailcap 
    application/ics; khal import %s;

Thanks!

Issue #215, opened 2015-07-04
Adding a new event results in error


nielsk:
I try to add a new event like this:
khal new -a work 05.07.2015 16:00 18:00 TestEvent

Then I get the following error:
```
Traceback (most recent call last):
  File "/usr/bin/khal", line 11, in <module>
    sys.exit(main_khal())
  File "/usr/lib/python3.4/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.4/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.4/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.4/site-packages/khal/cli.py", line 244, in new
    until=until.split(' ') if until is not None else None,
  File "/usr/lib/python3.4/site-packages/khal/controllers.py", line 170, in __init__
    collection.default_calendar_name,
  File "/usr/lib/python3.4/site-packages/khal/khalendar/khalendar.py", line 248, in default_calendar_name
    return names[0]
TypeError: 'dict_keys' object does not support indexing
```

geier:
I have pushed a commit that might fix this. The only way I could
reproduce this though, was by setting all calendars to read only. If
this is indeed the case for you, adding a new event will still fail,
albeit hopefully with a sensible error message.

Quoting Niels Kobschätzki (2015-07-04 06:21:44)

untitaker:
I don't think automatically choosing a random calendar is safe or desirable behavior.

geier:
This is the way it's always been, I just documented that behavior.  Happy to accept pull requests.

(This is only really relevant when adding a new event with `new`,  but we might want to ask the user) 

Am 4. Juli 2015 14:46:23 MESZ, schrieb Markus Unterwaditzer <notifications@github.com>:

nielsk:
Ok, thanks. And every calendar was set to read_only. My mistake. Sorry, I hope this wasn't too much work.

geier:
Don't worry, it was still a bug 
Quoting Niels Kobschätzki (2015-07-04 16:05:18)

PR #218, opened 2015-07-04
Khal doesn't pick arbitrary calendars as default

untitaker:
See #215

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -52,7 +52,7 @@ def time_args(f):
     return dates_arg(events_option(days_option(f)))
 
 
-def _calendar_select_callback(ctx, option, calendars):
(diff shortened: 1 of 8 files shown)


geier:
I'm not sure if `-c` is the best choice for the new parameter, I actually prefer it, if `-c` is only used as short form of `--config` (this is why I initially chose `-a` for **a**ctivating calendars, the only suggestion I have is also using `-a`, but I guess that is not ideal either.

Anyone else want to weigh in?

untitaker:
I'll just rename it to `-a`. Come to think of it I think consistency with other options is more important.

untitaker:
Done!

geier:
Thanks and merged!

Issue #222, opened 2015-07-12
ikhal crashes when translating event TZs across date boundaries


plega:
First, congratulations on khal, a great application.

Now, the bug.  I created the following event:

BEGIN:VCALENDAR
PRODID:-//Mozilla.org/NONSGML Mozilla Calendar V1.1//EN
VERSION:2.0
BEGIN:VTIMEZONE
TZID:Europe/London
BEGIN:DAYLIGHT
TZOFFSETFROM:+0000
TZOFFSETTO:+0100
TZNAME:BST
DTSTART:19700329T010000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
END:DAYLIGHT
BEGIN:STANDARD
TZOFFSETFROM:+0100
TZOFFSETTO:+0000
TZNAME:GMT
DTSTART:19701025T020000
RRULE:FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU
END:STANDARD
END:VTIMEZONE
BEGIN:VEVENT
CREATED:20150701T095323Z
LAST-MODIFIED:20150706T105425Z
DTSTAMP:20150706T105425Z
UID:5507-5593B880-1-15FFE820
SUMMARY:airport transfer :: LHR t5 - home [plega]
ORGANIZER;CN=[...]com
ATTENDEE;PARTSTAT=ACCEPTED;RECEIVED-SEQUENCE=1;ROLE=REQ-PARTICIPANT;RECEIVE
 D-DTSTAMP=20150701T095323Z;CN=plega:mailto:[...]
DTSTART;TZID=Europe/London:20150710T180500
DTEND;TZID=Europe/London:20150710T200000
LOCATION:[...]
CLASS:PUBLIC
TRANSP:OPAQUE
SEQUENCE:1
X-MOZ-GENERATION:1
END:VEVENT
END:VCALENDAR

This displays correctly in ikhal when "local_timezone = Europe/London".  However, if I set "local_timezone = Australia/Sydney" it crashes.  From testing it, I believe (though I haven't tested it exhaustively) that this will happen with any value for local_timezone which results in the above event being split across two days.

Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/khal-0.5.0-py2.7.egg/khal/ui/__init__.py", line 653, in start_pane
    loop.run()
  File "/usr/lib64/python2.7/site-packages/urwid/main_loop.py", line 273, in run
    self.screen.run_wrapper(self._run)
[...]

Issue #195, opened 2015-05-09
Events that end on the same local date, but different GMT date are displayed oddly


hobarrera:
```
$ khal search DAC
08.05.2015 10:30 - 08.05.2015 22:00: DAC Location:  Description:
15:30-17:30 2015-05-07: DAC Location:  Description:
...
```

As you can see by this output, the first event begins and ends on the same day, but it's displayed as a multiday event.

This is most likely because it ends on the next day, when working with GMT times (I'm UTC-0300).

geier:
@hobarrera does this still apply?

hobarrera:
Nope, at a quick glance, this is no longer an issue.

geier:
Thanks, for reporting and checking.

PR #223, opened 2015-07-14
[Please Test] rewrite of the event class and the database

geier:
this should fix some issues (like #222 and #195) and enable the deletion and modification of single instances of recurring events (not implemented yet). 

and yes, I know some tests are failling.

 * [x] enable simple sorting of events
 * [x] print warning when using default_timezone
 * [ ] support for RDATE, EXDATE, RRULE:UNTIL with invalid timezone information
 * [x] re-fix RRULE editing


khal/khalendar/event.py
```diff
@@ -132,143 +208,124 @@ def symbol_strings(self):
...
+        """neeeded for vdirsyncer compat"""
+        return self._vevents[self.ref]['UID']
 
-        return '{}: {}{}{}{}'.format(
-            rangestr, self.summary, location, repitition, description)
+    @property
+    def summary(self):
+        return self._vevents[self.ref].get('SUMMARY', '')
+
+    def update_summary(self, summary):
```
untitaker: Use

    @summary.setter
    def summary(self, summary):
        ...

khal/khalendar/event.py
```diff
@@ -132,143 +208,124 @@ def symbol_strings(self):
...
+        self._vevents[self.ref]['SUMMARY'] = summary
+
+    @property
+    def location(self):
+        return self._vevents[self.ref].get('LOCATION', '')
+
+    def update_location(self, location):
```
untitaker: Use setter.

khal/khalendar/event.py
```diff
@@ -132,143 +208,124 @@ def symbol_strings(self):
...
+        self._vevents[self.ref]['LOCATION'] = location
+
+    @property
+    def description(self):
+        return self._vevents[self.ref].get('DESCRIPTION', '')
 
-    def compact(self, day):
+    def update_description(self, description):
```
untitaker: Use setter.

khal/khalendar/event.py
geier: while this might make sense at the moment, all those update methods will soon get a `ref` and `thisandfuture` flag.

untitaker: How would you get the summary of *past* events? Is there an `Event` instance for each instance, and do these share all the same `self._vevents`?

untitaker: Perhaps what we need is an additional class called `Recurrence`, so `self._vevents` is a list of `Recurrence`-instances. This class then has all the properties (`location`, `summary`).

geier: Yes, an Event represents only one specific recurrence instance and all recurrence instances of the same proto Event have the same self._vevents (but not the identical object). I don't believe that with the current functionality of khal we really need to access other recurrence instances properties (if they are deviating).

geier:
while this isn't as finished yet as I would like, I'm merging now because it's better than before and I hopefully have the time to fix all new bugs in the next week.

PR #225, opened 2015-07-16
interactive khal: EventEditor: Define prompt for summary to Title:

sithglan:


untitaker:
The idea was that with the summary being the "title" of the current page/view, it wouldn't require a label. Personally I think the editor UI is a bit cluttered, this was an attempt at improving that. See this screenshot from Baikal 2:

2015-07-16-114346_429x313_scrot

The title of that tooltip is editable.

It seems that this is not working out too well -- I wonder if a default of "New Event" would aid in finding that input field.

untitaker:
I wonder if we could use one huge textfield ala `git` and use the first paragraph as summary.

sithglan:
My plan for the event editor is to make it look like this:

    Title:                              Today's Dates
    Location:
    Calendar:
    StartStopEditor                     Mini Calender
    Repeat                              for month
    Reminder
    Description as Textbox

And make it possible to call vim to at least edit the description maybe even more where it makes sense.

geier:
I believe this is an improvement over the current status, but I also think we should discuss how the editor should look in the future.

geier:
@untitaker I didn't even know that Baikal comes with a web interface, but I don't like it too much.

PR #249, opened 2015-08-15
adjust date used in printformats to better illustrate options

coachshea:
This is the small, but IMHO, important change to the default date in
printformats that I had suggested in an issue and you had suggested to do in a
pull request.


untitaker:
@coachshea I suppose this can be closed?

coachshea:
Yes

-----Original Message-----

Issue #256, opened 2015-08-25
type inconsistencies


geier:
see https://github.com/geier/khal/issues/252#issuecomment-134566223 and 8832c4e

geier:
Since moving to python 3 only this really shouldn't be an issue any more, in case it is we have some asserts in case.

PR #257, opened 2015-08-26
[WIP] Use six instead of our own `compat` module.

hobarrera:
Using [six](https://pythonhosted.org/six/) has a few advantages:

 * Less code to maintain. Less is more.
 * More thoroughly tested compatibility code.
 * A few more things that we can make use of to maintain python2-compatibility (or py3-compat, depending on what you're using).
 * I *think* we can use this for #256, though I've failed so far.

This is tagged WIP, since I still need to get rid of a single usage of `to_unicode` and `to_bytes` that I haven't managed to get my head around. I'd also like to hear some opinions on this change.

### DIFF
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -23,8 +23,6 @@
 """this module contains some helper functions converting strings or list of
 strings to date(time) or event objects"""
 
-from .compat import to_unicode
-
 from datetime import date, datetime, timedelta
 from datetime import time as dtime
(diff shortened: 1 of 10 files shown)


hobarrera:
Huh. This is odd. This seems to be failing on travis, but succeeding locally. o.O

geier:
I'm good with moving to six.

PR #260, opened 2015-09-03
crude fix for:

0mark:
I had two Errors occuring when importing one of my calenders:

 - TypeError: can't compare offset-naive and offset-aware datetimes
   (File "/usr/lib/python3.4/site-packages/khal/khalendar/aux.py", line 171, in sanitize_timerange)
This happened when a entry had start date with timezone, and end date without or vice versa (Those date where from a imported ical file)
I have not used python for a long time, and never used datetime. It works for me. But i guess the right thing to do would be to add the right timezone to the date without. I failed doing that, though :)

 - AttributeError: 'list' object has no attribute 'dt'
   (File "/usr/lib/python3.4/site-packages/khal/khalendar/aux.py", line 226, in invalid_timezone)
No idea where those came from. Anyway, i bet just returning false is only masking the real problem here.

I guess both problem correlate somehow. Maybe fixing the second one right will fix the first too :)

Here is one of the offending dates:
BEGIN:VCALENDAR
VERSION:2.0
PRODID:ownCloud Calendar 0.7.3
BEGIN:VEVENT
SUMMARY:Termin:Junghackertag
URL:http://wiki.hamburg.ccc.de/Termin:Junghackertag
UID:http://wiki.hamburg.ccc.de/Termin:Junghackertag
DTSTART:20150530T120000
DTEND;VALUE=:20150530T160000Z
DESCRIPTION:Junghackertag
DTSTAMP:20150526T182050
SEQUENCE:10172
END:VEVENT
END:VCALENDAR

### DIFF
--- a/khal/khalendar/aux.py
+++ b/khal/khalendar/aux.py
@@ -132,6 +132,11 @@ def sanitize(vevent, default_timezone, href='', calendar=''):
     # TODO do this for everything where a TZID can appear (RDATE, EXDATE,
     # RRULE:UNTIL)
     for prop in ['DTSTART', 'DTEND', 'DUE', 'RECURRENCE-ID']:
+        if prop=='DTEND' and isinstance(vevent[prop], list):
+            value = vevent[prop][1]
+            vevent.pop(prop)
+            vevent.add(prop, value)
+
         if prop in vevent and invalid_timezone(vevent[prop]):
             value = default_timezone.localize(vevent.pop(prop).dt)
             vevent.add(prop, value)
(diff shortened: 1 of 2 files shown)


0mark:
I

geier:
Hi and thank you for reporting this. I believe the behaviour for events with start times with and and times without time zones is undefined and I'm not really sure what the best solution here is (strip the tz information of one or add the timezone to the other time).

I also think that at least a warning should be given to the user.

I'll have a look at the code tonight.

geier:
Sorry it took me so long, I did forget about this.
I would suggest to only add a new *sanitize* function instead, that checks if DTSTART and DTEND are of the same kind and if not, adjust DTEND accordingly and print a warning. I believe the other problems will vanish than (if they do exist with the above example).

0mark:
Oh nice, i think that should be by far the better sollution :)

geier:
@0mark Are still interested in fixing this?

0mark:
I am, i just dont have time currently. Maybe next week. Sorry, to much to do and not enough time. A fix for that would be nice too ;)

hobarrera:
#360 addresses the first problem, but I can't (nor couldn't) reproduce the second one. @0mark: Is this still an issue for you?

geier:
closed due to inactivity. Please open a new one if needed.

Issue #89, opened 2014-06-24
ikhal & khal: highlight days with events in calendar overview


smu:
It would be great, if the ikhal calendar overview (left side) could already indicate which days have events. For example these dates could be shown by a bold font.

sante81:
Any develop in this way???

libre-man:
How could this be implemented the best?

Bold text is an option, but that doesn't give any information about the
calendar the event is in. This is especially annoying when there are
calendars that have events for all dates (like I have). So I wouldn't really
like.

We could color the day with the color of the calendar. But what color should
days with multiple events from multiple calenders get? Maybe a priority
setting?

And I don't know what the impact on performance would be, but it shouldn't be
to high.

On Fri, Jun 05, 2015 at 05:38:04AM -0700, Matteo wrote:
> Any develop in this way???
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/geier/khal/issues/89#issuecomment-109281860

-- 
Zorg dat e-mail echt priv� is: onderteken en/of versleutel uw antwoord met 
OpenPGP.

untitaker:
This is sort-of implemented in Android, the color gets black (in our case probably bold) if there are events from multiple calendars. We probably can't do any distinction between all-day events and the rest as Android does.

![](http://www.androidpolice.com/wp-content/uploads/2011/11/wm__0000_2011110214h30_43.jpg)

libre-man:
I think a background color would be better because in the normal khal output
week numbers are bold, so that could cause confusion.

I do think there should be a setting for each calendar if it should mark days
which have an event from that calendar. Some people have weather calendars and
they probably don't want them to mark all the days.

On Fri, Jun 05, 2015 at 05:54:09AM -0700, Markus Unterwaditzer wrote:
> This is sort-of implemented in Android, the color gets black (in our case probably bold) if there are events from multiple calendars. We probably can't do any distinction between all-day events and the rest as Android does.
> 
> ![](http://www.androidpolice.com/wp-content/uploads/2011/11/wm__0000_2011110214h30_43.jpg)
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/geier/khal/issues/89#issuecomment-109284069

-- 
Zorg dat e-mail echt privé is: onderteken en/of versleutel uw antwoord met 
OpenPGP.

untitaker:
[...]

PR #271, opened 2015-10-01
Implement #89 for khal

dzoep:
Implemented highlighting of days with events in khal calendar display.

Fully configurable - foreground/background highlight, multiple colors for one day, defaults.

Not supported: "light"/bold colors. Right now the implementation just strips dark/light from the color name. Will add that later.

### DIFF
--- a/khal/calendar_display.py
+++ b/khal/calendar_display.py
@@ -92,9 +92,9 @@ def str_highlight_day(day, devents, hmethod, default_color, multiple, color):
     return dstr
 
 
-def str_week(week, today, collection,
-             hmethod, default_color, multiple, color,
-             highlight_event_days, locale):
+def str_week(week, today, collection=None,
+             hmethod=None, default_color=None, multiple=None, color=None,
+             highlight_event_days=0, locale=None):
     """returns a string representing one week,
     if for day == today colour is reversed
 
@@ -107,16 +107,19 @@ def str_week(week, today, collection,
     :rtype: str
     """
     strweek = ''
-    localize = locale['local_timezone'].localize
     for day in week:
-        start = localize(datetime.datetime.combine(day, datetime.time.min))
-        end = localize(datetime.datetime.combine(day, datetime.time.max))
-        devents = collection.get_datetime_by_time_range(start, end) + collection.get_allday_by_time_range(day)
         if day == today:
             day = style(str(day.day).rjust(2), reverse=True)
-        elif len(devents) > 0 and highlight_event_days != 0:
-            day = str_highlight_day(day, devents, hmethod, default_color,
-                                    multiple, color)
+        elif highlight_event_days != 0:
+            localize = locale['local_timezone'].localize
+            start = localize(datetime.datetime.combine(day, datetime.time.min))
+            end = localize(datetime.datetime.combine(day, datetime.time.max))
+            devents = collection.get_datetime_by_time_range(start, end) + collection.get_allday_by_time_range(day)
+            if len(devents) > 0:
+                day = str_highlight_day(day, devents, hmethod, default_color,
+                                        multiple, color)
+            else:
+                day = str(day.day).rjust(2)
         else:
             day = str(day.day).rjust(2)
         strweek = strweek + day + ' '
--- a/khal/calendar_display.py
+++ b/khal/calendar_display.py
@@ -70,8 +70,12 @@ def str_week(week, today, collection=None, conf=None):
             day = style(str(day.day).rjust(2), reverse=True)
         elif len(devents)>0 and conf['default']['highlight_event_days']!=0:
             dstr = str(day.day).rjust(2)
-            hmethod=conf['highlight_days']['method']
-            dcolor=urwid_to_click(devents[0].color)
+            hconf = conf['highlight_days']
+            hmethod = hconf['method']
+            if hconf['color']=='':
+                dcolor = urwid_to_click(devents[0].color)
+            else:
+                dcolor = urwid_to_click(hconf['color'])
             if hmethod=="foreground" or hmethod=="fg":
                 day = style(dstr, fg=dcolor)
             else:
--- a/khal/calendar_display.py
+++ b/khal/calendar_display.py
@@ -49,7 +49,7 @@ def getweeknumber(date):
     return datetime.date.isocalendar(date)[1]
 
 
-def str_week(week, today, collection=None):
+def str_week(week, today, collection=None, conf=None):
     """returns a string representing one week,
     if for day == today colour is reversed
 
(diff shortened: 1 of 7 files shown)


khal/cli.py
```diff
@@ -230,7 +230,8 @@ def calendar(ctx, days, events, dates):
             weeknumber=ctx.obj['conf']['locale']['weeknumbers'],
             show_all_days=ctx.obj['conf']['default']['show_all_days'],
             days=days or ctx.obj['conf']['default']['days'],
-            events=events
+            events=events,
+            conf=ctx.obj['conf']
```
untitaker: Please decide on one strategy:

* Pass each piece of data explicitly (firstweekday, encoding, locale, ...)
* Pass the config as single param

Since the choice has already been made, please try to stick with it.

khal/terminal.py
```diff
@@ -93,3 +93,10 @@ def merge_columns(lcolumn, rcolumn, width=25):
     rows = ['    '.join(one) for one in zip_longest(
         lcolumn, rcolumn, fillvalue='')]
     return rows
+
+
+def urwid_to_click(color):
+    """convert urwid color name to click color name
+    """
+    colors = color.split()
+    return colors[len(colors) - 1]
```
untitaker: You can use `[-1]` here.

dzoep:
It doesn't look elegant at all, but it should be ok. Again, any suggestions are welcome.

khal/calendar_display.py
```diff
@@ -49,7 +51,50 @@ def getweeknumber(date):
...
-def str_week(week, today):
+def get_event_color(event, default_color):
+    """Because multi-line lambdas would be un-Pythonic
+    """
+    if event.color == '':
+        return default_color
+    return event.color
+
+
+def str_highlight_day(day, devents, hmethod, default_color, multiple, color):
```
untitaker: I'd just return the color here, and not take a string. `style(..., fg=...)` is short enough. This eliminates the `hmethod` parameter.

untitaker: Ah, I see you use multiple colors...

untitaker:
Do we really need to make every single detail of this configurable? The added complexity is staggering...

dzoep:
As noted in the #89 - it is an open question how exactly should the days with events be represented. Making it fully configurable seems to me the best approach. And also I find it pretty straightforward to implement.

untitaker:
Maintaining the additional code is another story...

khal/calendar_display.py
```diff
@@ -26,6 +26,8 @@
 
 from click import style
 
+from terminal import urwid_to_click
```
geier: you'll need to change this to ``from .terminal import urwid_to_click`` for python 3 compat

geier:
I had a look, gave it a try, like it and do think this will be a valuable addition to khal (I believe it has been the most sought-after feature). I do have a small nit-picks I'll add as comments.

Furthermore while I do like the customizability and I think we should add it like it is for the moment I believe we should than ask the users what they use and strip out the options nobody uses (def. not now, but before we hit a 1.0, if that ever is the case), just to keep the code base small. On the other hand, if somebody other than me is willing to maintain it (you?) I don't have a problem with it either.

khal/calendar_display.py
dzoep: Thanks for the hint, fixed.

khal/settings/khal.spec
```diff
@@ -138,6 +138,10 @@ days = integer(default=2)
 # event in text form, the path to where the event is now saved or nothing?
 print_new = option('event', 'path', 'False', default=False)
 
+# If set to 1, khal will highlight days with events. Options for
+# highlighting are in [highlight_days] section.
+highlight_event_days = integer(default=0)
```
geier: could you make this a boolean? (it's what we are using for the other yes no decisions) Or are you planning to make this even more configurable?

tests/settings_test.py
```diff
@@ -42,6 +42,7 @@ def test_simple_config(self):
                 'show_all_days': False,
                 'print_new': 'False',
                 'days': 2,
+                'highlight_event_days': 0
```
geier: if above comment applies, this and the next one would then need to change, too

khal/calendar_display.py
```diff
@@ -65,6 +110,16 @@ def str_week(week, today):
     for day in week:
         if day == today:
             day = style(str(day.day).rjust(2), reverse=True)
+        elif highlight_event_days != 0:
+            localize = locale['local_timezone'].localize
+            start = localize(datetime.datetime.combine(day, datetime.time.min))
+            end = localize(datetime.datetime.combine(day, datetime.time.max))
+            devents = collection.get_datetime_by_time_range(start, end) + collection.get_allday_by_time_range(day)
```
geier: could you perhaps split the line with a ``+ \`` ?

geier:
Oh, and could you add yourself to AUTHORS.txt, please?

dzoep:
UTF-8 or ASCII only?

There's this ů and č in y name :)

dzoep:
On 10/01/2015 10:25 PM, Christian Geier wrote:
>
> I had a look, gave it a try, like it and do think this will be a
> valuable addition to khal (I believe it has been the most sought-after
> feature). I do have a small nit-picks I'll add as comments.
>
Tried to resolve all of them.
>
> Furthermore while I do like the customizability and I think we should
> add it like it is for the moment I believe we should than ask the
> users what they use and strip out the options nobody uses (def. not
> now, but before we hit a 1.0, if that ever is the case), just to keep
> the code base small. On the other hand, if somebody other than me is
> willing to maintain it (you?) I don't have a problem with it either.
>
No problem maintaining it. I'm kind of used to maintaining various
things for longer than a decade. However I'm not a python person - so be
patient with me :)

geier:
Quoting Dominik Joe Pantůček (2015-10-01 22:43:57)
> UTF-8 or ASCII only?
> 
> There's this ů and č in y name :)

UTF-8 is good, if it breaks anything it needs to be fixed (the tool, not
your name).

khal/calendar_display.py
```diff
@@ -65,6 +110,17 @@ def str_week(week, today):
     for day in week:
         if day == today:
             day = style(str(day.day).rjust(2), reverse=True)
+        elif highlight_event_days:
+            localize = locale['local_timezone'].localize
+            start = localize(datetime.datetime.combine(day, datetime.time.min))
+            end = localize(datetime.datetime.combine(day, datetime.time.max))
+            devents = collection.get_datetime_by_time_range(start, end) + \
+                      collection.get_allday_by_time_range(day)
```
geier: it's too far indented now (yes, I know.... sorry about this)
https://travis-ci.org/geier/khal/jobs/83193750
you can also run flake8 on your code in addition to pep8 to see if it will raise any errors

geier:
Quoting Dominik Joe Pantůček (2015-10-01 22:57:43)
> However I'm not a python person - so be
> patient with me :)

Sure! As long as you are patient with all that formating crap. I found
it strange when getting into the python ecosystem, but now I'm so used
to it, code just looks wrong otherwise.

geier:
@dzoep Thx! From my point of view, this is good to merge. What do you say?

dzoep:
Let's merge it and I'll continue with ikhal and request another pull when that part is ready.

geier:
Done. And thank you again!

PR #274, opened 2015-10-02
Fix error when trying to exclude calendars via cli

fkemmer:
Previous behaviour:
khal agenda -d some_calendar

    [...]
    AttributeError: 'tuple' object has no attribute 'remove'

Now:
works (calendar is being excluded)

Reported-by: @matthiasbeyer


geier:
Thank you for this fix!

PR #276, opened 2015-10-04
highlight days with events in calendar overview

dzoep:
The ikhal part of #89 

Bear in mind that if you enable highlight_event_days the startup is as slow as hell. The problem lies in get_datetime_by_time_range and get_allday_by_time_range which is called number-of-days-displayed times.

We should definitely NOT make this default before optimizing the CalendarCollection's get_*_by_time_range methods or maybe even adding get_*_by_day and using some lookup structures that would be filled upon appending calendar to the collection (and updated upon calendar event creation/modification/deletion).


geier:
Thank you! Could you have a look at the [failing tests](https://travis-ci.org/geier/khal/builds/83559395)?

geier:
Otherwise this looks good to me, though you are right, this should definitively not be enabled by default as it is *slow*.

dzoep:
I'll look at the tests, there's some problem with using localize, I'm not entirely familiar with that so gimme some time.

Also right now there's no support for background highlighting. Would you be OK with adding styles like 'bg dark red'... in colors.py? Apparently you use styles named after colors for foreground case, so I would use similarly named styles for background case.

Regarding the speed - I think it can be improved, but as it relates to CalendarCollection internals, I'd open a separate issue for that and rather will not mess this up too much.

geier:
Quoting Dominik Joe Pantůček (2015-10-05 12:57:56)
> I'll look at the tests, there's some problem with using localize, I'm not entirely familiar with that so gimme some time.
If you need any help, please let me know.

> Also right now there's no support for background highlighting. Would you be OK with adding styles like 'bg dark red'... in colors.py? Apparently you use styles named after colors for foreground case, so I would use similarly named styles for background case.

Sure, no problem.

> Regarding the speed - I think it can be improved, but as it relates to CalendarCollection internals, I'd open a separate issue for that and rather will not mess this up too much.

I believe it will get even messier than that, but I agree in doing it
later.

dzoep:
OK, I think it is ready for merge. I expect further discussion in the issue page, first the speed and second - the ability to turn off repeated events highlighting.

geier:
Thanks @dzoep

PR #286, opened 2015-10-10
Fix for error when removing recursion.

### LINKED ISSUE #285: TypeError: 'NoneType' object is not iterable
When removing an recursion I have a TypeError because it tries to add the None recursion.

libre-man:
This fixes #285.


geier:
A merged this with 118b84752e08ee6cbea77a5a4ee3d2064f034d54, but it looks like github doesn't get it (probably didn't follow their protocol). Thanks again!

PR #287, opened 2015-10-13
update and improve zsh completion file

okapia:
This improves the zsh completion. Firstly, I've tried to cover all new options and subcommands in the latest git version of khal. Other changes bring it in line with standard good practice for zsh completion functions: e.g. use of _wanted and _call_program, setting curcontext to include the subcommand, adding exclusion lists for mutually exclusive options. I've removed the superfluous redefinition  of _directories. I've only used _dates for completion after the agenda --days option. It'd be good if it could give a bit more of a clue as to the formats accepted for dates/times by khal new but I'm a little confused as to what actually is valid. Which of the configurable date formats are used? I also didn't use _time_zone for timezone completion because I'm not sure it is helpful. What formats are allowable for timezone? "UTC" doesn't seem to work while Europe/Berlin does. And specifying, e.g +0100 results in 'DTSTART has invalid or incomprehensible timezone information' messages.

One final comment: why is the function named with double underscores? Just _khal would be more conventional.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -14,3 +14,4 @@ Thomas Glanzmann - thomas [at] glanzmann [dot] de - https://thomas.glanzmann.de/
 John Shea - coachshea [at] fastmail [dot] com
 Dominik Joe Pantůček - joe [at] joe [dot] cz - http://joe.cz/
 Thomas Schaper - libreman [at] libremail [dot] nl
+Oliver Kiddle - okiddle [at] yahoo [dot] co [dot] uk
--- a/misc/__khal
+++ b/misc/__khal
@@ -5,34 +5,21 @@
 # fpath=(~/.zsh/completion $fpath) 
 
 _calendars() {
-  local accounts
-  accounts=( ${${(f)"$(command khal printcalendars 2> /dev/null)"}/[$'\n']##/} )
-  compadd "$a" ${(kv)=accounts}
+  local expl
+  _wanted calendars expl calendar compadd \
+      ${(f)"$(_call_program calendars khal printcalendars)"}
 }
 
-_ics_files() {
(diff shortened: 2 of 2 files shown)


geier:
Hi @okapia, welcome to khal and thank you for this very much appreciated pull request!

I really have no idea how to properly write zsh completions (TBH I found the documentation on that a bit lacking, at least I did not find anything I liked). So if I did something stupid/non standard (like _khal instead of __khal), feel free to send a PR.

Regarding what `khal new` should accept, please have a look [at the documentaion](http://lostpackets.de/khal/usage.html#new), if it is still unclear, please do ask again. For timezones only Europe/Berlin etc. are currently supported.

Also, please add yourself to `AUTHORS.txt`, I'll merge this than.

okapia:
I've pushed again with the AUTHORS.txt addition.

The aspect of `khal new` that I was not sure on is exactly what formats are valid: is it essentially any of those printed by `khal printformats`? Is 'tomorrow' the only special keyword? If that's combined with a time, is timeformat used?

The zsh manual is a reference so unfortunately not a great guide. There are some blog posts out there but nothing really worth recommending. I might find some time to do something about that one day.

geier:
Thank you, I'll merge this for now, but please don't hesitate to improve it further.

And you are right, `new` also accepts some keywords, namely `today`, `tomorrow` and all weekdays (including their three letter abbreviations). If a keyword is used to specify a date, a `time` can be used to specify a time.

In general `new` accepts the following combinations:
```
datetime [datetime|time]
date [date]
time [time]
```
with 
```datetime = (longdatetimeformat|datetimeformat|keyword-date timeformat)```

and
```time = timeformat```

Yes, the documentation need some work.

PR #297, opened 2015-11-11
Test-running improvements

hobarrera:
Minor improvements to how tests are ran, as well as running tests with a negative UTC offset which might help catch bugs earlier in future.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,6 @@
 [tox]
-envlist = py27, pypy, py34, style
+envlist = py27, pypy, py34, py35, style
+skip_missing_interpreters = True
 
 [testenv]
 passenv = LANG
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,20 @@
 [tox]
-envlist = py27, pypy, py34, py35, style
+envlist = py27, pypy, py34, py35, negtz, style
 skip_missing_interpreters = True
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -9,7 +10,18 @@ deps =
 commands =
 	py.test {posargs}
 
+[testenv:negtz]
+passenv = LANG
+setenv =
+    TZ = America/Argentina/Buenos_Aires
+deps =
```
untitaker: I'm not really sure, but I suspect that these params could be inherited from `[testenv]`.

hobarrera: Yup, you're right: https://testrun.org/tox/latest/config.html#substitution-for-values-from-other-sections

I'll look into it later today.

hobarrera: These seems to be a bug in `tox` that prevents us from using this 100%:

`commands = {[testenv]commands}`

Does not substitute `{posargs}` in `commands` before the inheritance :(

untitaker: I mean this part:

    Defaults for each setting in this section are looked up in the:

    [testenv]
    ...

    testenvironment default section.

hobarrera: Duh! Thanks. Fixed.

untitaker:
LGTM, thanks!

PR #298, opened 2015-11-12
Remove unicode literals

geier:


khal/controllers.py
```diff
@@ -171,7 +169,7 @@ def agenda(collection, date=None, encoding='utf-8',
                               show_all_days=show_all_days, full=full, **kwargs)
     # XXX: Generate this as a unicode in the first place, rather than
     # casting it.
-    echo(to_unicode('\n'.join(event_column), encoding))
+    echo(to_unicode(u'\n'.join(event_column), encoding))
```
hobarrera: I don't thing we need to specify `u'` if we're passing the string to `to_unicode`.

untitaker: It has to be whatever type the strings in event_column are.

hobarrera: You're right.

Issue #207, opened 2015-06-13
Unicode UIDs


aitzkora:
Hi!
The issue related to vdirsyncer. First time I call ikhal after synchronized an empty calendar, I could see the following warnings

```
fux@txindoki ~ $ ikhal
Unknown exception happened.
Traceback (most recent call last):
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/khalendar.py",
line 202, in _update_vevent self._dbtool.update(event.raw, href=href, etag=etag)
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/backend.py",
line 252, in update for vevent in sorted(vevents, key=sort_key):
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/backend.py",
line 236, in sort_key uid = str(vevent['UID'])
UnicodeEncodeError: 'ascii' codec can't encode character
u'\xe9' in position 30: ordinal not in range(128)
warning: Skipping home/6f0f85340cdc4b26b8c74e953525fc3d.ics:
'ascii' codec can't encode character u'\xe9' in position 30: ordinal not in range(128)
warning: This event will not be available in khal.
Unknown exception happened.
Traceback (most recent call last):
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/khalendar.py", line 202, in _update_vevent
self._dbtool.update(event.raw, href=href, etag=etag)
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/backend.py",
line 252, in update for vevent in sorted(vevents, key=sort_key):
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/backend.py",
line 236, in sort_key uid = str(vevent['UID'])
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 30: ordinal not in range(128)
warning: Skipping home/29e504cfde4d4649a6187ea340caed42.ics:
'ascii' codec can't encode character u'\xe9' in position 30: ordinal not in range(128)
warning: This event will not be available in khal.
Unknown exception happened.
Traceback (most recent call last):
File "/home/fux/vdirsyncer2/lib/python2.7/site-packages/khal/khalendar/khalendar.py",
line 202, in _update_vevent self._dbtool.update(event.raw, href=href, etag=etag)
[...]

PR #304, opened 2015-11-28
Drop python2

### LINKED ISSUE #190: Review types of strings used internally
See https://github.com/geier/khal/pull/187

### LINKED ISSUE #302: Drop Python 2
OS/Distributions readiness to support python 3 only applications
===================

* :white_check_mark: Arch
* :white_check_mark: Fedora
*  :x:FreeBSD (no pre-build package available any more)
* :white_check_mark:: NetBSD/pkg.src 
* :question: openSUSE
* :white_check_mark: Debian/Ubuntu: there is no package anyway, but otherwise does look ready
* :white_check_mark: Gentoo
* what other (major) distributions/OSes are missing from this list?

**Feel free to edit this**

hobarrera:
If we want to merge this or not really depends on what we decide on #302.

In the meantime, I went ahead and did the basic cleanup to drop py27 and the compatibility layer we had for it. There's probably more, small, cleanups that we can do and will come across in future.

Note that pypy is compatible with python 3.2.5, so we can't use that.  
Also, some doc said the minimum version was 3.3, and other places said 3.4. ~~I've standardized it to 3.4; I guess the rest was outdated.~~ It's 3.3. See below.

Closes #302
Closes #257 (no-longer-valid)
Closes #190
Check how this affects #207, #256.


untitaker:
`3.3` should work, also there's `pypy3` (not sure if it's production-ready)

hobarrera:
Like I said, pypy3 is only python 3.2.5 compatible (hence <3.3).
Installation of some dependencies failed on it, so I didn't look any further.

untitaker:
Argh, yes.

untitaker:
:+

hobarrera:
You're right, py33 works too. I've added it to the tests suite, and everything passes.

hobarrera:
Is it okay to also remove these lines?

    # vim: set ts=4 sw=4 expandtab sts=4 fileencoding=utf-8:

I actually use vim too, but I think that most contributors should already have this configured this way. It's also present in very few files anyway, so anyone relying on this will break other files anyway.

untitaker:
Yes, probably.

If we care about this http://editorconfig.org/ might be the way to go.

geier:
Does python 3 assume files to be utf-8 encoded?

hobarrera:
> Does python 3 assume files to be utf-8 encoded?

It does: https://docs.python.org/3/howto/unicode.html#the-string-type

geier:
Than sure, let's get rid of those vim lines (it probably was me who put them there in the first place and I don't need them).

geier:
do we need to add something to `setup.py` to indicate it's python 3 only?

Also, two weeks are nearly up :) and it looks like we won't loose too many users (I'll probably add a [short description on how to run khal from a virtualenv](http://lostpackets.de/blog/vdirsyncer-on-debian.html) later).

@hobarrera  feel free to merge this

hobarrera:
> do we need to add something to setup.py to indicate it's python 3 only?

I changed the hashbang from `python` to `python3`. I also updated the classifiers, to pypi will reflect the python3-status.

PR #306, opened 2015-12-08
Basic event export

fpytloun:
Very basic event export into ICS (eg. for sending as mail attachment).
Some possible enhancements:
- ask what to do if file already exists
- handle exception or at least say something if event is saved
- export from events list

Related to https://github.com/geier/khal/issues/43

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -15,3 +15,4 @@ John Shea - coachshea [at] fastmail [dot] com
 Dominik Joe Pantůček - joe [at] joe [dot] cz - http://joe.cz/
 Thomas Schaper - libreman [at] libremail [dot] nl
 Oliver Kiddle - okiddle [at] yahoo [dot] co [dot] uk
+Filip Pytloun - filip [at] pytloun [dot] cz - https://fpy.cz
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -24,6 +24,7 @@
 
 from datetime import date, datetime, time, timedelta
 
+import os
 import icalendar
 import pytz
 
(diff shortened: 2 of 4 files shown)


geier:
Hi @fpytloun and thank you for this pull request. I believe your PR is a nice improvement over the current situation (no export at all). Before I'm merging this, please add yourself to `AUTHORS.txt`.

fpytloun:
Thank you :-)
I would like to get more familiar with the code and enhance it in near future - especially way of sending invitation via email which will make my life easier :-)

fpytloun:
Updated the code to be able to export from events list by `e` keybinding.
It duplicates the code a little but I don't know better place where to put the export logic.

hobarrera:
It **really** looks like lots of code should be deduped.  
Maybe the `Calendar` class would be a good place for the export code?

fpytloun:
Most of the duplicated code is about UI interaction which probably can't be placed anywhere else.
The export code alone is only about this:
```
            export_path = os.path.expanduser(user_data.get_edit_text())
            try:
                with open(export_path, 'w') as fh:
                    fh.write(self.event.raw)
```
I will try to put this part into `Calendar` class.

fpytloun:
I have put the `export_ics` code under `Event` as it basically only wraps `Event.raw`. It deduplicates the code a bit.

geier:
@fpytloun it does work for me, is this good to merge?

geier:
Regarding sending invitations per email: you are mostly there I guess, just save the file in /tmp and than run a (configurable) command like `mutt -a {file}`. But because one might want to configure this for several programs, the current keybinding config section might need to be redone.

If you need any help getting around, please get in touch with me.

fpytloun:
@geier yes, I am thinking about making action configurable per-calendar using something like `xdg-open mailto:?subject=invitation_whatever&attachment=/tmp/invitation.ics` or `mutt -a` as you noted.
And still considering possibility of implementing real rfc2445 invitations.. :-)

Anyway this change is probably finished and ready for merge.

geier:
alright, I'll merge this then. Thank you again!

PR #307, opened 2015-12-10
Fix broken path in doc/source/Makefile

lfam:
In the 0.7 release tarball on PyPI [0], this path is broken, because the
tarball unpacks into 'khal-0.7.0' instead of 'khal'.

This bug prevents the documentation from being built.

This patch corrects the path, without depending on the name of the
top-level directory that the tarball unpacks into.

[0]
https://pypi.python.org/pypi/khal/0.7.0


geier:
Thank you!

PR #308, opened 2015-12-11
Show birthdays for contacts with no FN

hobarrera:
Lots of my contacts don't have an `FN` so their birthdays are not listed. They *do* have `N`, so I used that instead.

[Related documentation](https://tools.ietf.org/html/rfc6350#section-6.2)

I'm not sure if this is always valid: are there scenarios in which the name orders should be different? Do we care about those?

### DIFF
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -552,7 +552,11 @@ class SQLiteDb_Birthdays(SQLiteDb):
                 logger.info('cannot parse BIRTHDAY in {} in collection '
                             '{}'.format(href, self.calendar))
                 return
-            name = vcard['FN']
+            if 'FN' in vcard:
+                name = vcard['FN']
+            else:
+                n = vcard['N'].split(';')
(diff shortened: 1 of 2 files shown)


hobarrera:
No longer `WIP`, so awaiting feedback. :D

geier:
While a vcard without an `FN` property is not valid and you should try to get that software fixed, but I do believe in "be lenient in what you accept, be strict in what you send out", so thanks and merged!

hobarrera:
I'll try and track down what keeps generating them and report the issue - I do belive it's Fastmail's UI, but I have to double-check first.

hobarrera:
FWIW, I tracked this down recently and it wasn't Fastmail's UI (let's not taint their name in vain), but Jolla/Sailfish's built-in address book.

I've reported this upstream and it have been recently [fixed](https://git.merproject.org/mer-core/buteo-sync-plugin-carddav/commit/9c7d183efb7524158599118e6c3a4fd340a0a8d9).

PR #310, opened 2015-12-14
Feature faster backend

geier:


geier:
Well PEP8 does "allow" up to 100 characters if the team agrees, and I remember seeing a PEP that updated this to something like 110, but I cannot find it now.

But I'll try to keep it to 80.

tests/backend_test.py
```diff
@@ -406,37 +430,34 @@ def test_event_rrule_this_and_future_allday_prior():
...
+    assert events[3].end == date(2014, 7, 20)
+    assert events[4].end == date(2014, 7, 24)
+    assert events[5].end == date(2014, 7, 31)
 
-    assert str(events[0].summary) == 'Arbeit'
-    for event in events[1:3] + events[4:6]:
-        assert str(event.summary) == 'Arbeit (lang)'
-    for event in events[3:4] + events[6:]:
-        assert str(event.summary) == 'Arbeit (neu)'
+    assert str(events[0].summary) == u'Arbeit'
```
hobarrera: This is probably due to being based on an older `master`, but `u'` should no longer be present.

hobarrera:
After inspecting the code (but not actually running it), I like the changes, and how a lot of things are quite clearer/tidier.

Given test coverage, I assume nothing broke, but I'll give it a try later on just in case and give you any feedback.

tests/backend_test.py
geier: oops

geier:
and this is only the first part...

Issue #309, opened 2015-12-13
Test failure: test_only_update_old_event


lfam:
Hi,

I'm trying to package khal for GNU Guix [0], but the build process fails during the test phase on ```test_only_update_old_event```.

Specifically, the assertion that ```cal._db_needs_update()``` fails.

I've attached the build log: [khal-0.7.0-build.txt](https://github.com/geier/khal/files/60379/khal-0.7.0-build.txt)

Please let me know if I can provide more information or help in any other way.

Here's the logged output of the test suite, cut from the full log:
```
starting phase `check'
============================= test session starts ==============================
platform linux -- Python 3.4.3 -- py-1.4.23 -- pytest-2.6.1
plugins: cov
collected 154 items

tests/aux_test.py ............
tests/backend_test.py ......................
tests/cal_display_test.py ....x
tests/cli_test.py .....x........
tests/controller_test.py ...
tests/event_test.py .............................
tests/khalendar_aux_test.py ................................
tests/khalendar_test.py .....................F
tests/settings_test.py .....
tests/terminal_test.py ...
tests/vtimezone_test.py ...
tests/ui/test_calendarwidget.py ...
tests/ui/test_widgets.py .

=================================== FAILURES ===================================
__________________________ test_only_update_old_event __________________________

cal_vdir = (<khal.khalendar.khalendar.Calendar object at 0x7ffff1e28518>, <FilesystemStorage(**{'path': '/tmp/nix-build-khal-0.7.0.drv-0/pytest-0/test_only_update_old_event0'})>)
monkeypatch = <_pytest.monkeypatch.monkeypatch object at 0x7ffff0f7b748>

    def test_only_update_old_event(cal_vdir, monkeypatch):
        cal, vdir = cal_vdir
    
        href_one, etag_one = vdir.upload(cal.new_event(dedent("""
        BEGIN:VEVENT
        UID:meeting-one
        DTSTART;VALUE=DATE:20140909
        DTEND;VALUE=DATE:20140910
        SUMMARY:first meeting
        END:VEVENT
        """)))
    
        href_two, etag_two = vdir.upload(cal.new_event(dedent("""
        BEGIN:VEVENT
        UID:meeting-two
        DTSTART;VALUE=DATE:20140910
        DTEND;VALUE=DATE:20140911
        SUMMARY:second meeting
        END:VEVENT
        """)))
    
        cal.db_update()
        assert not cal._db_needs_update()
    
        old_update_vevent = cal._update_vevent
[...]

PR #316, opened 2015-12-23
Refactor backend

geier:


geier:
It's getting better, thanks to @untitaker suggestion in #320 and not using FLOAT for the ctag in the SQLite DB. I'm periodically getting the same error as @lfam in #309 now. Not really sure where they come from yet.

khal/khalendar/khalendar.py
```diff
@@ -291,36 +113,40 @@ def default_calendar_name(self, default):
...
+            raise ValueError(
+                'Calendar "{0}" is read-only and cannot be used as default'.format(default))
+
+    def _local_ctag(self, calendar):
+        stat = os.stat(self._calendars[calendar]['path'])
+        mtime = getattr(stat, 'st_mtime_ns', None)
+        if mtime is None:
+            mtime = stat.st_mtime
+        import math
+        return str(int(math.floor(mtime * 1e9)))
```
untitaker: Two unconditional returns here.

geier: True, I was experimenting and not finished here. Also I'm still not sure why the `sleep(1)` is needed.

untitaker: Sleep shouldn't be necessary at all.

geier: but I need to at least introduce a `sleep(0.01)` to reproducible get the test to pass.

geier:
As this seems to work well enough I'll merge this now (I'm sick of rebasing).

Issue #301, opened 2015-11-19
Entering Alarms in khal


petr-fischer:
It will be nice If I can enter event alarms in khal UI. Something like "1 hour before", "2 hours...", "1 day before".... "1 week before" etc...

This enhancement is not about triggering alarms in the console, but about entering alarm in event edit screen. 

When synced to the phone (via CalDAV), alarms will be triggered on the phone.

geier:
Sure, this would be very doable. PRs are very welcome.

languitar:
I'd also be interested in that.

languitar:
I have started to implement something here: https://github.com/languitar/khal/tree/alarms

However, when I save an event with a changed date for an alarm I end up in this error and I cannot find out what is causing this:
```
Traceback (most recent call last):
  File "/homes/jwienke/workspace/src/khal/khal/ui/__init__.py", line 756, in start_pane
    loop.run()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 278, in run
    self._run()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 376, in _run
    self.event_loop.run()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 682, in run
    self._loop()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 719, in _loop
    self._watch_files[fd]()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/raw_display.py", line 393, in <lambda>
    event_loop, callback, self.get_available_raw_input())
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/raw_display.py", line 493, in parse_input
    callback(processed, processed_codes)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 403, in _update
    self.process_input(keys)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/main_loop.py", line 503, in process_input
    k = self._topmost_widget.keypress(self.screen_size, k)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/urwid/wimp.py", line 643, in keypress
    return self._current_widget.keypress(size, key)
[...]

Issue #148, opened 2015-01-29
Caldav daemon to launch popups before events 


teto:
I had looked for a program like khal one year ago without success but I am very glad to have found it. My question is not directly related but might become. I am looking for a daemon able to read my calendar and warn me via popups when an event is closing in. Thunderbird does warn me but only when it's open. So I was wondering if anyone knew of such a tool and if not, would it make sense to divide khal into a daemon and a client application (as urxvt or powerline).

untitaker:
Khal could use ``atd`` instead of launching its own daemon.

untitaker:
Also I think @michaeladler wrote something like this for his own use?

geier:
Without thinking too much about this, I'm not a fan of writing our own
daemon here. Using either `at` or `cron` seems and any notification
program comes to my mind.

Quoting Markus Unterwaditzer (2015-01-29 18:05:40)

teto:
Somehow the workflow would be, write a cronjob that polls every X minutes from khal events within next X minutes events and pass those events to `at` ?
This should be enough if X is low. Otherwise if the calendar changes (event removed, or close event added) during the next X minutes it may become complex to handle.

michaeladler:
@untitaker: Nope, I pretty much wrote the opposite: a tool that removes annoying alerts from ics-files :)

teto:
@michaeladler I would like to be annoyed by such a tool, any name in mind ?

geier:
I'm happy to support any such endeavor with any custom functions needed, but won't get to implement the tool itself for quite some time.

geier:
@mbaldessari 's suggestion:

It would be extremely cool to have something (just thinking aloud here) like:
`khal alarms --script foo --delta 60min` which will call foo for every event that will take place
in the next 60 minutes (passing the subject of the event + time as args). That way it could
be put in a cron job pretty easily.

untitaker:
I think that's problematic. If the command you posted were to be executed every hour, it could miss events that are happening within 60 mins and were synced recently. If the user tries to mitigate this by calling that command more often, the given script would be executed multiple times for one event.

I think a better approach is a hook in the config file that is called whenever khal pleases to do so. But in that case, the next version of vdirsyncer will include such a hook, it will invoke a script for each new or updated ics file, thanks to @michaeladler.

teto:
[...]

PR #321, opened 2016-01-07
Support for editing alarms

languitar:
As requested in #301, this is a work in progress pull request for editing alarms.

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -351,6 +351,40 @@ class Event(object):
     def update_summary(self, summary):
         self._vevents[self.ref]['SUMMARY'] = summary
 
+    @staticmethod
+    def _can_handle_alarm(alarm):
+        """
+        Decides whether we can handle a certain alarm.
+        """
+        return alarm.get('ACTION') == 'DISPLAY' and isinstance(alarm.get('TRIGGER').dt, timedelta)
+
+    @property
+    def alarms(self):
+        """
+        Returns a list of all alarms in th original event that we can handle. Unknown types of
+        alarms are ignored.
+        """
+        return [(a.get('TRIGGER').dt, a.get('DESCRIPTION'))
+                for a in self._vevents[self.ref].subcomponents
+                if a.name == 'VALARM' and self._can_handle_alarm(a)]
+
+    def update_alarms(self, alarms):
+        """
+        Replaces all alarms in the event that can be handled with the ones provided.
+        """
+        components = self._vevents[self.ref].subcomponents
+        # remove all alarms that we can handle from the subcomponents
+        components = [c for c in components
+                      if not (c.name == 'VALARM' and self._can_handle_alarm(c))]
+        # add all alarms we could handle from the input
+        for alarm in alarms:
+            new = icalendar.Alarm()
+            new.add('ACTION', 'DISPLAY')
+            new.add('TRIGGER', alarm[0])
+            new.add('DESCRIPTION', alarm[1])
+            components.append(new)
+        self._vevents[self.ref].subcomponents = components
+
     @property
     def location(self):
         return self._vevents[self.ref].get('LOCATION', '')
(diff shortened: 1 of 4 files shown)


khal/khalendar/event.py
```diff
@@ -342,6 +342,40 @@ def summary(self):
...
+    def alarms(self):
+        """
+        Returns a list of all alarms in th original event that we can handle. Unknown types of
+        alarms are ignored.
+        """
+        return [(a.get('TRIGGER').dt, a.get('DESCRIPTION'))
+                for a in self._vevents[self.ref].subcomponents
+                if a.name == 'VALARM' and self._can_handle_alarm(a)]
+
+    def updated_alarms(self, alarms):
```
untitaker: You can use the `alarms.setter` decorator here, FWIW, but better to stay consistent.

But: please rename to `update_alarms`

languitar: Oops, sure. this was a typo.

geier:
Thank you! As far as I see, the only thing that really is left, is validation of user inputs, right?

While it doesn't make the event editor any prettier, that's not your fault at all.
We probably should improve the theming a lot, e.g. by making titles and user input field better separable.

languitar:
Yes, that's what is missing to my mind. And optionally supporting more types of alarms. It's funny that the Ical spec supports so many more types of alarms than ical on OSX supports. ;) Unfortunately I am currently a bit out of time for fixing this. I hope I find some time in a few days.

sdx23:
I'm a big fan of this feature. Thanks for the good work, keep it going :)

geier:
I created a new widget type `ValidateEdit`, this should make validation really easy, just pass along a validation function and you should be good to go. If you need any help, just say so. If you don't have the time for it, I will rebase and merge your branch if that's ok with you.

languitar:
@geier great, I've seen the commit and already thought that this might be useful. I hope I will find some time at the weekend to finalize this.

fpytloun:
Looks great! Only few more wishes:
- it would be good to setup default alarm for every new event in config (eg. notification 10 minute before event start)
- it would be great to introduce some `--alarm` command (to be executed from cron) which would execute action at alarm time (eg. I have [alarm.sh](https://gist.github.com/fpytloun/4acc181d67d78b47b23b) for this purpose)

geier:
I agree with @fpytloun first suggestion (but I'll happily merge this without it), but I won't do anything about the second one (see #148). However, if someone else wants to implement this (and keep supporting it), I'm happy to accept a PR.

geier:
Hi @languitar, I took the liberty of rebasing your PR and making the EditWidgets in the DurationWidget validate input (see branch `alarms`). If you are okay with that, I'll merge that branch. Could you add yourself to the `AUTHORS.txt` file, please?

languitar:
@geier Thank you. I'm sorry, I didn't find the time to do this so far.

Should I overwrite my own branch with yours and add the authors change or what is the idea?

geier:
@languitar nothing to be sorry for, you did all the work. I don't mind really, you cant just do a PR against my `alarms` branch, another PR with just the AUTHOR change against master, rebase your own `alarms` branch against current master and cherry-pick my commit (or just do it by hand, I don't care about atttribution). Just do what is most convenient for you.

languitar:
Ok, I have updated my own branch.

I also changed the return statement for the `unsigned_integer` function as discussed.

khal/ui/widgets.py
```diff
@@ -424,3 +424,107 @@ def keypress(self, size, key):
...
                 return
         return super().keypress(size, key)
+
+
+class DurationWidget(urwid.WidgetWrap):
+
+    @staticmethod
+    def unsigned_int(number):
+        """test if `number` can be converted to a positive int"""
+        try:
```
hobarrera: Looks like my previous comment got lost in the void during rebasing.

IMHO, this looks simpler/cleaner:

```python
try:
    return int(number) >= 0
except ValueError:
    return False
```

geier:
Thank you for your patience and this big improvement!

Issue #314, opened 2015-12-17
Documented grey colors cannot be used


languitar:
The documentation states that the following colors are usable: black, white, brown, yellow, dark grey, dark green, dark blue, light grey, light green, light blue, dark magenta, dark cyan, dark red, light magenta, light cyan, light red and **

Using one of the grey colors results in:
```
jwienke@cinnabar ~/.vdirsyncer/status> khal
Traceback (most recent call last):
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/termui.py", line 372, in style
    bits.append('\033[%dm' % (_ansi_colors.index(fg) + 30))
ValueError: tuple.index(x): x not in tuple

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/homes/jwienke/local/all/bin/khal", line 9, in <module>
    load_entry_point('khal==v0.7.1.dev35-ng33c5373', 'console_scripts', 'khal')()
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 1037, in invoke
    return Command.invoke(self, ctx)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/khal-v0.7.1.dev35_ng33c5373-py3.4.egg/khal/cli.py", line 227, in cli
    ctx.invoke(cli.commands[command])
  File "/homes/jwienke/local/all/opt/khal/lib/python3.4/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
[...]

PR #322, opened 2016-01-09
[WIP] enhanced colour support

black2754:
I found it a bit limiting that khal can use only 16 colours when modern terminals often support 256 or even 24-bit colours.
So, I hacked at it and implemented support for more colours.
Unfortunately, `click` does not seem to support more than 16 colours. Instead of relying on `click` for colours, I changed the `colored` function and use it instead of `click.style`. As a positive side-effect, colouring is now more consistent, as it uses only one function.
In addition to implementing support for more colours, I also changed handling of the 16 "base" colours a bit. I believe this provides a better fix for issue #314.
Apart from that, my changes should be fully compatible with existing config files and code.

The new colours can be specified similar to colours in `git`: 0-255 for colours from the 256-colours palette and HTML-style colours for 24-bit colours. Does this seem sensible to you?

My work is not complete, yet. I am aware of the following open issues:

* Clearly, this needs more testing. I always forget something...
* I implemented support for using bright "base" colours without using the terminal's bold attribute. This has the advantage, that backgrounds can be coloured brightly, too. However, the ANSI codes for non-bold bright colours may not be supported by all terminals. The best solution seemed to make this configurable. However, I could not figure out how to access configuration options from within the `colored` function. I'd be grateful for any hints.
* The documentation needs to be updated, of course. I'd rather leave this to someone more familiar with the project.

So, what do you think? Is this useful? Are there any issues I forgot?


hobarrera:
The [vdir spec][spec] supports adding rgb hex-encoded colours in a `colour` file inside the vdir (we haven't been supporting this yet). I do believe that this is a good chance to add support for that, by:

 * Reading from that file rather than the config file.
 * Updating the docs to reflect this.

I'm okay with this PR as-is though, if others feel that that should be implemented separately.

Thanks for this, by the way, it's a very welcome change!

[spec]: https://vdirsyncer.readthedocs.org/en/latest/vdir.html#metadata

black2754:
I thought about getting the calendar colour from the vdir, too. It seems like a logical next step to me. However, it is not trivial to implement. Therefore, I think it is best to focus on one step at a time.

geier:
Hello and welcome to khal! This is certainly useful and I don't mind getting rid of click's style. If I see this correctly, the extended palette currently doesn't work in ikhal/urwid, does it? If so, we should have a look if something can be done about that.

But I really need to install a terminal which supports more colors to properly test it first.

black2754:
You are right, apparently I missed ikhal completely. I will look into that.

black2754:
Unfortunately, `urwid` does not support 24-bit colour. So I only implemented 256 colour support. 24-bit colours are converted to hopefully similar palette colours.

The current approach of using colours as attributes for styling the interface does not scale for 256 colours. So I switched to using the calender names as attributes instead. They are generated at run-time with the correct colours.
I *think* I found all places where colours were used as attributes. But maybe I didn't. Is there any way to make `urwid` *not* silently ignore undefined attributes? That would help tremendously.

On a side note: `ikhal` currently does not support grey colours, as the `khal` configuration calls it `grey` while `urwid` calls it `gray`.
I resolved this by adding a simple replacement rule.
However, using the American spelling "gray" might be better for consistency with the American spelling "color". But that is not for me to decide and I did not want to break existing config files with this PR.

geier:
Looks good too me, could anybody else have a look if this works for them? @dzoep or @languitar perhaps?

@black2754 Two more things: 1) Could you add yourself to `AUTHORS.txt` please?
2) the config spec file (from which the documentation is generated) does mention the "named colors" but does not specify them, could you put them into that "doc" string?

Also let's use the American spelling then, I think it's ok to break compatibility here, I'll put a note in the CHANGELOG. We could of course use both for the time being, issue a deprecating warning for BE spelling and so on, but I think that it's really worth it.

khal/ui/__init__.py
```diff
@@ -732,13 +732,112 @@ def cleanup(self, data):
...
+            # wrong (in the sense of human color perception). However, any
+            # serious color distance metric would be way more complicated.
+            dist = (r - rgb[0]) ** 2 + (g - rgb[1]) ** 2 + (b - rgb[2]) ** 2
+            if best is None or dist < bestdist:
+                best = index
+                bestdist = dist
+        color = 'h' + str(best)
+    # We unconditionally add the color to the high color slot. It seems to work
+    # in lower color terminals as well.
+    if hmethod == 'fg' or hmethod == 'foreground':
```
hobarrera: `if hmethod in ['fg', 'foreground']:`

black2754:
I changed the spelling of grey and updated the documentation in `khal.spec`. Is this what you meant?

I also made `khal` actually use the new config option I introduced earlier. I changed seven function definitions just to have that option available in one function. This seems somewhat inelegant and like a lot of overhead to me. Is there a better way to do this?

Anyway, I think this PR is complete now. It could probably still use more testing and I'd be happy to fix any issues that might come up.

Thank you both for all your feedback.

dzoep:
Haven't tried it yet, but the code looks much cleaner than how it looked when I hacked the initial color support in :). I'd definitely merge it.

geier:
@black2754 yep, that's what I meant (regarding the doc).

This could be made a lot prettier I agree.

Thank you for the PR again!

Issue #251, opened 2015-08-20
week view


geier:
provide a week view for ikhal or khal

see https://github.com/geier/khal/issues/65#issuecomment-132464366 and https://github.com/geier/khal/issues/148#issuecomment-133052856

pull requests are very welcome

mathstuf:
Might I suggest a "help-wanted" label for issues like this?

geier:
Thanks for that suggestion.

libre-man:
How would you imagine a week view for khal? Similar to `khal agenda/caledar --days 7` but for monday to sunday or something different?

protist:
@libre-man There is a screenshot [here](https://github.com/geier/khal/issues/65#issuecomment-75904814). I imagine tiling the screenshot's daily pane across horizontally seven times. The key for me is to have proportional times (unlike the agenda view), as per the other thread. Monday to Sunday makes most sense to me, but I think some people consider a "week" to be Sunday to Saturday, so perhaps a config is necessary.

libre-man:
@protist Ok I got a clear picture of it now. Proportional times would be nice to
have in more functions than just a weekview. The `agenda` function would benefit
from it too. So it could be an idea to make a function that parses events lists
and it could be used in the get_agenda function in controllers.py.
An other idea is to make a new function all together that implements the
proportional time and make --week a flag. However this would mean duplicate code
or a bit of reshuffling of code. But I think this wouldn't be such a bad thing
as the get_agenda function has gotten quite long.

I'll try somethings soon to find out what is the easiest way.

protist:
@libre-man Great! Sounds good. Looking at it again, the screenshot was quite wide. i.e. on a 1920 x 1080 pixel screen, you could probably only fit ~4 days across. My Android calendar (Business Calendar 2) has the additional option of only displaying a user-specified number of days across, instead of a full seven. I'm unsure if this is something that is useful, and/or would fit in better with your second suggestion.

[...]

PR #324, opened 2016-01-11
Introduce --week option for calendar and agenda view.

stweller:
This pull request is a first step towards #251. It provides a simple --week option for calendar and agenda, so you can view a whole week at once. It uses the European convention (Mo-Su), since this seems to be used everywhere in python.

I am not very familiar with github pull requests, so sorry if I made some errors.


geier:
Hello and thank you for your PR. I would suggest not starting the week on Monday but on the set `firstweekday` (in the locale settings). Otherwise this looks good to me, but I haven't tested it yet (will do later today/this week).

stweller:
I changed the code to use firstweekday as you suggested. I also fixed the formatting problems, linthub seems to be happy now.

stweller:
Just one more comment: At the moment, the --week switch leads to behavior that ignores --days. It might be good to issue a warning if both switches are present. Unfortunately I am not very familiar with click and do not know where to place this check.

geier:
We can probably even make them exclusive, but I'll have to look that up.

Quoting Stephan Weller (2016-01-11 12:40:04)

geier:
Sorry it took me such a long time to get back to you.

I cannot figure out how to properly make those two arguments exclusive with click, so I just added an Error the stupid way. I also [rebased your branch onto current master](https://github.com/geier/khal/commits/master_stweller). You can either take the latest commit from there (or just copy the change) and rebase yourself or just let me commit it as it is.

Please add yourself to `AUTHORS.txt` either way.

Thank you for your interest in khal, your contribution and patience!

stweller:
Ok, now I was the one to be a bit low. I rebased the branch again, squashed some commits, and updated the PR.

khal/controllers.py
```diff
@@ -143,9 +149,9 @@ def calendar(collection, dates=None, firstweekday=0, encoding='utf-8', locale=No
     lwidth = 25
     rwidth = term_width - lwidth - 4
     event_column = get_agenda(
-        collection, locale, dates=dates, width=rwidth,
-        show_all_days=show_all_days, full=full, bold_for_light_color=bold_for_light_color,
-        **kwargs)
+        collection, locale, dates=date, width=rwidth,
```
geier: I believe this really should be `dates=dates`

khal/controllers.py
```diff
@@ -96,6 +96,11 @@ def get_agenda(collection, locale, dates=None, firstweekday=0, days=None, events
             logging.fatal(error)
             sys.exit(1)
 
+    if week:
+        dates = [d - datetime.timedelta((d.weekday() - firstweekday) % 7)
```
geier: this probably should be `timedelta` instead of `datetime.timedelta` (we don't import datetime but do `from datetime import datetime, date, timedelta`)

geier:
Sorry that it took me again a week :crying_cat_face: Please have a look at my comments.

khal/controllers.py
hobarrera: Indeed. `date` is actually `datetime.date` here.

stweller:
Thanks for your comments. Both issues are fixed now.

Issue #333, opened 2016-02-11
VCARDs with more than one BDAY throws exception


lucc:
I think there are some points in the docs for the `[calendars]` section with `type = birthdays` that need expansion. 

What I gather from my tests with working and not working setups and the error messages are these constrains:

1. `path` must be set to a directory
2. this directory will be scanned for files with the `.vcf` extension
3. other files (extensions) will be ignored. 
4. the search is not recursive
5. a vcard file may not have more than one BDAY entry (see stack trace below)

Are these constrains correct? Did I misunderstand something or forget something.

Especially the combination of 1. and 2. and 3. took me some time to figure out. I think at least this should be mentioned in the docs.

(the trace for 5.
```
Unknown exception happened.
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/khal/khalendar/khalendar.py", line 233, in _update_vevent
    self._dbtool.update(event.raw, href=href, etag=etag)
  File "/usr/lib/python3.5/site-packages/khal/khalendar/backend.py", line 551, in update
    bday = parser.parse(bday).date()
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 1008, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 392, in parse
    res = self._parse(timestr, **kwargs)
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 492, in _parse
    l = _timelex.split(timestr)         # Splits the timestr into tokens
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 174, in split
    return list(cls(s))
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 164, in __next__
    token = self.get_token()
  File "/usr/lib/python3.5/site-packages/dateutil/parser.py", line 82, in get_token
    nextchar = self.instream.read(1)
AttributeError: 'list' object has no attribute 'read'
warning: Skipping birthdays/197.vcf: 'list' object has no attribute 'read'
warning: This event will not be available in khal.
```
)

geier:
You are right in your assumptions and this should really be better documented. While vcard's with more than BDAY are invalid, we should probably at least give a nicer error message.

geier:
As the doc part has been solved, I've renamed this issues.

geier:
[...]

PR #339, opened 2016-02-14
Update docs as discussed in #333.

lucc:


geier:
Thank you, we'll merge this after your other one.

geier:
Sorry, I missed something: please make these changes in `khal/settings/khal.spec` instead, `configspec.rst` gets generated from there on building the docs.

lucc:
Will do.

Why do you store a generated file in git?

geier:
Didn't remember, but Read The Docs couldn't generate that file (or I didn't know how). RTD builds are failing at the moment either way.

Thanks again!

PR #346, opened 2016-02-20
Update config with days and hightlight options

spk:
Just add two nice options to sample config.
(debug did not work for me)
Cheers


geier:
Hi @spk! Thank you for your interest in khal and your PR!

You are totally right, debug doesn't do anything anymore.
I also like your other suggestions, but I'm not sure the format
```
# default 3
days = 30
```
is the best. I think something like

```
days = 30  # the default is 2
```
would be better. Could you update the PR (and please also add yourself to `AUTHORS.txt`) ?

fpytloun:
Nice, didn't know these options :-)

spk:
Hi @geier ! I've updated the PR with your suggestions.
Cheers !

untitaker:
:confetti_ball:

PR #347, opened 2016-02-21
Remove redundant travis exclusion

hobarrera:
We don't test on pypy, so it's pointless to exclude a certain pypy build explicitly.


geier:
Thanks!

PR #349, opened 2016-02-22
Feature easy setup

geier:
some features hopefully making the initial setup for new users easier

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,10 @@ not released yet
   documentation for details (Sebastian Hamann)
 * renamed color `grey` to `gray` (Sebastian Hamann)
 * in `khal new` treat 24:00 as the end of a day/00:00 of the next (Christian Geier)
+* new allowed value for a calendar's color: `auto` (also the new default), if
+  set, khal will try to read a file called `color` from that calendar's vdir (see
+  vdirsyncer's documentation on `metasync`). If that file is not present or its
+  contents is not understood, the default color will be used (Christian Geier)
 
 ikhal
 -----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,9 @@ not released yet
   set, khal will try to read a file called `color` from that calendar's vdir (see
   vdirsyncer's documentation on `metasync`). If that file is not present or its
   contents is not understood, the default color will be used (Christian Geier)
+* new allowed value for calendar's type: `discover`, if set, khal will
+  (recursively) search that calendar's path for valid vdirs and add those to
+  the configured calendars (Christian Geier)
 
 ikhal
 -----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,8 @@ not released yet
 * new allowed value for calendar's type: `discover`, if set, khal will
   (recursively) search that calendar's path for valid vdirs and add those to
   the configured calendars (Christian Geier)
+* new command `configure` which should help new users set up a configuration
+  file (Christian Geier)
 
 ikhal
 -----
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -244,7 +244,7 @@ def _get_cli():
         '''Print calendar with agenda.'''
         controllers.calendar(
             build_collection(ctx),
-            date=dates,
+            dates=dates,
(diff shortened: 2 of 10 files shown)


khal/cli.py
```diff
@@ -185,6 +186,11 @@ def prepare_context(ctx, config):
         ctx.obj['conf'] = conf = get_config(config)
     except InvalidSettingsError:
         sys.exit(1)
+    except NoConfigFile:
+        logger.fatal(
+            'If you have no configuration file yet, you might want to run '
+            '`khal initialconfig`.')
```
hobarrera: ¿Why not just `configure`?

untitaker:
I'm very concerned about the way vdir discovery is done. A proper solution for vdir-discovery would be language independent and provide a stable interface -- using vdirsyncer's internal APIs for reading its configuration file is neither. I also think that khal currently depends too much on internal APIs of vdirsyncer. Yesterday I've tried to factor out the filesystem storage into its own package: https://github.com/vdirsyncer/python-vdir

In my mind there are two ways to provide vdir discovery:

* vdirsyncer exposes a simple CLI command to list all vdir paths.
* There are standard locations for tools to look for vdirs.

The first solution is much more flexible, but unfortunately requires the installation of vdirsyncer for any sensible use of vdirs, while khal would normally be perfectly able to run on its own, without sync.

The second solution comes with the problem that the user no longer explicitly defines where their data is stored. I think that's somewhat problematic from a UX perspective. It's also completely inflexible.

geier:
I agree that using khal for vdir access isn't great and I do welcome the vdir package.

But for this setup helper I do not see much harm. Worst case: user need to configure by hand is is the case now. We might want to wrap it in some more try/except blocks.

My approach here is this: it works (for now), it won't delete any data and lets worry about the future (i.e. when vdirsyncer changes its internals) when its here.

Furthermore I would suggest telling the user the standard place for calendars (which is to be determined) and suggest they use that and have khal's setup function searching that by default, too.

hobarrera:
> There are standard locations for tools to look for vdirs.

> Furthermore I would suggest telling the user the standard place for calendars (which is to be determined) and suggest they use that and have khal's setup function searching that by default, too.

`~/.local/share/calendars` seems like a sensible location for placing calendars, which could be the default.
Maybe the "default" or "recommended" location could also be mentioned in the vdir spec?

geier:
I like `multi` better than `multi-calendar`, even though `multi-calendar` is more explicit

CHANGELOG.rst
```diff
@@ -22,6 +22,15 @@ not released yet
   documentation for details (Sebastian Hamann)
 * renamed color `grey` to `gray` (Sebastian Hamann)
 * in `khal new` treat 24:00 as the end of a day/00:00 of the next (Christian Geier)
+* new allowed value for a calendar's color: `auto` (also the new default), if
+  set, khal will try to read a file called `color` from that calendar's vdir (see
+  vdirsyncer's documentation on `metasync`). If that file is not present or its
+  contents is not understood, the default color will be used (Christian Geier)
+* new allowed value for calendar's type: `collection`, if set, khal will
```
hobarrera: `s/collection/discover/`

khal/cli.py
hobarrera: `s/initialconfig/configure`

khal/controllers.py
```diff
@@ -27,11 +27,12 @@
 from collections import defaultdict
 from shutil import get_terminal_size
 
-import datetime
+import datetime as dt
```
hobarrera: I'd rename variables called `datetime` that shadow this import, rather than `import datetime as`. It will probably make the code more readable to newcomers (or to us in future).

khal/cli.py
geier: thanks, thought I fixed that already, some goes for the other one

khal/controllers.py
geier: renaming all variables called `datetime` to something else is probably a good idea, but I still don't like to `import datetime`, I prefer to either `from datetime import datetime` or `import datetime as dt` so that all `datetime`s are always `datetime.datetime`.

hobarrera: +1 for `from datetime import datetime, date`. It makes everything below much cleaner.

PR #356, opened 2016-02-23
khal and ikhal: startup optimization with highlight_event_days = True

dzoep:
Backend optimization of get_localized and get_floating. Slight optimization in merging events for single day (floating+localized). And a fix for incorrectly highlighted days in ikhal.

The merging function is not necessary (it is only a small improvement over itertools) but it is somewhat faster and opens the possibility for more optimizations like generating whole week using single backend call.


untitaker:
lgtm

untitaker:
I've applied your fix without extra merge commits. Thanks!

PR #357, opened 2016-02-24
Refactor khalendar.expand

untitaker:
- Don't modify the vevent -- I think there was a bug where we strip the
  timezone information from DTSTART, but never add it back.

- Remove special return-path for events without RRULEs


untitaker:
Linthub is for Python 2 code. It doesn't recognize Python 3 syntax.

geier:
I looked at this just now and I think it's okay, but because this is one of the most complex functions in all of khal, I'm going to do it again (hopefully tomorrow) before merging it.

untitaker:
Force-push to update commit status

geier:
looks good, please rebase and merge

PR #358, opened 2016-02-24
WIP: Make events SQL query slightly lighter

hobarrera:
Use BETWEEN to determine if a date is inside a range. This is ligher since SQLite will optimize that. From the BETWEEN documentation:

> The BETWEEN operator is logically equivalent to a pair of comparisons. "x BETWEEN y AND z" is equivalent to "x>=y AND x<=z" except that with BETWEEN, the x expression is only evaluated once. The precedence of the BETWEEN operator is the same as the precedence as operators == and != and LIKE and groups left to right.

### DIFF
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -451,17 +451,7 @@ class SQLiteDb(object):
         start = aux.to_unix_time(start)
         end = aux.to_unix_time(end)
         if minimal:
-            sql_s = (
-                'SELECT events.calendar FROM '
-                'recs_loc JOIN events ON '
-                'recs_loc.href = events.href AND '
-                'recs_loc.calendar = events.calendar WHERE '
-                '(dtstart BETWEEN ? AND ? OR '
-                'dtend BETWEEN ? AND ? OR '
-                '? BETWEEN dtstart AND dtend OR '
-                '? BETWEEN dtstart AND dtend) '
-                'AND events.calendar in ({0}) '
-                'ORDER BY dtstart')
+            sql_s = 'SELECT events.calendar FROM events'
         else:
             sql_s = (
                 'SELECT item, recs_loc.href, dtstart, dtend, ref, etag, dtype, events.calendar '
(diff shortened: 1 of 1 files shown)


khal/khalendar/backend.py
```diff
@@ -451,24 +451,18 @@ def get_localized(self, start, end, minimal=False):
...
-            sql_s = (
-                'SELECT events.calendar FROM '
-                'recs_loc JOIN events ON '
-                'recs_loc.href = events.href AND '
-                'recs_loc.calendar = events.calendar WHERE '
-                '(dtstart >= ? AND dtstart <= ? OR '
-                'dtend > ? AND dtend <= ? OR '
-                'dtstart <= ? AND dtend >= ?) AND events.calendar in ({0}) '
-                'ORDER BY dtstart')
+            sql_s = 'SELECT events.calendar FROM events'
```
hobarrera: Ugh, it looks like this line is **never** executed during tests. Nothing broken when manually testing, but we need to increase coverage.

dzoep:
@hobarrera This just selects all calendars. The original query selects only calendars which have recurrence records in given time range - it is used (for example) for choosing correct colors for days.

BETWEEN would make sense though ...

hobarrera:
> @hobarrera This just selects all calendars. The original query selects only calendars which have recurrence records in given time range - it is used (for example) for choosing correct colors for days.

In that case, I'm convinced that the new query is an improvement, since it's cheaper to get all calendars (3? 5? 10?) than to have sqlite filter across all the events (100? 1000?).

geier:
@hobarrera but we want to know in which calendars there are events during that period (for the coloring)

hobarrera:
Okay, maybe I was too dramatic with @d48a358. Though I can't quite follow from where this code is reached. `ikhal`?

hobarrera:
Actually, we *can't* use `BETWEEN` (or, if we do, it requires further work), since:

`dtstart BETWEEN ? and ?` is equivalent to `dtstart >= ? and ststart <= ?`, but what we need is `dtstart >= ? AND dtstart < ?)`.

Not that the last operator is different, `<` vs `<=`. Again, tests should have revealed that. :(

PR #359, opened 2016-02-25
Codecov

hobarrera:
Add codecov to the project. This show us a fancy UI with detailed test coverage (including line-by-lineview ) for each push/PR.

[Sample](https://codecov.io/github/hobarrera/khal/khal?ref=49523c66ea40e50f9ed2c325382b011a2316369e)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,12 +3,25 @@ envlist = {py33, py34, py35}-{tests, tests-negtz, style}
 skip_missing_interpreters = True
 
 [testenv]
-passenv = LANG
+passenv =
+  LANG
+  CI
+  TRAVIS
+  TRAVIS_BRANCH
+  TRAVIS_BUILD_ID
+  TRAVIS_COMMIT
+  TRAVIS_JOB_ID
+  TRAVIS_JOB_NUMBER
+  TRAVIS_PULL_REQUEST
+  TRAVIS_REPO_SLUG
 deps =
+    codecov
     pytest
+    pytest-cov
     pytest-capturelog
 commands =
-	py.test {posargs}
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -3,12 +3,25 @@ envlist = {py33, py34, py35}-{tests, tests-negtz, style}
...
+  TRAVIS_REPO_SLUG
 deps =
+    codecov
     pytest
+    pytest-cov
     pytest-capturelog
 commands =
-	py.test {posargs}
+    py.test --cov khal {posargs}
+    codecov -e TOXENV
```
untitaker: This is problematic, it will upload coverages from local testruns.

Better to call it from travis.yml.

hobarrera: No, it doesn't upload anything since the upload token is undefined (see the travis-specific variables passed above).
It exits gracefully too, rather than bail with an error making our tests fail.

untitaker: I see.

hobarrera:
@geier: This isn't in-progress it's ready to merge if you're okay with using codecov.

untitaker:
Could we remove codecov's comments? They're extremely annoying. I think only @geier can do that though.

Otherwise :+1:

hobarrera:
Yeah, it *is* pretty ugly, and it doesn't really make use of github's PR statuses.

hobarrera:
The bot that leaves comments is actually disabled, as seen [here](https://github.com/hobarrera/khal/pull/1).

My guess is that is left a comment in geier/khal because the setting only applies to hobarrera/khal (for now).

geier:
Sure, can give this a try.

PR #361, opened 2016-02-25
khal calendar: Strip colors if not isatty

### LINKED ISSUE #338: feature request: 'clean' output for use in conky
Hi,

I've been running khal in conky with the following line:  
`${execpi 36000 khal}`  

Which while it works, gives a lot of garbage on the display. I'd be interested to see a CLI flag for khal that would give a conky-compatable output. Picture attached is how it looks right now (Top) with how it would ideally look (Bottom) - I did this by simply piping the khal output to a text file and manually deleting the extras for the bold/colourful characters.  

Apparently this is an issue gcalcli has too (Which I discovered when searching for a fix) and seems to be a limitation with Conky not supporting the special characters used.

untitaker:
Fix #338

Click already does this work for us, but not on bytes.

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -156,8 +156,6 @@ def calendar(collection, date=None, firstweekday=0, encoding='utf-8', locale=Non
         locale=locale,
         bold_for_light_color=bold_for_light_color)
     rows = merge_columns(calendar_column, event_column)
-    # XXX: Generate this as a unicode in the first place, rather than
-    # casting it.
     echo('\n'.join(rows))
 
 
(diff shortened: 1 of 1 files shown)


hobarrera:
I'd really like to have a `--colour` flag to force colouring (which is useful when grepping).

untitaker:
We can add that later as well -- for now, let's do what the other commands (e.g. agenda) already do.

geier:
While this is a quick fix, I think we should not generate colors in the first place when we don't need them (especially calendar coloring can be very expensive).

hobarrera:
> While this is a quick fix, I think we should not generate colors in the first place when we don't need them (especially calendar coloring can be very expensive).

This makes me wonder if we shouldn't store calendar colours in the db, and somehow generate colours though a db query (which would be quite more efficient).

untitaker:
I don't know if that's sufficiently advantageous to warrant the added complexity. db is necessary due to time-range queries.

untitaker:
Let's do the color stripping for another PR. Click's logic regarding whether to strip colors is quite complex, and I'm not comfortable introducing logic in khal that works in parallel. I'd much rather add an API to click that exposes its decision.

PR #363, opened 2016-02-28
New option: --force-color for calendar and agenda

geier:
Enables coloring even if stdout is not a tty.


geier:
closed in favor favor of #365

PR #369, opened 2016-03-03
WIP: Allow configuring excluded calendars

hobarrera:
Implement a configuration option [default][exclude_calendars] which allows excluding calendars from khal by default.

This is equivalent to always running khal with `d CALENDAR`, so this setting might be considered a default `-d` setting, that DOES NOT clash with `-a`.

I find this very useful for calendar which I want to track but don't want listed all the time (maybe due to lack of importance, or maybe because it has too many events).

### DIFF
--- a/doc/source/configspec.rst
+++ b/doc/source/configspec.rst
@@ -396,6 +396,17 @@ behaviours are set here.
       :type: boolean
       :default: False
 
+.. _default-exclude_calendars:
+
+.. object:: exclude_calendars
+
+    
+    A list of calendars that should be excluded by default when running `khal`.
+    These can be included again explicitly by using the `-a` flag.
+
+      :type: list
+      :default: 
+
 .. _default-default_calendar:
 
 .. object:: default_calendar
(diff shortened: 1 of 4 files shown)


untitaker:
I'd much rather have a default_calendars option instead of the opposite.

hobarrera:
On 2016-03-19 06:28, Markus Unterwaditzer wrote:
> I'd much rather have a default_calendars option instead of the opposite.
> 
> ---
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/pimutils/khal/pull/369#issuecomment-198703868

Yeah, maybe that makes more sense and would have a cleaner implementation.
I'll adapt this PR to do that.

-- 
Hugo Osvaldo Barrera
A: Because we read from top to bottom, left to right.
Q: Why should I start my reply below the quoted text?

hobarrera:
I haven't had time to work more on this, and I prefer @untitaker's approach rather than having this merged (plus, it still lacks tests).

PR #378, opened 2016-03-19
Fix calendar selection when importing events

black2754:
When importing an event, khal asks for the calendar into which the event is to be imported. If a calendar is selected by name, rather than by number, khal crashes with a `TypeError: 'filter' object is not subscriptable`. This happens, in particular, when the user does not enter anything at all in order to use the default calendar.

This PR fixes the issue by replacing the filter with a list comprehension. I'm not sure if this is the best possible fix, but it works and is fairly readable.


untitaker:
Thanks!

Issue #70, opened 2014-05-05
edit duration of events (while adding and in ikhal)


geier:

geier:
durations can be used when added or editing via khal. Good enough for now.

PR #380, opened 2016-03-19
cli option for alarms in new events

sdx23:
With the new alarms feature surely some people (at least me ;)) would like to be able to specify alarms when creating new events using the cli interface. Beware: This also supports negative alarm times, so the alarm can actually be _before_ the event starts.

Also python is not my native tounge, feel free to modify this PR in any way seeming sensible to you. (I'd also appreciate feedback)

### DIFF
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -26,6 +26,7 @@ from calendar import isleap
 from datetime import date, datetime, timedelta, time
 import random
 import string
+import re
 from time import strptime
 
 import icalendar
@@ -210,6 +211,43 @@ def guessdatetimefstr(dtime_list, locale, default_day=None):
     raise ValueError()
 
 
+def guesstimedeltafstr(delta_string):
+    """parses a timedelta from a string
+
+    :param delta_string: string encoding time-delta, e.g. '1h 15m'
+    :type delta_string: str
+    :rtype: datetime.timedelta
+    """
+
+    tups = re.split(r'(-?\d+)', delta_string)
+    if not re.match(r'^\s*$', tups[0]):
+        raise ValueError('Invalid beginning of timedelta string "%s": "%s"'
+                         % (delta_string, tups[0]))
+    tups = tups[1:]
+    res = timedelta()
+
+    for num, unit in zip(tups[0::2], tups[1::2]):
+        try:
+            numint = int(num)
+        except ValueError:
+            raise ValueError('Invalid number in timedelta string "%s": "%s"'
+                             % (delta_string, num))
+
+        ulower = unit.lower().strip()
+        if ulower == 'd' or ulower == 'day' or ulower == 'days':
+            res += timedelta(days=numint)
+        elif ulower == 'h' or ulower == 'hour' or ulower == 'hours':
+            res += timedelta(hours=numint)
+        elif (ulower == 'm' or ulower == 'minute' or ulower == 'minutes' or
+              ulower == 'min'):
+            res += timedelta(minutes=numint)
+        else:
+            raise ValueError('Invalid unit in timedelta string "%s": "%s"'
+                             % (delta_string, unit))
+
+    return res
+
+
 def generate_random_uid():
     """generate a random uid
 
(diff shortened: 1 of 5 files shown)


untitaker:
https://github.com/pimutils/vdirsyncer/issues/154 might be relevant, where we explore options for serializing/unserializing timedeltas.

untitaker:
BTW @sdx23 your commit has some styling issues, see https://travis-ci.org/pimutils/khal/jobs/117118064

sdx23:
Is it really iso8601-timedeltas what you want a user to write down when creating a new event? With the current interface design I followed the input accepted by unix `sleep`. Imo "1h 15m" is much clearer than "PT1H15M".

untitaker:
I agree, perhaps the leading "P" could be made optional... still I see the value in standardizing day and month suffixes, especially as those are wildly varying between programming languages.

geier:
IMO we should be pretty lenient in what we accept for timedeltas in khal (as long as its unambiguous). I personally won't be typing something like `PT1H15M`. `1h 15m` much more natural in my opinion, but we should also support `15 minutes` etc. This discussion is also relevant for #70.

@sdx Could you add some tests (at least for `guesstimedeltafstr()`) and add yourself to `AUTHORS.txt`?

khal/aux.py
```diff
@@ -210,6 +211,27 @@ def datefstr_year(dtime_list, dateformat):
     raise ValueError()
 
 
+def guesstimedeltafstr(delta_string):
+    """
+    :type detimedelta
```
geier: the type should be `str`, right?

sdx23:
Fixed the type annotation (was indeed messed up), added me to authors.txt, added tests.

How do I correctly test the tests? Also, don't we need a `# coding=utf8` line or something?

geier:
To run the tests you need to at least install pytest and freezegun, you can than run `py.test khal/tests` (the excecutable might be called pytest instead of pytest).

We should probably add something to the documentation about how to run the tests.

geier:
re coding:
as khal only runs with python3, we do not need a line like that (utf-8
encoding is the default, perhaps even the only possible one, I'm not
sure).


Quoting sdx23 (2016-03-29 23:09:07)

hobarrera:
Actually, just running `tox` should be enough. Upsides it that it installs
everything into its own virtualenv, and will tests with multiple python
versions, if available.

On 2016-03-29 14:27, Christian Geier wrote:
> To run the tests you need to at least install pytest and freezegun, you can than run `py.test khal/tests` (the excecutable might be called pytest instead of pytest).
> 
> We should probably add something to the documentation about how to run the tests.
> 
> ---
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/pimutils/khal/pull/380#issuecomment-203113303

-- 
Hugo Osvaldo Barrera
A: Because we read from top to bottom, left to right.
Q: Why should I start my reply below the quoted text?

sdx23:
The test can right now not be adapted to fit the changes introduced in #383, as the dict does not allow for duplicate entries :cry:

geier:
I'm not sure if we should accept something like `23foo10min` and interpret it as `10 minutes`, at least not longterm. 

Also, why don't you just make the TRIGGER times all negative?

sdx23:
For `23foo10min` true. I'll have to look into how to fail the whole thing for invalid input (should anyways, I guess).

Regarding negative: I'd rather do it consistent with gui first and fix #386 together with this.

three-comrades:
> The test can right now not be adapted to fit the changes introduced in #383, as the dict does not allow for duplicate entries :cry:

I wouldn't bother. Most test cases in `aux_test.py` are very similar and only differ in two or three lines. For those, #386 made sense. Your test case is a lot longer, so they wouldn't be reduced by much, anyway. :)

sdx23:
The times are now inverted, consistent with #386. Also, I cleaned up all that mess of commits, rebasing and squashing it into one.

tests/aux_test.py
```diff
@@ -96,6 +96,41 @@ def test_time_tomorrow(self):
...
+
+    def test_multi(self):
+        assert timedelta(days=1, hours=-3, minutes=10) == \
+            guesstimedeltafstr(' 1d -3H 10min ')
+
+    def test_multi_nospace(self):
+        assert timedelta(days=1, hours=-3, minutes=10) == \
+            guesstimedeltafstr('1D-3hour10m')
+
+    def test_garbage(self):
```
geier: you could instead do:

```python
with pytest.raises(ValueError):
    guesstimedeltafstr('10mbar')
```

sdx23: done

geier:
@sdx23 Thanks!

PR #383, opened 2016-03-21
Clean aux_test.py a bit.

three-comrades:
Hey,

when trying to understand a failing test I thought it would be useful to reorganize the file and use a default `vevent` - IMO this reduces the noise in the test definitions.

Edit: Ooops,  this does of course not work, because **kwargs is not an OrderedDict.


geier:
Thanks for cleaning this up. Please add yourself to `AUTHORS.txt`, I'll merge this afterwards.

hobarrera:
👍 Thanks for this.

geier:
Thank you!

PR #387, opened 2016-03-30
Don't require config for configure command

### LINKED ISSUE #384: `khal configure` does not work without a configuration file.
Starting khal without a configuration file prints:

> critical: Cannot find a config file.
> critical: If you have no configuration file yet, you might want to run `khal configure`.

Calling `khal configure` then prints:
> critical: Cannot find a config file.
> critical: If you have no configuration file yet, you might want to run `khal configure`

I would have expected a configuration wizard to be started.
Edit: This is on current master.

untitaker:
Fix #384

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -188,26 +188,29 @@ def build_collection(ctx):
     return collection
 
 
+class _NoConfig(object):
+    def __getitem__(self, key):
+        logger.fatal(
+            'Cannot find a config file. If you have no configuration file '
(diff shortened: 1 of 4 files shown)


untitaker:
@three-comrades Could you try whether this fixes your issue?

Also I think we need a testcase...

three-comrades:
This fixes the issue for me, thanks! Also, interesting to see the test case, because I had no idea how to implement that.

untitaker:
cool

PR #392, opened 2016-04-04
Fix 373

### LINKED ISSUE #373: ikhal: pageup/pagedown leave traces behind
Independently of the color scheme, when I use ikhal and immediately press pageup/down, the current selected color of the day is not cleared. If I move over it using the arrow keys it gets cleared, so it might just be some state to reset.

geier:
fix #373


hobarrera:
If I click on empty space between days, the events list does not update (as expected), but the selected day is un-highlighted.

geier:
good catch

hobarrera:
> good catch

Thank my poor aim when clicking.

PR #400, opened 2016-04-14
Echo unicode string, such that --color works

### LINKED ISSUE #399: piping output from search includes color codes
I don't know when this was introduced, but the other day I piped some content from the search and it didn't include the color codes. However I have updated khal to the latest dev version and now I noticed that when I pipe the search to different tools the escape codes are included.

Is this intentional or is this a regression. If it is intentional perhaps we should add a `--no-color` option to khal.

untitaker:
Fix #399


untitaker:
This obliterates the `encoding` setting in the `[locale]` section as far as I can tell, because Click automatically figures out the wished encoding. Any objections to removing it?

geier:
@untitaker no, let's get rid of it, it wasn't applied consistently everywhere anyway (please put a note into the changelog)

untitaker:
Ready to review @geier

geier:
Thanks!

PR #401, opened 2016-04-17
Ignore dotfiles when for `discover`-type paths

### LINKED ISSUE #375: discover lists bogus directories as calendars
Wow, khal moved to `/pimutils` just as I posted the issue a moment ago, and the issue disappeared!

---

I've this config:
```
[calendars]
 [[calendars]]
  path = ~/.local/share/calendars/
  type = discover
```

Since `~/.local/share/calendars` is a git repository, khal lists all subdirectories in `~/.local/share/calendars/.git` as additional calendars - that's dozens of subdirectories inside ikhal's calendar picker.

Maybe dotfile-directories should be ignored when using discover?

hobarrera:
Fixes #375

### DIFF
--- a/khal/settings/utils.py
+++ b/khal/settings/utils.py
@@ -142,13 +142,15 @@ def get_unique_name(path, names):
 
 def get_all_vdirs(path):
     """returns (recursively) all directories under `path` that contain
-    only files (not directories).
+    only files (not directories) and are not dotfiles.
     """
     # TODO take care of links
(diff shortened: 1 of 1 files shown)


geier:
If you find the time, could you please:
- [ ] create a test
- [ ] update the doc
- [ ] update the changelog

youngmoney:
I'm closing this as we just merged using glob instead of listdir, which has the same effect

PR #403, opened 2016-04-18
Warn user trying to install on python < 3.3.

geier:
### TODO
- [x] update docs


hobarrera:
I'd also add a link to the docs to a page describing this a bit more, eg:

 * py2 and py3 can co-exist.
 * If installing via pip, use `pip3`
 * If installing via source, use `python3 setup.py install`

geier:
@hobarrera good idea, will do that later.

PR #408, opened 2016-04-25
Use pkg_resources instead of raw __version__

### LINKED ISSUE #406: icalendar>=3.9.2 required for testsuite
There's this test in `tests/backend_test.py` that uses `__version__`, which was only introduced in `icalendar==3.9.2``(https://github.com/collective/icalendar/commit/86e4106a3b7a6ec9d602084e40d33eb4f78c9c70#diff-4913494b4cdb8751d12e71febe0a243c) to check for versions older than 3.9.2.

Several options:

* Fix the test to use `pkg_resources` such that it actually works with older versions
* Just drop support for older icalendar versions

untitaker:
Fix #406

### DIFF
--- a/tests/backend_test.py
+++ b/tests/backend_test.py
@@ -2,6 +2,8 @@
 import pytest
 import pytz
 
+import pkg_resources
+
 from datetime import date, datetime, timedelta, time
 import icalendar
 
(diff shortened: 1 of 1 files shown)


geier:
thank you, @untitaker

PR #410, opened 2016-05-02
Creation of command "list"

youngmoney:
This work in progress pull request is to illustrate where I am with the list command, currently it supports lots of formatting and grabs the correct events, however there are date/datetime problems:

1.  It cannot accept times (this is crucial for functionality that myself and others want)

2. events need to accept a range of times to fully understand how to format themselves

3. I would like to factor out a lot of date range/datetime if else statements into their own object

This request also contains a change to the event less than function allowing it to work across days

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -89,6 +89,131 @@ string *now*. ``at`` defaults to *now*.
 
         khal at [-a CALENDAR ... | -d CALENDAR ...] [DATETIME | now]
 
+list
+****
+shows all events scheduled for a given date (or datetime) range, with custom
+formatting
+
+::
+        khal list [-a CALENDAR ... | -d CALENDAR ...] [--format FORMAT]
+        [--once] [--notstarted] [START DATE] [END DATE]
+
+if no end date is given, midnight of the start date is assumed. If no start
+date is given the today is assumed. The `--once` option only allows event to
+appear once even if they are on multiple days. The `--notstarted` option only
+shows events that start after `START DATE`. `END DATE` can also be of the form
+`I{m,h,d}` where `I` is an integer and `m` means minutes, `h` means hours, and
+`d` means days, this is the offset from the start date. In addition `eod` can
+be used to specify the end of day of the start date, this is useful if you have
+set a timedelta in the config.
+
+The format is a string that will be used with python's `string.format()`. The
+available templates are:
+
+.. option:: title
+
+        The title of the event.
+
+.. option:: description
+
+        The description of the event.
+
+.. option:: start
+
+        The start datetime in datetimeformat.
+
+.. option:: start-long
+
+        The start datetime in longdatetimeformat.
+
+.. option:: start-date
+
+        The start date in dateformat.
+
+.. option:: start-date-long
+
+        The start date in longdateformat.
+
+.. option:: start-time
+
+        The start time in timeformat.
+
+.. option:: end
+
+        The end datetime in datetimeformat.
+
+.. option:: end-long
+
+        The end datetime in longdatetimeformat.
+
+.. option:: end-date
+
+        The end date in dateformat.
+
+.. option:: end-date-long
+
+        The end date in longdateformat.
+
+.. option:: end-time
+
+        The end time in timeformat.
+
+.. option:: recurse
+
+        A repeating symbol (loop arrow) if the event is repeating.
+
+.. option:: description
+
+        The event description.
+
+.. option:: location
+
+        The event location.
+
+.. option:: calendar
+
+        The calendar name.
+
+.. option:: calendar-color
+
+        Changes the output color to the calendar's color.
+
+.. option:: start-style
+
+        The start time in timeformat OR an appropriate symbol.
+
+.. option:: to-style
+
+        A hyphen "-" or nothing such that it appropriatly fits between
+        start-style and end-style.
+
+.. option:: end-style
+
+        The end time in timeformat OR an appropriate symbol.
+
+.. option:: start-end-time-style
+
+        A concatenation of start-style, to-style, and end-style OR an
+        appropriate symbol.
+
+.. option:: start-date-once
+
+        The start date, so long as that date has not yet been printed.
+
+.. option:: start-date-once-newline
+
+        A newline if start-date-once is not "".
+
+In addtion there are colors: `black`, `red`, `green`, `yellow`, `blue`,
+`magenta`, `cyan`, `white` (and their bold versions: `red-bold`, etc.). There
+is also `reset`, which clears the styling.
+
+For example the below command with print the title and description of all events today.
+
+::
+
+        khal list --format "{title} {description}"
+
 calendar
 ********
 shows a calendar (similar to :manpage:`cal(1)`) and agenda. ``khal calendar``
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -23,3 +23,4 @@ Laurent Arnoud - laurent [at] spkdev [dot] net - http://spkdev.net/
 Julian Mehne
 Stephan Weller
 Max Voit - max.voit+dvkh [at] with-eyes [dot] net
+Taylor L Money - - http://taylorlmoney.com
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ not released yet
 
 * removed configuration variable `encoding` (in section [locale]), the correct
   locale should now be figured out automatically (Markus Unterwaditzer)
+* added command `list`
 
 0.8.1
 =====
--- a/doc/source/configspec.rst
+++ b/doc/source/configspec.rst
@@ -431,6 +431,17 @@ behaviours are set here.
       :type: option, allowed values are *event*, *path* and *False*
       :default: False
 
+.. _default-timedelta:
+
+.. object:: timedelta
+
+    
+    The default timedelta to use in the list command if no end is specified.
+    Leave blank for end of day, otherwise a timedelta in the form `I{m,h,d}`.
+
+    :type: string
+    :default: None
+
 The [view] section
 ~~~~~~~~~~~~~~~~~~
 
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -248,6 +248,104 @@ def guesstimedeltafstr(delta_string):
     return res
 
 
+def guessrangefstr(daterange, locale, default_timedelta=None):
+    """parses a range string
+
+    :param daterange: date1 [date2 | timedelta]
+    :type daterange: str or list
+    :param locale:
+    :rtype: (datetime, datetime)
+
+    """
+
+    range_list = daterange
+    if isinstance(daterange, str):
+        range_list = daterange.split()
+
+    try:
+        if len(default_timedelta) == 0:
+            default_timedelta = None
+        else:
+            default_timedelta = guesstimedeltafstr(default_timedelta)
+    except:
+        default_timedelta = None
+
+    for i in range(1, len(range_list) + 1):
+        start = ' '.join(range_list[:i])
+        end = ' '.join(range_list[i:])
+        try:
+            if start is None:
+                start = datetime_fillin(end=False)
+            elif not isinstance(start, date):
+                split = start.split(" ")
+                start = guessdatetimefstr(split, locale)[0]
+                if len(split) != 0:
+                    continue
+
+            if end is None or len(end) == 0:
+                if default_timedelta is not None:
+                    end = start + default_timedelta
+                else:
+                    end = datetime_fillin(day=start)
+            else:
+                if end.lower() == 'eod':
+                    end = datetime_fillin(day=start)
+                else:
+                    try:
+                        delta = guesstimedeltafstr(end)
+                        end = start + delta
+                    except:
+                        split = end.split(" ")
+                        end = guessdatetimefstr(split, locale, default_day=start.date())[0]
+                        if len(split) != 0:
+                            continue
+                    end = datetime_fillin(end)
+            return start, end
+        except Exception:
+            pass
+
+    return None, None
+
+
+def datetime_fillin(dt=None, end=True, locale=None, day=None):
+    """returns a datetime that is filled in (with time etc)
+
+    :param dt:
+    :type dt: datetime or date or time if None then day is used
+    :param end:
+    :type end: boolean set True if time.max should be used (else min)
+    :param locale:
+    :type locale: if set the time will be in this locale
+    :param day:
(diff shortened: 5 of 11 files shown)


geier:
Good to see that you are making progress. Do you need any help?

Perhaps you could split what you have into several patches/PRs which we could integrate faster (and which are also easier to review). `format` would probably be a good candidate for that. Also, please write some (simple) tests for new/changed features and run flake8 over your code.

Youngmoney:
I don't think I need any help, I am just busy the next week or so. Hopefully after that I can finish this off. I would like to break it into two different PRs, one for the updates to the even object, which is `format` and `__lt__` and another that would follow that sets up the list command. I noticed that github is saying that I am failing the travis builds, is that something I should look into? nothing I run locally is having any problems I'm aware of

geier:
Yes, some tests are failing, which might be expected in an early stage and you probably should run `flake8` over the code. Please have a look at the [documentation](https://khal.readthedocs.io/en/latest/contributing.html#hacking), which is very much work in progress. If something isn't clear or could be more helpful please point it out (or submit a PR).

Youngmoney:
so, I have finished all the fixes I wanted to make, now I am trying to prepare this for a pull request. Because my second round of fixes is small, I'm hoping one pull request will the list and format together is fine. I have some question about testing...well more I'm just confused. I don't do any real python work (with tests and all) so I have no idea how this testing suite works. I tried installing tox and running it but that errored out without running any tests.

geier:
Sadly, tox doesn't work very reliable for me either. You can install
pytest instead and than run `pytest tests` (if you install from pyPI
it's called `py.test`.

Additionally, install `flake8 --max-line-length=100` and fix any errors.

Quoting Youngmoney (2016-05-07 00:24:29)

Youngmoney:
I'm a bit lost as to why I have triggered so many failed tests, from what I can see (through my diff with master) is that I have only added new self contained functions with the exception of the change of `__lt__` in the event class, and when I change that back I still fail the same tests

geier:
I'll have a look later today. 

Am 7. Mai 2016 02:05:34 MESZ, schrieb Youngmoney <notifications@github.com>:
>I'm a bit lost as to why I have triggered so many failed tests, from
>what I can see (through my diff with master) is that I have only added
>new self contained functions with the exception of the change of
>`__lt__` in the event class, and when I change that back I still fail
>the same tests
>
>---
>https://github.com/pimutils/khal/pull/410#issuecomment-217590665

-- 

Youngmoney:
found the problem. I named the command list, `def list()` was causing problems so renamed it. I am now only failing style. Will try to update style and write changelog/authors.txt soon

Youngmoney:
I have now updated authors, the changelog, fixed all style, and added a test. Let me know if there is anything else I need to do. I would love to see this in master soon!

doc/source/usage.rst
```diff
@@ -89,6 +89,95 @@ string *now*. ``at`` defaults to *now*.
...
+
+.. option:: start-date-once-newline
+
+        A newline if start-date-once is not "".
+
+For example the below command with print the title and description of all events today.
+
+::
+
+        khal list --format "{title} {description}"
```
geier: This example doesn't work for me, I get:

```
khal list --format "{title} {description}" today                                               ~/workspace/khal
Traceback (most recent call last):
  File "/home/cg/.virtualenvs/khal/bin/khal", line 9, in <module>
    load_entry_point('khal==0.8.2.dev20+ngd64a1e3.d20160425', 'console_scripts', 'khal')()
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/home/cg/.virtualenvs/khal/lib/python3.4/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/home/cg/workspace/khal/khal/cli.py", line 327, in klist
    locale=ctx.obj['conf']['locale'],
  File "/usr/home/cg/workspace/khal/khal/controllers.py", line 263, in khal_list
    end=end, once=once, nostarted=nostarted, **kwargs)
  File "/usr/home/cg/workspace/khal/khal/controllers.py", line 187, in get_list_from_str
    start = aux.guessdatetimefstr(start.split(" "), locale)[0]
  File "/usr/home/cg/workspace/khal/khal/aux.py", line 211, in guessdatetimefstr
    raise ValueError()
ValueError
```

geier:
Nice work so far!

Before having a look at the implementation, I do have one conceptual question: what differentiates the new ``list`` command from ``at`` and ``agenda``? I would like to remove (or at least deprecate) at least one of them with the new ``list``.

Also, could you squash this into two or three topical commits, e.g. "\_lt\_ fix", "--format" and "list"?

Youngmoney:
the fact that list is failing there is weird, I have it working on my end with a copy-paste of what you put in. does it work if you leave it blank (no "today")?

I believe this can replace both at and list. I just added an option to make that more true: the end argument can now be one of +INT{d,m,h} where +5m means start + 5 minutes, and the same for days or hours. This will allow for at and agenda to be almost the same (from what I understand/use them for) (i pushed this change).

I will rebase later today, I have a weird commit history will pulls in there and some mixed commits

Youngmoney:
I'm really very confused as to why its not working for you, I would say try to see why the calls to `aux.guessdatetimefstr` are not accepting today, as it should be the final check in the loop there.

Youngmoney:
i have squashed, let me know if you bug is still occurring I can try to look into it on my end, but like I said its working for me so Im a bit confused

geier:
Works now, thanks.

Would you say this ``khal list 10.05.2016 12:30 +2d`` is supposed to work?

Youngmoney:
this `khal list "10.05.2016 12:30" +2d` should work, I have each date as a separate argument instead of trying to parse it into its parts like agenda does. Im more for this option because it makes more sense in a unix way, i could also see us having a distinct separator such as `..` (which gcalcli uses) to separate start and end, but I think that is a different discussion.

Youngmoney:
before you look too much into he implementation I wanted to say that now that I have more time I am going to go through and try to clean it up a bit and also add some format options for calendar and calendar color if possible, as well as a config option for default formatting

Youngmoney:
I have done the refactoring I want to do, I looked into the calendar but the names I can get from "event" is the wrong name (not the display name), so I have no added the calendar name.

geier:
Here are some notes (in no particular order) on your PR (I cannot give them
inline, as I'm on a train and internet connection is notoriously bad on German
trains). Please note, that while I have a lot of comments and improvement
suggestions, it's because you are touching two very core components of khal,
both of which I would like to become the main ways of interacting with khal.


Conceptual remarks on `list`
============================
  
* with the current implementation we cannot replace list AFAIK, because `khal
  list DTSTART` will always print all events on the day given by DTSTART. Yet
  making `list` print only events which are scheduled for DTSTART unless an
  explicit end datetime is given is ugly from some (my) user scenario
  (but perhaps we should just keep `at` and be a bit less ambitious about
  replacing both `agenda` and `at` with `list`)

* I really would like to see a default timedelta (configurable in the
  configuration file), because a) because my general usage is looking for events
  scheduled for today and tomorrow and I don't wont to type `khal list today
  +1d` every time I could just type `khal list`, and it's what
  calendar(1) does (shows events for today and tomorrow by default)

* `khal list Tomorrow 10:30` should work without quotation marks,
  (and be interpreted as start='tomorrow 10:30') one might want to argue
  that only having optional arguments at the end of an argument list is
  normal UNIX behaviour, but a) I think it's still pretty clear what
  `khal list Tomorrow 10:30` is supposed to mean, b) it's better to be
  consistent than to be right, and c) IMO `khal list 'Today 12:30'
  'Tomorrow 16:30'` is just plain ugly

  In case you, or other people strongly disagree here, we can a) make
  both approaches work or b) introduce (optional) --start DATETIME and
  --end DATETIME options.

* I like the current `agenda` output, where it display's a line for every new
  day (with its date in it), this behaviour could probably be introduced with an
  option, but this is not priority and I'll accept the PR without it

Other remarks
=============

* it's currently too easy to see a stacktrace as a user, please wrap those place
  with a try-except-log-exit, some examples:
  * khal list todayx
  * khal list today 1days
  * khal list today --format '{titlex}'


* regarding: `--nostarted`, shouldn't it be `--notstarted`? And I believe the
  help would be better with an added `only`: "Print only events which have not
  started", btw. I really like that option, I see myself aliasing something like
  `khal list now +1d --notstarted` to `khal next`. Perhaps we can support
  aliases like git does, @untitaker is this something click is currently capable
  of?

* could you use guesstimedeltafstr for timedelta parsing? As both versions are
  probably not optimal yet, I would be a waste to keep improving both

* some properties `.format()` probably should support (but we can always add
  these later):
  * start-long, end-long (not sure about the names)
  * color (the color of the calendar this event belongs to)
  * calendar (even if it's not the display-name yet)

* the 'XXX generate as unicode in the first place' comment can be removed, this
  is unicode in all cases, at the very least since we are python 3 only (I know
  there is another comment like it, but it makes no sense either)

* please also run `pep8 --ignore=E501` on your PR (there are some operators with
  missing whitespace), I'm not sure why flake8 doesn't catch that

* no single line `ifs` please, e.g. if not "seen-days" in env: env["seen-days"] = set()

If you make this two PR (list and format) I believe we can merge format a lot
earlier and can remove `event_description`.

That's it for now...

Youngmoney:
So I think I've fixed these all:

On Tue, May 17, 2016 at 02:38:42AM -0700, Christian Geier wrote:
>Here are some notes (in no particular order) on your PR (I cannot give them
>inline, as I'm on a train and internet connection is notoriously bad on German
>trains). Please note, that while I have a lot of comments and improvement
>suggestions, it's because you are touching two very core components of khal,
>both of which I would like to become the main ways of interacting with khal.
>
>
>Conceptual remarks on `list`
>============================
>
>* with the current implementation we cannot replace list AFAIK, because `khal
>  list DTSTART` will always print all events on the day given by DTSTART. Yet
>  making `list` print only events which are scheduled for DTSTART unless an
>  explicit end datetime is given is ugly from some (my) user scenario
>  (but perhaps we should just keep `at` and be a bit less ambitious about
>  replacing both `agenda` and `at` with `list`)

* In regards to replacing both at and agenda, I think at is a special
  command but I think we could just make it a wrapper for list very
  easily, so at least supporting it would be simple. In addition it
  would be great to see all commands use the list functions.

>
>* I really would like to see a default timedelta (configurable in the
>  configuration file), because a) because my general usage is looking for events
>  scheduled for today and tomorrow and I don't wont to type `khal list today
>  +1d` every time I could just type `khal list`, and it's what
>  calendar(1) does (shows events for today and tomorrow by default)

* timedelta now exists in config under default.timedelta
>
>* `khal list Tomorrow 10:30` should work without quotation marks,
>  (and be interpreted as start='tomorrow 10:30') one might want to argue
>  that only having optional arguments at the end of an argument list is
>  normal UNIX behaviour, but a) I think it's still pretty clear what
>  `khal list Tomorrow 10:30` is supposed to mean, b) it's better to be
>  consistent than to be right, and c) IMO `khal list 'Today 12:30'
>  'Tomorrow 16:30'` is just plain ugly

* daterange is now accepted as a serious of words and they are parsed,
  along with attempting to use the timedelta, in addition the option for
  the end to just be `eod` is available (this is to override any
  timedelta defaults and see normal day behaviour)
>
>  In case you, or other people strongly disagree here, we can a) make
>  both approaches work or b) introduce (optional) --start DATETIME and
>  --end DATETIME options.
>
>* I like the current `agenda` output, where it display's a line for every new
>  day (with its date in it), this behaviour could probably be introduced with an
>  option, but this is not priority and I'll accept the PR without it
>

* newline/header is possible: `--format
  "{start-date-once-newline}{start-date-once}{start-date-once-newline}{title}
  ..."`, however it might be nice in the future to support some sort of
  "pretty date" for this, but that should be in the config as a
  datestring and such and I did not want to complicate the number of
  date options.

>Other remarks
>=============
>
>* it's currently too easy to see a stacktrace as a user, please wrap those place
>  with a try-except-log-exit, some examples:
>  * khal list todayx
>  * khal list today 1days
>  * khal list today --format '{titlex}'
>

* stacktraces are pretty

>
>* regarding: `--nostarted`, shouldn't it be `--notstarted`? And I believe the
>  help would be better with an added `only`: "Print only events which have not
>  started", btw. I really like that option, I see myself aliasing something like
>  `khal list now +1d --notstarted` to `khal next`. Perhaps we can support
>  aliases like git does, @untitaker is this something click is currently capable
>  of?

* nostarted is notstarted
>
>* could you use guesstimedeltafstr for timedelta parsing? As both versions are
>  probably not optimal yet, I would be a waste to keep improving both

* guesstimedeltafstr is used
>
>* some properties `.format()` probably should support (but we can always add
>  these later):
>  * start-long, end-long (not sure about the names)
>  * color (the color of the calendar this event belongs to)
>  * calendar (even if it's not the display-name yet)

* `.format()` supports: calendar, calendar-color, start-long, end-long,
  start-date-long, end-date-long.
>
>* the 'XXX generate as unicode in the first place' comment can be removed, this
>  is unicode in all cases, at the very least since we are python 3 only (I know
>  there is another comment like it, but it makes no sense either)
>
>* please also run `pep8 --ignore=E501` on your PR (there are some operators with
>  missing whitespace), I'm not sure why flake8 doesn't catch that
>
>* no single line `ifs` please, e.g. if not "seen-days" in env: env["seen-days"] = set()

* these things fixed
>
>If you make this two PR (list and format) I believe we can merge format a lot
>earlier and can remove `event_description`.

* I would love to, I'm no sure I know how to (first because I'd have to
  look it up), but also because I'm not sure how to untangle them, list
  is heavily dependent on the event.format, and if I change things there
  then I'll have to pull them back and such. if we need to move on one
  at a time then I'll make an event.format, and after that is merged
  make a list request

geier:
Thank you for fixing everything I brought up and the big improvement you made to khal!

PR #411, opened 2016-05-03
Fix building man pages

Duncanla:
Commit 7edaf9604ff246b9d201b69ac7297e44606a5a0b broke the man page build. For some reason, man.rst was added to the list of excluded source files used for building the documentation. This reverts that change.


geier:
I wanted to get rid of a warning when building the documentation and the man page still built for me because it was still in ``built/``. Thanks for the PR and fix!

PR #412, opened 2016-05-04
New widget FocusLineBox

geier:
This new widget will change its border if in focus or not. There are currently two alternatives, one where the focused widget has a double lined border and one with a thick border. Which one do you like better? (see `FocusLineWidget_thick` for the other version).
2016-05-05-010636_652x530_scrot
2016-05-05-010714_652x530_scrot

**Edit:** Obviously the exact layout will depend on your font, terminal color scheme etc.

### DIFF
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -27,7 +27,8 @@ import urwid
 
 from .. import aux
 from . import colors
-from .widgets import ExtendedEdit as Edit, NPile, NColumns, NListBox, Choice, AlarmsEditor
+from .widgets import ExtendedEdit as Edit, NPile, NColumns, NListBox, Choice, AlarmsEditor, \
+    FocusLineBox
 from .base import Pane, Window
 from .startendeditor import StartEndEditor
 from .calendarwidget import CalendarWidget
@@ -357,7 +358,7 @@ class EventColumn(urwid.WidgetWrap):
         current_day = self.container.contents[0][0]
 
         if self.pane.conf['view']['frame']:
-            ContainerWidget = urwid.LineBox
+            ContainerWidget = FocusLineBox
         else:
             ContainerWidget = urwid.WidgetPlaceholder
         new_pane = urwid.Columns([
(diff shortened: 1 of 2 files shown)


Tonus1:
My vote for the thick border ! 

By the way, nice job :-)

geier:
Using the thick border has the advantage that we could make (at least the right column) look like this
```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┃                                              ┃   
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│                                              │   
│                                              │   
│                                              │   
│                                              │   
│                                              │   
│                                              │   
│                                              │   
└──────────────────────────────────────────────┘ 
```

untitaker:
Definetly prefer the thick border, though I wonder if changing the color instead is a good idea (to stay consistent with tiling WMs and tmux)

geier:
Good idea, I'll give that a try.

Quoting Markus Unterwaditzer (2016-05-05 18:20:56)
> Definetly prefer the thick border, though I wonder if changing the color instead is a good idea (to stay consistent with tiling WMs and tmux)

geier:
Question is, which colors. Bold and standout does not make a difference (at least with the fonts I use). With solarized and xterm's standard color scheme I cannot distinguish `white` and `light gray`. As a highlighting color `yellow` and `light blue` look ok in solarized, but `yellow` looks ridiculous in xterm and `light blue` not too good either.

I'm going to update the other branch `FocusLine_thick` and if you are interested in it, you can give it try, too (need to edit the line `frame` and `frame focus` in colors.py).

untitaker:
I just tried this and I'm unable to see any border. I've never seen one for khal. What am I doing wrong?

geier:
in your config file:
[view]
frame = True

Quoting Markus Unterwaditzer (2016-05-05 21:08:28)

geier:
I've planed making `frame = True` for some time, but never come around to do it.

untitaker:
What do you think of: https://asciinema.org/a/1bv5sl3gmva6p32tdwkssuput

geier:
I really like the visual effect, the dark magenta (?) looks really nice! But I'm kind of missing the visual separation between the calendar and the event-column, the event-editor and the event-list, and the border of the terminal... I'm undecided and will leave this open for a bit.

More ideas and suggestions are very welcome.

untitaker:
It's bright red. https://github.com/untitaker/khal/tree/FocusLineWidget_untitaker

untitaker:
I personally think `frame = True` is a bit wasteful in managing space. I suspect implementing the separator like in tmux is too complicated though.

geier:
Bright red looks a bit dirty with my color scheme... 

As long as it's only two columns, the tmux way is really easy, it gets a bit more complicated when we want to use the `T` bit, but I want to do that anyway.

PR #414, opened 2016-05-09
Support ICS 'categories' field

pdav:
- Add a `categories` field to the `event` class
- Add the `categories` field to editor
- Add the `-g` flag to `ikhal new`


pdav:
Travis fails for the test I've modified. I bet this is just a silly question of order of fields in the VEVENT, but I cannot see the "-v" output of travis. Would someone be kind enough to send me the output of travis with the "-v" option set?

untitaker:
Implementation looks straightfoward, :+1:

khal/khalendar/event.py
```diff
@@ -451,6 +458,8 @@ def relative_to(self, day, full=False):
                 body += ', ' + self.description.strip()
             if self.location.strip() != '':
                 body += ', ' + self.location.strip()
+            if self.categories.strip() != '':
+                body += ', ' + self.categories.strip()
```
untitaker: I think it would make sense to group the calendar selector and the categories textfield together.

khal/ui/__init__.py
```diff
@@ -451,6 +451,9 @@ def __init__(self, conf, event, collection=None):
         if event.location != '':
             lines.append(urwid.Text('Location: ' + event.location))
 
+        if event.categories != '':
+            lines.append(urwid.Text('Categories: ' + event.categories))
```
untitaker: Same here.

khal/khalendar/event.py
pdav: It's a matter of personal taste.

The right solution would be to have a configuration item for the order of fields. For example
> order_print_normal = Time, Title
> order_print_full = Title, Title, Description, Categories
> order_interactive = Title, Organizer, Location!, Date, Calendar, Categories
> order_editor = Title, Calendar, Location, Categories, Description, Date, Repeat, Alarm

With a !-mark if an item must be displayed even if it is empty. Would be useful to also mask "Alarm" in the editor, which I never use.

geier:
Thanks for the PR!

First let me say that I'm going to merge this PR, because categories are
probably useful for some people.

But...
adding any element between the title and start/end times of an event
means that I need to press `tab` one more time than before, therefore I
believe ordering or even en-/disabling of editor interface elements (as
suggested by Pierre) is a very good idea and I'll see if I can make some
time for that (but only after I finished the timezone support in the
editor and the new daywalker)!

geier:
[In the RFC](https://tools.ietf.org/html/rfc5545#section-3.8.1.2) the example for multiple categories (``       CATEGORIES:APPOINTMENT,EDUCATION``) is comma separated, while the current implementation here escapes the comma with a backslash (example: ``CATEGORIES:CatA\,CatB``), is this an oversight or deliberately done this way (and if so, why?)?

pdav:
This could be related to: https://github.com/collective/icalendar/issues/127

geier:
Asking for your opinions: is this feature usable with that bug in icalendar?

untitaker:
Only with one category entered I suspect.

pdav:
I intensively use categories with SOGo server, which supports only one category by event (UI design limit). So, I agree with @untitaker, this feature is usable as is, provided one does not try to use more than one category.

geier:
Thank you, @pdav!

gpl34:
ikhal, version `0.8.2.dev118+ng4c1562e`

I use fastmail. If I modify an event in ikhal, I get:
```
debug: Sending request...
debug: 403
debug: {'Date': 'Wed, 29 Jun 2016 18:34:12 GMT', 'Content-Encoding': 'gzip', 'Transfer-Encoding': 'chunked', 'Cache-Control': 'no-cache', 'Vary': 'Accept-Encoding', 'Accept-Patch': 'application/calendar-patch; charset=utf-8', 'Server': 'nginx', 'Content-Type': 'application/xml; charset=utf-8', 'Connection': 'keep-alive'}
debug: b'<?xml version="1.0" encoding="utf-8"?>\n<D:error xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav">\n  <C:valid-calendar-data/>\n  <D:responsedescription>No value for CATEGORIES property. Removing entire property:</D:responsedescription>\n</D:error>\n'
```
ikhal is adding a `CATEGORIES` even if empty (I checked the .ics file) and fastmail does not like that. If I add a category, there's no sync error.

hobarrera:
I've been seeing this for a while too (though I'd been thinking it one was of my many broken, semi-experimental events).

All khal-created events are not uploading to fastmail (which is quite strict in it's validations).

PR #419, opened 2016-05-24
Use glob instead of listdir in conf path expansion

youngmoney:
The reason for this is that I have a directory with many accounts worth of calendars that I sync with vdirsyncer, I want them all in khal, so I want to be able to do `path = ~/.calendar/*` to have all my account's calendars load.


hobarrera:
I've the same directory setup as you, and this works for me:

```
[calendars]

  [[calendars]]
  path = ~/.local/share/calendars/
  type = discover
```

geier:
Why is glob better than listdir here?

Quoting youngmoney (2016-05-24 07:07:23)

untitaker:
He has a setup like the following: `~/.calendar/<account_name>/<calendar_name>/`.

By specifying `~/.calendar/*` in the config, the resulting glob is `~/.calendar/*/*/`, which is what he needs.

hobarrera:
It seems to me that my suggested config should work fine here too.
 
 
On Tue, May 24, 2016, at 06:43, Markus Unterwaditzer wrote:
> He has a setup like the following:
> ~/.calendar/<account_name>/<calendar_name>/.
> By specifying ~/.calendar/* in the config, the resulting glob is
> ~/.calendar/*/*/, which is what he needs.
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub[1]
>
 
--
Hugo Osvaldo Barrera
 

Links:

  1. https://github.com/pimutils/khal/pull/419#issuecomment-221218703

untitaker:
Ah, yes, I forgot that khal traverses recursively. I think globs are a nicer interface though.

geier:
Let's see what @youngmoney's thoughts are. In any case, before merging this PR the docs need to be updated as well (``[calendar]`` parameter `type`).

youngmoney:
I actually didn't realize that it was recursive, I guess I misread it/tested it wrong. I'm fine with it either way, I think glob is more intuitive for people but since it works either way and this is really a style choice I will leave it up to you. The only thing that is nice is that files that are hidden (i.e. starting with .) would be ignored in the glob case, which would solve #375 (and #401)

As a point of clarity as the code is now, glob has simply replaced listdir and it is still recursive. This means both `~/calendars` and `~/calendars/*` would produce the same result, and both would ignore dot files

youngmoney:
This also allows for more complex setups such as `~/account/*/calendars`

geier:
Sounds good, this is probably a good change. Could you please elaborate a bit more in the docs, e.g. just give the example you gave above?

youngmoney:
So I updated the config, but I have no idea why that made a test start failing. In fact that test does not fail on my local machine, and did not fail until I changed the config some more...any help on understanding this would be great

youngmoney:
accidentally closed this one while trying to figure out why the test was failing, will open a new PR referencing this

Issue #422, opened 2016-05-26
Test failures for 0.8.2


lfam:
I'm trying to build khal 0.8.2 for Guix, but I'm getting some test failures, as shown in the attached build log.

On #pimutils, it was suggested to increase the sleep times in khalendar_test.py. Sleeping for 1 second,   TestCalendar.test_db_needs_update did pass, but the other failures remained.

Versions of some dependencies and other potentially useful information:
linux-4.5.0
ext4 filesystem
x86_64 architecture
python-icalendar-3.10
python-dateutil-2.5.3
pytest-2.7.3
freezegun-0.3.7

Please let me know how I can help debug this.

Also, if these test failures indicate a problem with the tests themselves, rather than with khal, please let me know so I can disable them and update our package.

[khal-log.txt](https://github.com/pimutils/khal/files/285124/khal-log.txt)

geier:
I believe the failure of `aux/test_repeat()` is a bit different than the others, but I cannot reproduce it locally. If anyone can, can you paste the output of `pytest -v`? (or perhaps with some more added `v`s).

lfam:
Here is the output of `py.test --verbose tests`, from within the Guix build environment. What's weird is that the only failing test is 'tests/aux_test.py', even though I haven't changed anything besides invoking pytest with `--verbose`.

```
starting phase `check'
============================= test session starts ==============================
platform linux -- Python 3.4.3 -- py-1.4.31 -- pytest-2.7.3 -- /gnu/store/s7dh9x5f6vl4ww3vyjh831mh828jfr17-python-wrapper-3.4.3/bin/python
rootdir: /tmp/guix-build-khal-0.8.2.drv-0/khal-0.8.2, inifile: 
collecting ... collected 177 items

tests/aux_test.py::test_normalize_component PASSED
tests/aux_test.py::TestGuessDatetimefstr::test_today PASSED
tests/aux_test.py::TestGuessDatetimefstr::test_tomorrow PASSED
tests/aux_test.py::TestGuessDatetimefstr::test_time_tomorrow PASSED
tests/aux_test.py::TestGuessTimedeltafstr::test_single PASSED
tests/aux_test.py::TestGuessTimedeltafstr::test_negative PASSED
tests/aux_test.py::TestGuessTimedeltafstr::test_multi PASSED
tests/aux_test.py::TestGuessTimedeltafstr::test_multi_nospace PASSED
tests/aux_test.py::TestGuessTimedeltafstr::test_garbage PASSED
[...]

PR #424, opened 2016-05-28
IGNORE - this is to track down a test failure

youngmoney:
this is based on #419, I am trying to figure out why the test fail even when no code was changed. After solving this I will repush #419 as a new PR (since I cannot reopen)  Yes, I know this is a mess, sorry, but github leaves me no choices


untitaker:
There is a flaky test, see #422

untitaker:
Basically don't worry about it right now, and just submit the PR

PR #426, opened 2016-05-28
list command/event.format allday event endings

### LINKED ISSUE #418: khal list shows AllDayEvents as starting and ending at the same time
```
khal list
22.05. 00:00-22.05. 00:00 An awesome Event
```

youngmoney:
fixes #418

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -27,7 +27,7 @@ import os
 import icalendar
 import pytz
 
-from ..aux import generate_random_uid
+from ..aux import generate_random_uid, datetime_fillin
 from .aux import to_naive_utc, to_unix_time, invalid_timezone, delete_instance
 from ..log import logger
 from click import style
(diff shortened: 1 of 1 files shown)


geier:
This is definitively an improvement and I'm going to merge it therefore.

I'm not really happy with it for conceptional reasons:
 * all-day events just do not end at 23:59, the last minute of that day is still "covered" by the event (and printing 00:00 of the following day is confusing/wrong if you only  print days)
* all-day events really should not have start and end times at all

Some ideas on how to fix this off the top of my head:
- just use an empty string as the start and end time for all-day events
- use different --format options for all-day and datetime events (perhaps even floating and localized events), perhaps with a default --format options that works for all event types in combination with the previous point and than two (or three) additional ones
- use a better templating language, but it should be pretty lightweight (in terms of `import templating` speed, e.g., jinja2 takes much too long)

PR #428, opened 2016-05-29
Remove vdirsyncer dependency

untitaker:


untitaker:
@geier This finally passes! Opinions?

geier:
I really think its a good idea to remove the hard dependency on vdirsyncer from khal.

I'm not sure though, if we shouldn't just copy the vdir code over to khal:

 * adding another dependency that only we use puts an unnecessary strain on packagers.
 *  do you consider vdir stable? Because if not, the situation is not any better then before.
 * if you think we really should depend on an new package, I think you should make a first release of vdir before we merge this

untitaker:
I guess I agree that vendoring is a better option.

khal/khalendar/vdir.py
```diff
@@ -0,0 +1,336 @@
+'''
+Based off https://github.com/pimutils/python-vdir, which is itself based off
+vdirsyncer.
```
untitaker: I don't think there's any work of other copyright holders left in python-vdir. I wrote it as-is, and as I'm already listed in khal's AUTHORS, I think khal's copyright can be used.

geier: I guess that is totally up to you, if you want to be listed as the author, go ahead and fix that.

hobarrera:
+1 to vendoring. It makes sense to split it off once it's stable and mature enough.

When I have time on my hand once again, I'll try to see what I can adapt for `todoman` (you may have noticed I've been quite absent for some time).

khal/khalendar/vdir.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def set_meta(self, key, value):
+        value = value or u''
+        assert isinstance(value, str)
+        fpath = os.path.join(self.path, key)
+        with atomic_write(fpath, mode='wb', overwrite=True) as f:
+            f.write(value.encode(self.encoding))
+
+
+class cached_property(object):
```
hobarrera: Also, no need to extend object, this can just be `class cached_property:`.

untitaker:
Done, ready to merge

geier:
@homu r+

geier:
@untitaker thanks!

PR #429, opened 2016-05-31
New container widgets changing style if in focus

geier:
Which of those new widgets is used (if any) can be changed in the
configuration file. Please provide feedback which ones you like (if
any), so the not often used ones can be removed later.

Supersedes #412


geier:
included with 5da93fbe500a5297479a0feceec3c1c19c68b3ae

Issue #430, opened 2016-05-31
--format printing of all-day start and end times


geier:
with --format we currently print all-day events like this (with the default format):
`31.05. 00:00-31.05. 23:59 all day`

see #426 

I'm not really happy with it for conceptional reasons:

    all-day events just do not end at 23:59, the last minute of that day is still "covered" by the event (and printing 00:00 of the following day is confusing/wrong if you only print days)
        all-day events really should not have start and end times at all

Some ideas on how to fix this off the top of my head:

    just use an empty string as the start and end time for all-day events
    use different --format options for all-day and datetime events (perhaps even floating and localized events), perhaps with a default --format options that works for all event types in combination with the previous point and than two (or three) additional ones
    use a better templating language, but it should be pretty lightweight (in terms of import templating speed, e.g., jinja2 takes much too long)

youngmoney:
I'd say lets add an ending to the existing start/end that removes the time if its an all day event: so `--format "{title}{start-clean}{end-clean}"` would show the start/end normally for an event, but would just show the dates for an all day event. I could easily add the -clean ending to each of the current formats. This way users have the option as to weather they want the times shown (and if they are shown I say we leave them as 23:59)

geier:
fixed with #443

youngmoney:
I think you meant #433 (just so we have a record)

geier:
Thanks, I meant that indeed.

PR #433, opened 2016-06-01
fixes allday events in format

youngmoney:
see #430   In this case I decided to have the default be that times are hidden for all day events, and the addition of `-full` to the end of a template option forced the showing of times.   Right now it considers all day events those that are starting and ending on the boundry of the day being listed, I could change this easily, I just need to know what we consider all day events.

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -204,6 +204,11 @@ available templates are:
 
         A newline if start-date-once is not "".
 
+By default all day events will not have times. To see the times anyway simply
+add `-full` to the end of any template with start/end, for instance
+`start-time` becomes `start-time-full` and will always show the time (instead
+of being empty for all day events).
+
 In addtion there are colors: `black`, `red`, `green`, `yellow`, `blue`,
 `magenta`, `cyan`, `white` (and their bold versions: `red-bold`, etc.). There
 is also `reset`, which clears the styling.
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -491,17 +491,33 @@ class Event(object):
         self_start = datetime_fillin(self.start_local, locale=self._locale, end=False)
         self_end = datetime_fillin(self.end_local, locale=self._locale)
 
+        allday = isinstance(self, AllDayEvent)
 
         attributes["start"] = self_start.strftime(self._locale['datetimeformat'])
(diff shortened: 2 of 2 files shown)


geier:
All-day events are `khalendar.event.AllDayEvent`s and I don't think we should put the flag "into" the format string, rather put them behind.

geier:
@youngmoney could you look into the failing tests? They seem to be related to the new discover code.

youngmoney:
the only reason I suggest allowing the -full ending is because the default event_format could then contain it, and I think semantically it makes sense, its no different than an option for a timestring or any other templating/string thing. I can add an option if you'd like, though if we eventually bring the --format option out to all commands (which I plan to do) then this could lead to a lot of extra flags. Anyway, up to you

youngmoney:
Ill look into the tests as well

youngmoney:
I've updated to use isinstance

I have not added a command line option yet, I can if you want me to. I do believe that the template option is best because it allows for forcing different things at the same time eg seeing start times no matter what but only seeing end times of non-allay events. Maybe not the strongest use case but like I said its at least an option

geier:
The current solution has one big advantage: we can easily configure the default format from the configuration file with just one setting, I agree that's a big plus. On the other hand, if we'll add more options, we will a big mess on our hand as we'll have to do proper configuration parsing on that string. I'm undecided on what's the best solution. I guess we'll leave it as is for now.

youngmoney:
I'm not sure I see entirely why the current code would cause problems if we add more options/why there would be parsing issues. I'll just try to explain fully what I'm thinking, just to get it out there (so I may be saying things I already said, but just want it all in one place)

Right now (is this PR) `--format "{start} {title}"` prints the date and time of normal events and the date only for allday events. 

`--format "{start-full} {title}"` always shows the date and time regardless of the type of event

To me this makes more sense as template option just like in date time there is a different symbol for the hours in 12, 24, 12 zero padded, 24 zero padded. The -full ending is like zero padding to me. I'm not sure there are that many other things we will want to add that are like this, so it does not worry me as much.

Also having a rich vocabulary to print events does not feel like a negative. Maybe for code complexity but to me it is better to have the format string be able to print everything about an event than have to support tons of configuration options.

Anyways like I said just my full thoughts. An option for full is certainly not that big of a deal either.

geier:
Ok, let's see how it works out. Thanks for the fix!

PR #434, opened 2016-06-01
Fix of discover tests

youngmoney:
this fixes the discover tests that were broken by globbing essentially, I just removed the recursions that are no longer support and added `/*/*`


geier:
Awesome! Thanks!

Issue #273, opened 2015-10-02
How should at and agenda work?


geier:
Please post your ideas on how to improve on the current situation of ``at`` and ``agenda``. Do we need two commands? What arguments should they accept, what should they print, etc.

Also see #93 and #231.

matthiasbeyer:
* Command: `agenda`
    * `--help`
    * `--events n` - a working version please
    * `--days n`
    * `--calendars ...` - To select which calendars are shown
    * `--not-calendars ...` - To exclude calendars
    * `--json` - To print results as JSON
    * `--grep key=value,key=value` - List for grepping in a field (for example "SUMMARY" from the ical files) for a certain value. Additionally not a list of `key=value` pairs but something with `or` and `and` logical operators
    * `--before-time` - All events which start/end before a certain daytime
    * `--after-time` - All events which start/end after a certain daytime

I like the current output of the `agenda` command, though it could use some styling, some blank lines after the events of one day, some more table-ish style, where you can see things more instantly.

---

I never used the `at` command.

untitaker:
I also never used the at command.

I'd like to see as many UI config params as possible deprecated, and only allow configuration through CLI options.

On 3 October 2015 21:31:28 CEST, Matthias Beyer <notifications@github.com> wrote:
>* Command: `agenda`
>    * `--help`
>    * `--events n` - a working version please
>    * `--days n`
>    * `--calendars ...` - To select which calendars are shown
>    * `--not-calendars ...` - To exclude calendars
>    * `--json` - To print results as JSON
>* `--grep key=value,key=value` - List for grepping in a field (for
>example "SUMMARY" from the ical files) for a certain value.
>Additionally not a list of `key=value` pairs but something with `or`
>and `and` logical operators
>* `--before-time` - All events which start/end before a certain daytime
>  * `--after-time` - All events which start/end after a certain daytime
>
>I like the current output of the `agenda` command, though it could use
>some styling, some blank lines after the events of one day, some more
>table-ish style, where you can see things more instantly.
>
>---
>
>I never used the `at` command.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/geier/khal/issues/273#issuecomment-145280985

-- 
Sent from my phone. Please excuse my brevity.

geier:
see #325 (Option to display only the first occurrence of reoccurring events)
see #272 (Hide past events)

youngmoney:
I think it might be time to close this as most of the work on these commands has been done and they have changed quite a bit

PR #435, opened 2016-06-03
Use format and list for at command

youngmoney:
I have replaced the at code with a simple call to list, and given it format
options. This includes a `--notstarted` flag, which I think will be nice.

To get the same feel as the previous at command, the format string should be
set to:

`"{start}-{end}: {title}\nLocation: {location}\nCategories:
{categories}\nRepeat: {repeat-pattern}\nDescription: {description}"`

this will not have the lines removed with no values like at currently does, but
it is close and I think it is worth sacrificing for consistency.

### DIFF
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -106,7 +106,7 @@ def calc_day(dayname):
     :returns: date
     :rtype: datetime.datetime
     """
-    today = datetime.today()
+    today = datetime.combine(date.today(), time.min)
     dayname = dayname.lower()
     if dayname == 'today':
         return today
@@ -169,6 +169,12 @@ def guessdatetimefstr(dtime_list, locale, default_day=None):
             a_date = datetime(*(default_day.timetuple()[:3] + a_date.timetuple()[3:5]))
         return a_date
 
+    def datetimefwords(dtime_list, _):
+        if len(dtime_list) > 0 and dtime_list[0].lower() == 'now':
+            dtime_list.pop(0)
+            return datetime.now()
+        raise ValueError
+
     def datefstr_year(dtime_list, dateformat):
         """should be used if a date(time) without year is given
 
@@ -201,6 +207,7 @@ def guessdatetimefstr(dtime_list, locale, default_day=None):
             (datefstr_year, locale['dateformat'], True),
             (datetimefstr, locale['longdateformat'], True),
             (datefstr_weekday, None, True),
+            (datetimefwords, None, False),
 
     ]:
         try:
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -255,7 +255,7 @@ def guesstimedeltafstr(delta_string):
     return res
 
 
-def guessrangefstr(daterange, locale, default_timedelta=None):
+def guessrangefstr(daterange, locale, default_timedelta=None, first_weekday=0):
     """parses a range string
 
     :param daterange: date1 [date2 | timedelta]
@@ -284,12 +284,20 @@ def guessrangefstr(daterange, locale, default_timedelta=None):
             if start is None:
                 start = datetime_fillin(end=False)
             elif not isinstance(start, date):
-                split = start.split(" ")
-                start = guessdatetimefstr(split, locale)[0]
-                if len(split) != 0:
-                    continue
+                if start.lower() == 'week':
+                    today_weekday = datetime.today().weekday()
+                    start = datetime.today() - timedelta(days=(today_weekday - first_weekday))
+                    end = start + timedelta(days=7)
+                else:
+                    split = start.split(" ")
+                    start = guessdatetimefstr(split, locale)[0]
+                    if len(split) != 0:
+                        continue
 
-            if end is None or len(end) == 0:
+            if isinstance(end, datetime):
+                pass
+
+            elif end is None or len(end) == 0:
                 if default_timedelta is not None:
                     end = start + default_timedelta
                 else:
@@ -297,6 +305,9 @@ def guessrangefstr(daterange, locale, default_timedelta=None):
             else:
                 if end.lower() == 'eod':
                     end = datetime_fillin(day=start)
+                elif end.lower() == 'week':
+                    start -= timedelta(days=(start.weekday() - first_weekday))
+                    end = start + timedelta(days=7)
                 else:
                     try:
                         delta = guesstimedeltafstr(end)
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -270,7 +270,7 @@ def guessrangefstr(daterange, locale, default_timedelta=None):
         range_list = daterange.split()
 
     try:
-        if len(default_timedelta) == 0:
+        if default_timedelta is None or len(default_timedelta) == 0:
             default_timedelta = None
         else:
             default_timedelta = guesstimedeltafstr(default_timedelta)
@@ -308,7 +308,7 @@ def guessrangefstr(daterange, locale, default_timedelta=None):
                             continue
                     end = datetime_fillin(end)
             return start, end
-        except Exception:   # XXX FIXME what kind of exceptions do we want to catch here?
+        except ValueError:
             pass
 
     return None, None
@@ -347,7 +347,7 @@ def datetime_fillin(dt=None, end=True, locale=None, day=None):
     if locale is not None:
         try:
             dt = locale['local_timezone'].localize(dt)
-        except:
+        except ValueError:
             pass
 
     return dt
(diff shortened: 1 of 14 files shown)


geier:
Why not use `collection.get_events_at()`? It's more precise than list +1m. I'm not (yet :wink:) saying we need to use it, just wanting your opinion.

It would enable us to delete a lot of code...

youngmoney:
Due to my slight ignorance of the full suite of functions, I did not think of using that. I was primarily concerned with getting formatting into `at`. I have no problem with using it. I'd only suggests list because then we have one "universal" implementation of listing events in a range especially when the future of agenda/at/list is still under discussion, but that's not a big deal since it might be true that `at` needs to be more precise.

Either is fine, let me know and I'll set it up right

youngmoney:
It turns out that the list command supports `0m` as an end, so I changed it to
that, not it is just as precise as `events_at`. I would now say that using list
for all listing is a good idea, especially considering its precision. Plus I
will probably continue to send pull requests turning things into list with format.

khal/aux.py
```diff
@@ -112,6 +112,8 @@ def calc_day(dayname):
         return today
     if dayname == 'tomorrow':
         return today + timedelta(days=1)
+    if dayname == 'now':
```
geier: this function is called `calc_day` and currently only returns `datetime.date` objects, I don't think we should start returning `datetime.datetime` objects from here.

geier:
You are right, we probably can replace `at` internally with `list`. There is one difference from the old behavior we should tell the users about: `at` used to be `event_start <= query_time <= event_end` using list it will be `event_start <= query_time < event_end` (note the missing second `=`).

Also, I will not merge this yet, because I there are some things in the `list/format` code we should take care of first before we "break" the working `at` command:

* I want to get the old style event printing back as default first (at the very least calendar colors)
* we should increase test coverage (e.g. guessrangestr)
* specifying timedelta in config file is broken or does not work as I expect it would (and should probably be validated)

geier:
We should probably go one step further, and replace all printing of events with `format` (at least in khal for now), that leaves `search`, `agenda` (which we might want to remove altogether, or rename list), `calendar`, `new` (which can also print the just created event) and probably `import`.

geier:
I've got the color issue covered, will commit tomorrow.

youngmoney:
I'll start adding commits to this PR fixing those issues, btw the timedelta
should only be working for the list command as the at command has a timedelta
set in stone, I will look into it.

converted to list: at, search, (will remove agenda)
to be converted: calendar, new, import

youngmoney:
I've pushed all the changes except color (which you are working on) and removing agenda, which I will push later

khal/aux.py
youngmoney: Calc day actually returns date time objects, and used to call datetime.today() which is the same as datetime.now(), whereas date.today() is a date of today. I updated the code to have it return the start of today, and left now as an option there since the return type is the same, let me know your thoughts/if I should move now

youngmoney:
ignore what I said about calc_day, I've changed to commit to add a separate function, I have also fixed calc_day to return the start of the day instead of the current time in that day

geier:
Color should be fixed in master, you might want to rebase your branch. 

I'm sorry that I don't have time to review atm, hopefully Thursday night (cest), otherwise Friday. 

Also, I know that I'm more critical with your code than with mine. It's not intentionally, I just don't notice those things when looking at my own code (feel very free to leave notes or questions, you are probably touching a lot of code nobody but me ever touched, especially on my pull requests). 

Am 8. Juni 2016 00:29:02 MESZ, schrieb youngmoney <notifications@github.com>:
>ignore what I said about calc_day, I've changed to commit to add a
>separate function, I have also fixed calc_day to return the start of
>the day instead of the current time in that day
>
>---
>https://github.com/pimutils/khal/pull/435#issuecomment-224434003

-- 

youngmoney:
I have pushed changes that remove agenda completely and switched over to format/list for everything that is not ikhal

not sure why but CI isn't always triggered, anyway should be passing all tests, however there might be a style problem with the docs but I don't really understand it/how to test it on my machine

if we merge all of this, it will close: #273, #430 (if we want to close it)

geier:
Some notes:

I think we need different (default) formats for different commands, the
current default command doesn't work well for search (and probably isn't
great for new either)

I really don't like putting seen-days in that non-local dict, I think
it's much cleaner to add the day in get_list_from_str to event_column
If we introduce another format string for the dayname, that can be
customized as well.
The current setup also leads to `search` and `calendar` being broken,
example:

```
khal search Beer
Today: →1:00           Beer
20:00-22:00           More Beer
```

I have no suggestion for naming those two different default formats
though.

The documentation does not describe all currently allowed format
strings (perhaps the documentation can be build from event.py somehow,
so that documentation and code are at the same place, not a priority
though).

youngmoney:
I agree that having two different options is a good idea, I'll rename the current default to agenda_event_format and change the default event_format to be more sane.

For the dictionary are you saying you don't like a non-local dictionary in general or just for the seen-days, because I do actually add the days to the format string and simply keep track of the first occurrence in the dict. Are you proposing another format string for header lines?

I'll update the docs

geier:
Yeah, I don't like the nonlocal dict, it makes the behavior rather unpredictable. 

Am 10. Juni 2016 17:50:09 MESZ, schrieb youngmoney <notifications@github.com>:
>I agree that having two different options is a good idea, I'll rename
>the current default to agenda_event_format and change the default
>event_format to be more sane.
>
>For the dictionary are you saying you don't like a non-local dictionary
>in general or just for the seen-days, because I do actually add the
>days to the format string and simply keep track of the first occurrence
>in the dict. Are you proposing another format string for header lines?
>
>I'll update the docs
>
>> On Jun 10, 2016, at 8:32 AM, Christian Geier
><notifications@github.com> wrote:
>> 
>> Some notes:
>> 
>> I think we need different (default) formats for different commands,
>the
>> current default command doesn't work well for search (and probably
>isn't
>> great for new either)
>> 
>> I really don't like putting seen-days in that non-local dict, I think
>> it's much cleaner to add the day in get_list_from_str to event_column
>> If we introduce another format string for the dayname, that can be
>> customized as well.
>> The current setup also leads to search and calendar being broken,
>> example:
>> 
>> khal search Beer
>> Today: →1:00           Beer
>> 20:00-22:00           More Beer
>> I have no suggestion for naming those two different default formats
>> though.
>> 
>> The documentation does not describe all currently allowed format
>> strings (perhaps the documentation can be build from event.py
>somehow,
>> so that documentation and code are at the same place, not a priority
>> though).
>> 
>> —
>> You are receiving this because you authored the thread.
>> Reply to this email directly, view it on GitHub, or mute the thread.
>> 
>
>
>---
>https://github.com/pimutils/khal/pull/435#issuecomment-225220547

-- 

youngmoney:
I added the option for two different format strings, one for list and calendar
and one for the rest (agenda_event_format, event_format) in addition there is
now agenda_day_format which describes the headers.

I've removed the non-local dictionary in event, however I did have to add a
nonlocal set to get_list so that --once would work. I think this nonlocal set
is worth having because it allows us to have get_list be its own function. (its
also a completely optional parameter)

geier:
I like this much better, thanks. Some more notes:

* calendar is not colored, is this intentional?
* events ending at 00:00 are displayed with an arrow as if not ending on one that day, but not on the next

youngmoney:
I've fixed both, I forget to pass the colors in

geier:
Thanks. Can you rebase or should I?

youngmoney:
I'll do it now

Thanks,
Taylor Money

youngmoney:
I don't understand some of the style tests and why they are failing, (why does travis not explain whats wrong with the lines). In particular I do not understand the .rst files and their problems

geier:
I'll have a look. 

Am 18. Juni 2016 02:48:17 MESZ, schrieb youngmoney <notifications@github.com>:
>I don't understand some of the style tests and why they are failing,
>(why does travis not explain whats wrong with the lines). In particular
>I do not understand the .rst files and their problems
>
>---
>https://github.com/pimutils/khal/pull/435#issuecomment-226911693

-- 

geier:
The issues is the typo `seperator` (should be `separator`), we grep for the wrong spelling with tox (see `tox.ini`), because I make that error a lot.

youngmoney:
alright its pushed and passing, ready for merge. Let me know if you have any other comments

geier:
two more (hopefully) the last requests:

* [ ] please comment on removed and added new configuration file options in the CHANGELOG
* [ ] did you remove `show_all_days` on purpose? and is there an equivalent for it?

youngmoney:
I added show all days back, I removed it but realize people might actually want it. I've also commented on the config

geier:
Thank you for a lot of hard work! I've found one more error which I've fixed and will push after I've merged.

PR #437, opened 2016-06-05
ESC key backtracks, but does not exit from ikhal

pdav:
Usability improvement: ESC key backtracks, but never exit ikhal (only 'quit' keybinding do).

This behaviour is similar to `vi`: ESC key brings you back to a known state, whatever the number of times you press this key.

The existing behaviour is very frustrating since in everyday life, I often exit when I do not want to.


geier:
Probably a good idea. Opinions? ( @untitaker, @hobarrera )

untitaker:
Yes, IMO a good idea.

geier:
Thanks @pdav! Merged.

Issue #49, opened 2014-04-11
delete events from khal


geier:
suggested usage:

`khal --delete 04.15 09:00 11:30 some event`

`khal --edit 04.15 09:00 11:30 some event - new time 09:00 12:00`

(not sure how practical that really is)

lucc:
Or the shell completion could get the uids and complete on them. Zsh can present the user with augmented completion entries so that the human readable version of the event could be displayed next to the uid. The git zsh completion presents recent commit ids in this way.

geier:
@lucc good idea, example with zsh: `kill` <kbd>tab</kbd>

geier:
as editing has been merged, I'll update the title.

geier:
this can now be done with the new `edit` command.

PR #438, opened 2016-06-05
add command edit

### LINKED ISSUE #407: Fixes to keep up with gcalcli
Hi, I am in the process of switching from gcalcli to khal (since vdirsyncer now supports google). I have noticed that in terms of commands gcalcli seems to be ahead in a few ways. I would like to help bridge that gap.

A few ideas (in order of priority for me):

allow agenda to have different formats (#273) including json (and for me TSV), or at the very least a pretty default, the setup I have looks like: 
screen shot 2016-04-25 at 9 26 56 am
this setup was created using awk piped from gcalcli tsv output, and recreated using some regex from khal agenda output

an upcoming command, in gcalcli this calls an external command with event that will start in the next N minutes, all I care about is having either at or agenda support a next N minutes start and then a reasonable output format so that I can parse for notifications

interactive new and edit/delete, these commands would either drop right into the new/edit or interactive or have their own flow, the point here would be to initiate from the command line. In gcalcli one can type `gcalcli edit SEARCH` which will loop through all matches and ask what you want to do (same with delete). We could easily have a loop over matching searches and ask if edit or delete is waste and or use a similar edit flow and the interactive new. The point is that some people (myself) do not want to open ikhal, then search, then edit, when a simple `khal edit SEARCH` would find the event and allow me to change one detail. (shot below)
screen shot 2016-04-25 at 9 35 50 am

As I am new to the project I am really just trying to gauge whether these changes are within what khal is trying to be. Please let me know any thoughts, I will hopefully start on this in the next few weeks.

youngmoney:
I've created a new interactive command called `edit`, it works completely in
the command line using just prompts, this will close #407, and I believe it solves #49.

This command accepts a search string, and then loops through each event and
asks if the user wants to edit it. If so the user chooses an aspect (eg title),
enters a new value and it is saved. events can also be deleted.

In the future future it would be good to add command line options and a batch
mode, this would enable other commands to edit events. Of course we would need
a way to be specific.

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -407,6 +407,18 @@ additional description
 adds a new all day event on 26th of July to the calendar *work* which recurs
 every week.
 
+
+edit
+****
+an interactive command for editing and deleting events using a search string
+
+::
+
+    khal edit [--hide-past] event_search_string
+
+the command will loop through all events that match the search string,
+prompting the user to delete, or change attributes.
+
 printcalendars
 **************
 prints a list of all configured calendars.
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -318,8 +318,8 @@ allows for adding new events. ``khal new`` should understand the following synta
 
 ::
 
-    khal new [-a CALENDAR] [OPTIONS] [START] [END | DELTA] [TIMEZONE] [SUMMARY]
-    [:: DESCRIPTION]
+    khal new [-a CALENDAR] [OPTIONS] [START [END | DELTA] [TIMEZONE] SUMMARY
+    [:: DESCRIPTION]]
 
 where start- and enddatetime are either datetimes, times, or keywords and times
 in the formats defined in the config file. If no calendar is given via
@@ -419,7 +419,7 @@ an interactive command for editing and deleting events using a search string
 
 ::
 
-    khal edit [--hide-past] event_search_string
+    khal edit [--show-past] event_search_string
 
 the command will loop through all events that match the search string,
 prompting the user to delete, or change attributes.
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -318,7 +318,8 @@ allows for adding new events. ``khal new`` should understand the following synta
 
 ::
 
-    khal new [-a CALENDAR] [OPTIONS] startdatetime [enddatetime] [timezone] summary [description]
+    khal new [-a CALENDAR] [OPTIONS] [START] [END | DELTA] [TIMEZONE] [SUMMARY]
+    [:: DESCRIPTION]
 
 where start- and enddatetime are either datetimes, times, or keywords and times
 in the formats defined in the config file. If no calendar is given via
@@ -360,6 +361,10 @@ accepted as the end of a given date.
 If the **summary** contains the string `::`, everything after `::` is taken as
 the **description** of the new event, i.e., the "body" of the event (and `::`
 will be removed).
+
+Passing the option `--interactive` (`-i`) makes all arguments optional and
+interactively prompts for required fields, then the event may be edited, the
+same way as in the `edit` command.
 
 Options
 """""""
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -392,6 +392,34 @@ def generate_random_uid():
     return ''.join([random.choice(choice) for _ in range(36)])
 
 
+def rrulefstr(repeat, until, locale):
+    if repeat in ["daily", "weekly", "monthly", "yearly"]:
+        rrule_settings = {'freq': repeat}
+        if until:
+            until_date = None
+            for fun, dformat in [(datetimefstr, locale['datetimeformat']),
+                                 (datetimefstr, locale['longdatetimeformat']),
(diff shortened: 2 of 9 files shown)


youngmoney:
As a side note I would like to get this merged (if it will be) before I start
working on the interactive part of the `new` command so that I can reuse parts
of it.

youngmoney:
actually, forget that (and sorry for so many comments) Ill actually push the
interactive additions to `new` to this PR as well hopefully this weekend. I
would like to see #406 closed within the next week or so.

geier:
Can we have some tests for `aux.timedelta2str()` please? Otherwise this looks good.

BTW, in case you didn't know, if you install pytest-cov, you can do `py.test --cov=khal --cov-report html tests` and get a nice graphic coverage in a folder named `htmlcov`.

youngmoney:
I just pushed the interactive new command option, there is a problem that
occurs when I try to update the new event, the assertion in khalendar.update
that event.etag exists is failing. I do not know enough to understand why.

geier:
I'll have a look. 

Am 15. Juni 2016 23:06:21 MESZ, schrieb youngmoney <notifications@github.com>:
>I just pushed the interactive new command option, there is a problem
>that
>occurs when I try to update the new event, the assertion in
>khalendar.update
>that event.etag exists is failing. I do not know enough to understand
>why.
>
>---
>https://github.com/pimutils/khal/pull/438#issuecomment-226320136

-- 

youngmoney:
Great! Thanks, it appears if you interactively add an event, and then it switched to interactive editing of the same event (at the end of the new function in controllers) it occurs when one interactively changes an attribute eg description

geier:
The issue with updating a "new" event should be fixed with  9e560c9, you best rebase on master.

Some notes on `new -i`:
* I don't think it's a good idea to cram the interactive part into `new_from_string`, because it's not actually from string, better put it in `cli.new`
* putting invalid values into any field gives a generic error message ('critical: Error`), I'd suggest validating those values on the fly and reporting back to the user without exiting
* running `khal new` without any arguments gives less than helpful error message
* cannot give no description
* when creating an event the summary is called title (or vice versa), I think it's better to stay consistent
* should we tell the user the default timezone?

youngmoney:
okay I have pushed rebased changes that address all of those concerns, I have added the options to accept "None" as an option to certain details in edit and interactive new, in addition I've factored out interactive new and made errors way better

youngmoney:
any thoughts? I would like to finish this up this week if possible

khal/aux.py
```diff
@@ -392,163 +444,85 @@ def generate_random_uid():
...
-                    except ValueError:
-                        pass
-                if until_date is None:
-                    logger.fatal("Cannot parse until date: '{}'\nPlease have a look "
-                                 "at the documentation.".format(until))
-                    raise FatalError()
-                rrule_settings['until'] = until_date
-
-            event.add('rrule', rrule_settings)
+def eventinfofstr(info_string, locale, default_timedelta=None,
```
geier: please write a docstring that describes what those parameters do and what you expect `info_string` to look like

khal/cli.py
```diff
@@ -344,26 +342,45 @@ def new(ctx, calendar, start, end, timezone, summary, description, location, cat
         everything behind them is taken as the event's description.
         '''
         # ugly hack to change how click presents the help string
-        eventlist = [start, end, timezone, summary] + list(description)
-        eventlist = [element for element in eventlist if element is not None]
+        if not info and not interactive:
+                raise click.BadParameter(
+                    'no details provided, '
+                    'did you mean to use --interactive/-i'
+                )
```
geier: missing `?` ?

geier:
Why do I need to type `None` instead of just entering an empty string?

geier:
when creating a new event with `-i`, it's not clear, that `Edit? no` actually means save, not sure if it's worth the hassle though.

doc/source/usage.rst
```diff
@@ -318,7 +318,8 @@ allows for adding new events. ``khal new`` should understand the following synta
 
 ::
 
-    khal new [-a CALENDAR] [OPTIONS] startdatetime [enddatetime] [timezone] summary [description]
+    khal new [-a CALENDAR] [OPTIONS] [START] [END | DELTA] [TIMEZONE] [SUMMARY]
+    [:: DESCRIPTION]
```
geier: I think `khal new [-a CALENDAR] [OPTIONS] [START [END | DELTA] .... summary ... ]` would be clearer, as the order is still important.

youngmoney:
I've made all of those changes. the reason for the None is that the prompt is handled by click and click does not accept empty prompt responses because there is an option for a default value (which is used a lot in edit and in new) and when there is a default value then hitting enter with no text is considered accepting the default value. As such the only way to get an empty string in most cases is to explicitly type  None, otherwise we could not have a default value, which would be annoying in most cases. None seems to me to be the best option, of course the wording could change to be for instance the literal empty quotes "" instead of None

geier:
Thanks! I'll guess someone else can take care of this.

One more thing though: this currently breaks creating new events from ikhal, please have a look.

geier:
Thanks and merged.

PR #440, opened 2016-06-13
[RFC] New layout of ikhal's event column

geier:
Known issues:
- [x] highlighting issues on the right when moving up in the calendar
- [x] colorscheme improvements (dark is a bit too aggressive, there is no light one yet)
- [x] find and delete dead code (if there is any, but I do suspect it)
- [x] Check if bindings for movements work
- [x] Update event list after adding/editing events
- [x] better support for (shift) tab
- [x] fix search
- [x] make work for long event lists (convert DatePile to some kind of ListWalker)
- [x] deal with event view (mainly removing it properly)


untitaker:
This is excellent work. I would still have a separate indicator for "which column is selected" instead of relying on the cursor color.

geier:
@untitaker Thanks. What kind of indicator are you thinking of? (I actually created the frame that change their appearance for that).

untitaker:
@geier Ah, yes, I thought those were mutually exclusive options.

Not sure if an additional cursor color is needed then though, assuming we make the border mandatory.

geier:
This branch in action: https://asciinema.org/a/5rcutk8the05cw2m02g3fcko2

Issue #449, opened 2016-06-23
multi-day all-day events don't show range_end and range_start


geier:

youngmoney:
fixed with #450

PR #450, opened 2016-06-23
fix allday range printing

youngmoney:
shows range_start on first day
shows range (double arrow) in middle
shows range_end on last day
if only one day shows nothing

I can change the one day to be both start and end, let me know


youngmoney:
fixes #449 (for bookkeeping)

geier:
Thanks!

Issue #455, opened 2016-06-30
formated start/end date ugly


geier:
I'm not too happy yet, with how start and end dates are printed (not that it was much better before), e.g. results from search:

```
01.04.-01.04. Event Summary
```
with `--format '{start}-{end-long}....` it gets a bit better:

```
01.04.-01.04.2016. Event Summary
```

but it would be much nicer if we could get this result:
```
01.04.2016 Event Summary
```
Another not optimal example:
```
19.02. 18:30-19.02.2016 20:30 Event Summary
```
much nicer:
```
19.02.2016 18:30-20:30 Event Summary
```

While all the above examples are with traditional German locale settings, I guess it's more or less the same with other locale/datetimeformat settings.

`str.format()` alone probably doesn't allow something like this, bringing in a proper templating language is too expensive (time wise).

Ideas on how to solve this are very welcome.

geier:
first idea: new template option `start-end-pretty` which tries to do the right thing.

youngmoney:
Something here that would work is {end-necessary} which either shows the end-long in the case of all day events that end on different days or in non-all day it shows either the end time or if it ends on a different day the end datetime.

We could actually have {end-necessary} and {end-long-necessary}

geier:
Nice. For allday events, {end-necessary} is a big improvement. but it also leads to dropped end dates and times for datetime events that end on the same day.

geier:
Also end-necessary is not yet documented. Perhaps we should generate the documentation from the code somehow.

youngmoney:
oh yes I have not yet created it! (sorry I see above it seemed like I did)

so I'll make it work like this:
allday: nothing if same day, else end date
normal: time if same day, else datetime

PR #456, opened 2016-06-30
add end-necessary to event.format

youngmoney:
simple fix for #455

comment on style/name/etc

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -130,6 +130,14 @@ Several options are common to almost all of :program:`khal`'s commands
         A concatenation of start-style, to-style, and end-style OR an
         appropriate symbol.
 
+   end-necessary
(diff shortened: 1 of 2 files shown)


geier:
good for now, only some docs please. And you might want to remove the idiotic defaultdict I put in there (or I will later, no rush).

youngmoney:
I've added documentation, as a side note I dont thing the defaultdict is
"idiotic", It is certainly fine to have it there as incorrect format strings
will not error, but instead just be empty strings

geier:
I'm unsure, the problem is, you might not notice typos, or, as happened to me in one of our last conversations, I thought `{end-necessary}` would already exist, but was just bugged.

youngmoney:
I'll fix this in day format #457

PR #457, opened 2016-07-04
day-format does not print empty string

youngmoney:
as the title says `--day-format ''` will not print any day headers

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -161,6 +161,9 @@ Several options are common to almost all of :program:`khal`'s commands
    name
         The date's name (`Monday`, `Tuesday`,…) or `today` or `tomorrow`.
 
+   If the `--day-format` is passed an empty string then it will not print the
+   day headers (for an empty line pass in a whitespace character).
+
 
 
 dates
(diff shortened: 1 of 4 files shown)


geier:
@youngmoney Please fix the test (or even better, also write a new one)

geier:
also: do you know why --day-format '\n' does print an actual `\n`? Some click.echo magic?

youngmoney:
the \n is actually a pythonism/other languages but not in bash in bash it is very literal. In bash one must use a real newline (and that works). It might be true that if you change the config a '\n' might work. (an assumption was made you are using bash, as that is my case)

youngmoney:
I'll write the test soon!

geier:
I think it's rather a Cism than a pythonism. But it looks like python escapes it itself. There is the `unicode_escape` escoding for usages like this, not sure if it is a good idea to actually use it though.

```
% echo 'x\ny'
x
y
% cat test2.py
import sys

arg = sys.argv[1]
print(arg)
print(arg.encode('utf-8').decode('unicode_escape'))
% python test2.py 'x\ny'
x\ny
x
y
```

geier:
sorry, hit the wrong button.

youngmoney:
I think we should leave it for now, unless we want to make \n usable everywhere

youngmoney:
I've fixed/added a test

I also as a bash sidenote:
```
$ echo 'x\ny'
x\ny
```

geier:
Thanks! (and I agree, we would need to make `\n` usable everywhere, but while nobody asks for it, it's not really worth discussing it).

Issue #469, opened 2016-08-13
Khal as a ics pretty printer


dopefishh:
Since khal has ics parsing and event printing it could be handy to add an option to print an ics file without importing.

`khal printics [-f] ics`

With this functionality khal can be used as a previewer in for example mutt.

In mutt you add something like this:
`auto_view text/calendar`
and in your mailcap:
`text/calendar; khal printics -f '%s'; copiousoutput;`
If necessary I can implement this myself but I thought I'd coin the idea first.

Best,

Mart

youngmoney:
I think this is a great idea and I would certainly use it. It would be great if you would implement it and I'd be happy to help, if not I will eventually get to it

geier:
I'm fine with this going into khal, I won't do it myself though (but I'm happy to review and help).

fpytloun:
+1
I am currently using this one: https://github.com/terabyte/mutt-filters/blob/master/vcalendar-filter

geier:
closed by #478

PR #478, opened 2016-08-17
add printics feature as mentioned in issue #469

dopefishh:


hobarrera:
LGTM. I think we need tests for this as well though.

untitaker:
Yes please, the new function has no test coverage. You can add a test in either https://github.com/pimutils/khal/blob/master/tests/controller_test.py or https://github.com/pimutils/khal/blob/master/tests/cli_test.py

geier:
feel free to ask for help if it isn't clear how to write some tests for this. Good work btw!

dopefishh:
I'll try to do this today or tomorrow. If there are any questions I'll let you guys know.

dopefishh:
I've added some tests (sorry it took so long). But it should be functional and tested now. While testing I've noticed that I had to change the config spec as well:)

dopefishh:
something went wrong with merging back into master. I'll fix this

dopefishh:
@geier all test succeed:)

geier:
Sorry for not having had a propper look at this. I'll try ASAP.

dopefishh:
NP, I was very late as well

khal/controllers.py
```diff
@@ -585,3 +586,25 @@ def import_event(vevent, collection, locale, batch, format=None, env=None):
...
+    if format is None:
+        format = conf['view']['agenda_event_format']
+    cal = icalendar.Calendar.from_ical(ics)
+    events = [item for item in cal.walk() if item.name == 'VEVENT']
+    events_grouped = defaultdict(list)
+    for event in events:
+        events_grouped[event['UID']].append(event)
+
+    if format is None:
+        format = conf['view']['event_format']
```
geier: I believe this second `if format is None` is superfluous, see line 591. Also according to the text in khal.spec `event_format` is the right format string, the current default just sucks.

geier:
one more note: please also add a new paragraph in `doc/source/usage.rst`. Thank you!

dopefishh:
I think I've squashed it correctly and reviewed your changes:)

geier:
οne more thing (sorry!): please run `make html` in the `doc` directory, that should produce a diffing configspec.rst file, and also check that in (you might need to install sphinxcontrib-newsfeed).

dopefishh:
np. Done

geier:
I'm sorry to  bother you again, but are the changes to the configwizard intentionally in this commit? They look suspiciously like my changes to the configwizard I commited recently, perhaps a merge/rebase gone wrong?

dopefishh:
Probably not... My git-fu is not that great so it seems...

geier:
an alternative might be to create a new branch (off master) and copying
over the relevant files and creating a new PR.

Quoting dopefishh (2017-01-10 11:00:45)

dopefishh:
at last:) this should do it

geier:
Thank you a lot and merged!

PR #483, opened 2016-08-21
Fix German locale tests on *BSD

geier:


tests/cal_display_test.py
```diff
@@ -137,6 +137,23 @@ def test_str_week():
     '    20 21 22 23 24 25 26 ',
     '\x1b[1mM\xe4r \x1b[0m27 28 29  1  2  3  4 ']
 
+example_de_bsd = [
```
untitaker: I think a regex might be nicer here since there is very little difference in the outputs.

pwnage101: I do think that it's clearer the way it is currently written (two separate variables).  It makes the dot version explicitly related to bsd, so the bread crumb trail is easier to follow.

If anything, I'd define a set of OS-specific abbrevations, e.g.:

```python
default_de_months = [ "Jan ", "Feb ", "M\xe4r ", ...]
bsd_de_months = [ "Jan.", "Feb.", "M\xe4rz", ...]
```

then substitute them into `example_de` and `example_de_bsd`.

pwnage101:
FWIW, debian gnu/linux and freebsd both apparently source from www.unicode.org.  Here is the latest data from unicode.org:

http://www.unicode.org/cldr/charts/29/by_type/date_&_time.gregorian.html#Months_-_abbreviated_-_Standalone

The abbrevations at that link match the ones FreeBSD uses.  It seems that whatever the OS uses is just a snapshot of whatever they pulled from unicode.org at release time.  In light of all of this, these tests seem quite inflexible, debian and guix may change again when they update their locale data.

/me shrugs

pwnage101:
I think we should change the tests to explicitly target versions of the CLDR, rather than specific operating systems.  In the link above, "29" represents the version.  CLDR version 29 matches FreeBSD and NetBSD, version 28 matches Debian and Guix.

In general, the tests should reflect the latest version of the CLDR, and some finite number of prior versions.  Currently it seems that all is covered by versions 28 and 29 (latest).

Thoughts?

pwnage101:
Nevermind!  It looks like `glibc` is itself a source of locale data...

https://sourceware.org/glibc/wiki/Locales#Creating_and_Updating_Locale_Data

That would explain why debian and guix locale data is currently matching.

geier:
@pwnage101 awesome thanks for following that up! I find the unicode table especially interesting.

On the main matter: I believe this is mostly bikeshedding... Does anyone of you want to take care of this or should I merge as is? I have no opinion in this matter.

untitaker:
I don't particularly care, if you want to merge it'd be fine.

pwnage101:
:+1:

geier:
fixed with 72c8efc929bd245ed233cc80cfc6d696cc511831

PR #484, opened 2016-08-22
Change 'themes' to 'colors' in docs

pfertyk:
It seems that the name of the file has changed, but the name in the docs hasn't. This should fix the problem.

According to the comment in `khal/docs/source/configspec.rst` this is the file that needs to be changed. If it isn't the right file, please let me know where should the change be applied.

### DIFF
--- a/khal/settings/khal.spec
+++ b/khal/settings/khal.spec
@@ -185,7 +185,7 @@ event_view_always_visible = boolean(default=False)
 #
 # This is very much work in progress. Help is really welcome! The two currently
 # available color schemes (*dark* and *light*) are defined in
-# *khal/ui/themes.py*, you can either help improve those or create a new one
+# *khal/ui/colors.py*, you can either help improve those or create a new one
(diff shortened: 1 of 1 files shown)


hobarrera:
Well spotted, thanks!

pfertyk:
Thanks for merging. I can see that the documentation still contains the old value `khal/ui/themes.py`:
https://khal.readthedocs.io/en/latest/configure.html
When will it be updated?

hobarrera:
It'll be updated when a new release is made.

geier:
No, by default readthedocs pulls from master branch on every push. After editing the spec file, the documentation should me `make`ed again, because the part on configuration gets extracted from the spec file.
@pfertyk Thanks for your attention!

hobarrera:
Oh, sorry about that, I assumed differently and didn't think to check.

geier:
no worries :)

Quoting Hugo Osvaldo Barrera (2016-08-25 16:37:18)

PR #485, opened 2016-08-24
Improve dark colors

untitaker:
With: https://asciinema.org/a/0u4x1653sntmee2vn70u6u6lv
Without: https://asciinema.org/a/6nwhnmyj4ct23na3u9wqc0biv

### DIFF
--- a/khal/ui/colors.py
+++ b/khal/ui/colors.py
@@ -37,8 +37,8 @@ dark = [
     ('today', 'dark gray', 'dark green',),
 
     ('date', 'light gray', ''),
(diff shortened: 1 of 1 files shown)


untitaker:
I just noticed that the current variant looks nicer on asciinema. I guess there's something wrong with my color scheme...

untitaker:
Actually it turns out that I don't have the same colorscheme set on asciinema. It rather looks like this: 
2016-08-24-185027_566x443_scrot

When focused:

2016-08-24-185119_676x302_scrot

geier:
Works for me. I'd actually love to get rid of the yellow, because it looks really awful in `standard` colors.

PR #489, opened 2016-08-25
NetBSDs German locale is different from FreeBSDs

geier:
I agree, this is getting somewhat ridiculous and a regexp might be a good idea. On the hand, if we ever need to change it, we might need to look up the differences again...

### DIFF
--- a/khal/calendar_display.py
+++ b/khal/calendar_display.py
@@ -31,12 +31,6 @@ from .terminal import colored
 setlocale(LC_ALL, '')
 
 
-def month_abbr(month_no):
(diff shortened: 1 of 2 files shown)


geier:
This whole thing  makes less and less sense to me, the variant on (Debian) Linux and FreeBSD actually seems to return a latin-1 encoded "Mär", even though we explicitly set a UTF-8 locale before (and on NetBSD it seems to do the right thing)...

```python
>>> b'M\xe4rz'.decode('latin-1')
'März'
```

geier:
best guess atm: it has something to do with UCS2 vs UCS4 builds of python

Issue #332, opened 2016-02-09
Feature request: make at command more intelligent


ralphholz:
Hi,

The current at command is relatively strict in what it accepts as a parameter. I think it would be nice if it could accept things like this - maybe with a new command "on":

khal on 11 - where it guesses the day is the 11th
khal on tomorrow - self-explanatory
khal on tomorrow+1, +2, +3, - day after tomorrow, day after that, etc.

Or maybe also:

khal in mar - show events in March etc.

I realise creating this software means a lot of work for the developer - so, first of all, many thanks for doing this. It really helps me in day-to-day's work. Second, I am happy to take a stab at this myself soonish if you're OK with it?

geier:
While some easy things like `khal at tomorrow` (and all the weekdays) already work, some improvements would be very welcome.
Also see #169 and #273

Issue #169, opened 2015-02-25
Use more convenient dates and times


bcudini:

geier:
This is a good idea and I've thought about this myself.

The problem with using GNU date is, that while it my be legal to import GPL licensed code I'm not sure its not against the wishes of the author. So I would prefer a MIT/BSD etc. licensed package to be used. An apache licensed module for which could be helpful is [parsedatetime](https://github.com/bear/parsedatetime).

For the where, have a look at `generate_random_uid()` in `khal/aux.py`.

If you have any more questions, don't hesitate to ask. I'd love to see a pull request for this.

untitaker:
I'd consider using GNU tools a valid option re legal situation, but doing so might harm portability.

untitaker:
You might also want to do this the way GNOME California does: Just take a description, and extract the dt from there if found. We should also add CLI options for enforcing the date with strict behavior (for scripts invoking khal new), and make the user use that if we fail to extract a date.

geier:
[California's quick add](https://wiki.gnome.org/Apps/California/HowToUseQuickAdd) looks pretty nice, lots of good ideas waiting to be stolen

bcudini:
The parsedatetime module looks good, california seems even better but their code is unusable here (looks like it's written in Vala and GPL licenced).

For the where, did you mean `construct_event()` in `khal/aux.py`, the function just after `generate_random_uid()`? I'm looking into it right now and will try to have a working PoC asap

geier:
@bcudini yes, I did mean `construct event()`, I don't know how that error happened (corrected it now)

geier:
I just had a look at california's [details parser](https://git.gnome.org/browse/california/tree/src/component/component-details-parser.vala). While it is obviously a lot better than what we currently have, their approach is at least similar: *Tokenize* the input string and then use different parsers for different components (start, end, duration, location etc) on that stack, popping elements if they are 'used'.

geier:
If you want to help with this issue, just write down all variants you can think of a clever quick add function should understand.

hobarrera:
AFAIK, California's implementation was rather broken in that it expected US-formated date (eg: "Monday 6pm") and did not work with standard time formats ("Monday 1800hs", "Monday 1800", etc), as well as some other really common variations.. However, the high level idea should work.

> If you want to help with this issue, just write down all variants you can think of a clever quick add function should understand.

Possible formats for input:

 * Beer with Alice tomorrow 1400
[...]

PR #491, opened 2016-08-29
WIP: Use parsedatetime for `khal list`

untitaker:


untitaker:
Example usage:

    khal list -e "one week"  # list seven days from one
    khal list -s "tomorrow"  # list events tomorrow
    khal list -s "tomorrow" -e "one week"  # list events from tomorrow to (tomorrow + 7 days = 8 days)

geier:
Some ideas, opinions, reservations:
* well tested code base (much better than our own)
* khal new and at should also use parsedatetime
* would probably make configuring of all `[long](date|time)formats` at the very
  least optionally (which is probably a good thing for most users)
* I dislike the `khal list -s 2017-4-4 -e 2017-4-12` format and much prefer
  `khal list 2017-4-4 - 2017-4-4` (which event works out of the box with
  evalRanges), but I'm not against making -s/-e optional
* as a `-` would be needed between start and end date (probably a good idea
  anyway), we should figure out a way to warn users for at least on release who
  enter start and end date without a `-`
* if setting a locale which isn't `en`, daynames, relative descriptions etc get
  translated as well (which is a big minus in my book)
* users can only set locale, no finer grained control over datetime in/output
  (but probably not necessary)
* with german locale settings cal.parse('1.10.') works, but
  cal.evalRanges('1.10.-10.10.') does not, strange

youngmoney:
I believe that at, new, and list all use the same function to parse dates and more (at least I think I wrote it that way) if true then this is where the parse date time should go

Just for some background, gcalcli allows optional parse datetime, if it is installed it will use it, otherwise it uses its own stricter parser (like ours) I say we try something like that, maybe even only use parse datetime if our parser fails and it is installed

Anyway just my opinion

untitaker:
>khal new and at should also use parsedatetime

I'd rather remove at in favor of this command (entering just the start date gets you the same behavior).

>would probably make configuring of all [long](date|time)formats at the very least optionally (which is probably a good thing for most users)

In todoman we fall back to the configured dateformat, also an option here.

>I dislike the khal list -s 2017-4-4 -e 2017-4-12 format and much prefer khal list 2017-4-4 - 2017-4-4

Strong disagree, makes the whole thing much more complex. Arguments are not optional. Options are optional.

>if setting a locale which isn't en, daynames, relative descriptions etc get translated as well

I didn't observe that behavior at all. If I set `LANG=de_DE.UTF-8`, weeknames in ikhal are translated, but parsedatetime behaves the same for the values `"in one week"`, `"thuesday"`, etc.

geier:
re locale: you are right, setting a environment variable works fine, I followed the example in the [documentation](https://github.com/bear/parsedatetime/blob/master/examples/with_locale.py) and than it doesn't.

geier:
I won't merge anything that removes the argument parsing, I'm not against having options as an alternative.

untitaker:
I don't understand why you insist on having your own argument parsing instead of using normal options. This makes the entire CLI harder to understand since it's a non-standard thing.

geier:
Quoting Markus Unterwaditzer (2016-09-06 22:31:48)
> I don't understand why you insist on having your own argument parsing instead of using normal options.

I do not insist on having my own argument parsing, I'll happily use any module that does this for us. Also, I do reject the notion that options are normal and arguments are not, e.g., see the classic UNIX utility cal(1).

So, why do I insist on it:
 1. I'm lazy, typing `khal list 2016-10-1 2016-10-20` is easier than typing `khal list -s 2016-10-1 -e 2016-10-20`, even worse when datetimes are given and need to be put into quotation marks.

 2. I strongly believe, that computer programs should strive to be able to understand unstructured user input (as long as its unambiguous). Of course, khal's commands currently do a very bad job at this, but options-only would be a step into the wrong direction in my opinion. See #169 and #332 for some examples of what khal should be able to understand.  I'd also like to be able to just pipe an email containing an event description into khal which should extract the relevant information itself.  Yes, this shifts complexity into the program, but I for programs I spend more time using than building, that's fine.


> This makes the entire CLI harder to understand since it's a non-standard thing.

Regarding the claim of that being non-standard, see cal(1) (by the way, the very first script called khal was a replacement for cal, because I disliked both cal's and ncal's output).  But, as said before, I'm happy to have khal's commands either accept both, arguments and options, or have two sets of commands, one that accepts only options and one that accepts arguments.


Quick anecdote only somewhat connected to any of this:

Just today, a colleague of mine looked at scipy's current documentation for its [Mann-Whitney-U Test](http://docs.scipy.org/doc/scipy-0.18.0/reference/generated/scipy.stats.mannwhitneyu.html) (function signature: `mannwhitneyu(x, y, use_continuity=True, alternative=None)`, where alternative can be one of None, ‘less’, ‘two-sided’, or ‘greater’). The colleague than went ahead and used it (wrongly) like this: `mannwhitneyu(x, y, 'two-sided')`, which produced a result, but was obviously not what the colleague wanted, who expected the function to sort out what was meant. While this is mostly a usage error (which can also be somewhat attributed to python's dynamic typing), to any human it is pretty clear what my colleague wanted to achieve, and the function *could* have figured out what the colleague wanted to achieve.
OK, I'm rambling now...

untitaker:
Arguments should be used for (1..n) parameters of the same type, options for optional parameters of different types. To see what I mean, look at

* `ls`, `cat` -- arguments are one or more files/directories. Options are there to modify the output format.
* `bash` -- arguments are one or more files. Options are there to change error behavior or verbosity.
* `find` -- arguments are one or more files. Options, while unnecessarily deviating from the norm (`-name` instead of `--name`) are there to filter out names or execute commands for each line.
* `diff` -- arguments are, again, files. Options are there to fine-tune behavior.
* anything else

Cal is different since it takes a day, month, or year as argument, but at least one could argue that it takes a single "date" arg, with varying precision.

>Quick anecdote only somewhat connected to any of this:

I'm going to interpret this as an argument for using keyword arguments everywhere. Coincidentally options are close to that :P

---

I have an alternative idea:

* Rename `-s` to `--from`
* Make `-e` an argument

    khal list --from tomorrow two weeks

    khal list --from "in two weeks" two weeks

untitaker:
I guess it wasn't important to me after all. Oh well.

PR #502, opened 2016-09-06
Remove MyBoxAdapter

untitaker:
The height calculation is possible without patching classes from the
parent widget.


untitaker:
I added a commit that improves the spacing in DayWalker: https://asciinema.org/a/7zz6pi0ftz7chdlkxtubl28kn

geier:
Thanks.

Issue #436, opened 2016-06-04
ikhal creates empty field for LOCATION and DESCRIPTION


Brixy:
Hi!

Just realized that some GUI programmes print “notes:” oder “location:”, even if no value is set.

When you create a new event using `ikhal` (unlike `khal`) without values for DESCRIPTION and/or LOCATION, in the .ics file you will find

```
DESCRIPTION:
LOCATION:
```

It would probably be better not to create these two keys, if no values exist.

Thanks.

geier:
Probably a good idea, patches are welcome (and this should be a rather easy patch).

geier:
closed with #504

PR #504, opened 2016-09-07
Prevent saving empty location and description in ICS file

pfertyk:
This should fix issue #436.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -25,4 +25,4 @@ Stephan Weller
 Max Voit - max.voit+dvkh [at] with-eyes [dot] net
 Taylor L Money - - http://taylorlmoney.com
 Troy Sankey - sankeytms [at] gmail [dot] com
-Paweł Fertyk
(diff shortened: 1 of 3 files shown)


geier:
Looks good, thanks! Please also add yourself to AUTHORS.txt, I don't believe you are in there yet, are you?

geier:
While thinking about this, we also need to remove the current location/description if the update changes this to `''`.

geier:
Sorry, misclicked.

geier:
*Should* work something like this: self._vevents.pop('LOCATION'), but a test would be great as well.

pfertyk:
I added more tests for removing the existing location/description if set to empty and I implemented this feature.

geier:
Looks good. Can you rebase onto master? (git rebase master)

pfertyk:
I rebased (and added a missing email address). Could you confirm that this is what you meant?

geier:
Yes, that's what I meant. Thanks!

PR #508, opened 2016-09-19
Add cli option for eol character for khal_list

greole:
in some situations I prefer to list events next to each other and save some screen space.

### DIFF
--- a/khal/aux.py
+++ b/khal/aux.py
@@ -35,6 +35,14 @@ import pytz
 from khal.log import logger
 from khal.exceptions import FatalError
 
+def intersperse(iterable, element):
+    """ intersperse an iterable with element
+    """
+    it = iter(iterable)
+    yield next(it)
+    for x in it:
+        yield element
+        yield x
 
 def timefstr(dtime_list, timeformat):
     """converts a time (as a string) to a datetimeobject
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -295,12 +295,14 @@ def _get_cli():
     @click.option(
         '--once', '-o', is_flag=True,
         help='Print events only once, even if they span multiple days')
+    @click.option('--event-separator', '-e',
+        help='inserts separator between calendar events')
     @click.option('--notstarted', help=('Print only events that have not started'),
                   is_flag=True)
     @click.argument('DATERANGE', nargs=-1, required=False,
                     metavar='[DATETIME [DATETIME | RANGE]]')
     @click.pass_context
-    def klist(ctx, daterange, once, notstarted, format, day_format):
+    def klist(ctx, daterange, once, notstarted, format, day_format, event_separator):
         """List all events between a start (default: today) and (optional)
         end datetime."""
         controllers.khal_list(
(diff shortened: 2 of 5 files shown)


geier:
Hi @greole and thank you for making khal better!

Would you be using this interactively or with a script? (In the latter case, some awking should do the trick). 

I'm actually a bit hesitant to accept this, because adding this new option also increases the complexity. 
@hobarrera @untitaker opinions?

If we are to merge this, we some tests would be really nice though.

untitaker:
I think you can just pipe through a command which replaces newlines with something else...

greole:
@geier @untitaker  valid points!

However ultimately it would be great if one could lists events like

      Sunday:  Event 1 Event 2
      Monday:  Event 3 Event 4  Event 5
      Tuesday: Event 6

Anyways,  thanks a lot for khal :+1:!

geier:
This looks really good, especially if you have only a few events, perhaps we *should* get this in, as most other aspects of event printing are already fully configurable. 

As I said above, if you have the time, please write some tests for the new behavior.

untitaker:
Then I think we should call it `event-separator` or something like that, not eol.

greole:
I could imagine different solutions. For example the separators `A`,`B` and `C` in 

      Sunday A Event 1 B Event 2 C
      Monday A Event 3 B Event 4 B Event 5 C
      Tuesday A Event 6 C
could be made modifiable. Then if A=:\n and B=, and C=\n

      Sunday:
      Event  1, Event 2
      Monday:
      Event 3, Event 4, Event 5
      Tuesday:
      Event 6
or 

      Sunday: Event 1 Event 2
      Monday: Event 3 Event 4 Event 5
      Tuesday: Event 6

for  A=: and B= and C=\n etc. Concerning the CLI, the args could be specified as

**Version1**
      --event-separator "\n" "," "\n" # needs many colons 

and click seems not to support a mixture of given and default options, hence
     
      --event-separator "\n" "," # invalid
**Version2**
Some kind of formatting string

      --event-separator "\n, ,\n" # coma separated separators
or something that can be parsed with regex like

     --event-separator "{\n}{,}{\n}"

**Version3**
 Each separator would get an additional option.

What do you think?

geier:
I believe that if we are going to do this, we should go all the way and make it completely configurable (as you suggested).

What way to specify those works best, I have no idea. Feel free to experiment.
We could think about removing option A (and make it a part of the day-format option), but than it would be less consistent with B and C.

greole:
Ok in the latest PR newlines can be added explicitly by the  `day-format`  parameter.  `B` can be specified by --event-separator and a true `C` would need a new option if really necessary . A line break after each event  can also be specified by the `format` arg. I wrote a first (basic) test for the event-separator and adapted some tests.

hobarrera:
What's the status for this?

geier:
Closing for row, reopen this (or a new PR if you are still interested in this).

hobarrera:
Rather than include such complex settings and flags for so much output customization, maybe we can do something similar to todoman's `--porcelain`, which outputs everything in an easily-parsable format (JSON), so scripts can reformat this as needed.

I'd use this, for example, to show my next event in my i3 status bar. Any other weird formatting need can just be a wrapper around `khal --porcelain`.

I'll try to see if I can work on this after I've finished #611

PR #510, opened 2016-09-21
Allow to pass arguments to sqlite3.connect

teto:
For instance now in my code, I can do:
`        self.collection = khal.cli.build_collection(self.config, None, check_same_thread=False)`
to prevent exceptions from being read. According to https://github.com/ghaering/pysqlite/issues/106, as long as each thread owns its own connection, it shouldn't be a problem.

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -160,7 +160,7 @@ def global_options(f):
     return config(verbose(color(version(f))))
 
 
-def build_collection(conf, selection):
+def build_collection(conf, selection, **kwargs):
     """build and return a khalendar.CalendarCollection from the configuration"""
     try:
         props = dict()
@@ -182,6 +182,7 @@ def build_collection(conf, selection):
             default_color=conf['highlight_days']['default_color'],
             multiple=conf['highlight_days']['multiple'],
             highlight_event_days=conf['default']['highlight_event_days'],
+            **kwargs,
         )
     except FatalError as error:
         logger.fatal(error)
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -88,16 +88,17 @@ class SQLiteDb(object):
                     None, a place according to the XDG specifications will be
                     chosen
     :type db_path: str or None
+    :param kwargs: These additional parameters will be forwarded to sqlite3.connect
     """
 
-    def __init__(self, calendars, db_path, locale):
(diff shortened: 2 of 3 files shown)


rogerbinns:
You should not disable the same thread check!  In python you do not control the garbage collector, and non-closed pysqlite objects can very easily end up having their destructors run in different threads.  Those destructors will make SQLite calls.  That is part of the reason why the thread check exists.

teto:
The main thread just opened the connection, ran a new thread that would do the calls with that connection. The main thread shouldn't do any sqlite calls  so there shouldn't be a problem  ?
Anyway rather than modifying khal, I will just modify my module so that it opens the connection within the main thread. Simpler and certainly safer in the long run.
Thanks for the warning.

geier:
@rogerbinns Thank you for your input! May I ask how you stumbled upon this PR? Are you actively looking for people mis-using SQLite?

@teto I guess this doesn't solve you problem at all, does it? I believe we need to find a better solution to run khal in i3pystatus than you keep your own fork of it.

teto:
No it's fine, in fact I just needed to open the database in the thread rather than on i3pystatus launch. So I dont need any ikhgal modifications at all. Right now I just need some advice on khal's api to improve my i3pystatus module :)

rogerbinns:
@geier I do an alternate binding of SQLite to Python (APSW) (differences to pysqlite [documented here](https://rogerbinns.github.io/apsw/pysqlite.html)).  I am subscribed to pysqlite issues so I can check that bugs/issues there don't affect APSW.  The pysqlite maintainer and myself manage the [python-sqlite group](https://groups.google.com/forum/#!forum/python-sqlite).

rogerbinns:
@teto An approach I use is sticking the connection in threading.local.  A convenience method gets the existing connection, or makes a new one for the thread.

One thing to be *extremely* careful of is if the process ever does a fork.  (You may not do it manually, but multiprocessing does.) That will trash database locks etc, hence content too.

geier:
@rogerbinns would opening the database in read-only mode help with these issues?

rogerbinns:
@geier read-only opening will not help since it doesn't change threading behaviour.  Additionally read-only connections can still do writing.  A common case is if the previous opener of the database exited while in a transaction.  The next opener has to rollback the incomplete transaction.

geier:
@rogerbinns Thank you!

PR #511, opened 2016-09-25
Fix colors for new urwid version

untitaker:


untitaker:
Sorry for the vague information, I can't figure out which version of urwid I had previously installed.

Before upgrading urwid:

2016-09-25-150316_468x60_scrot

After upgrading urwid:

2016-09-25-150737_460x49_scrot

This PR reverts to the previous style.

hobarrera:
LGTM. Looks like the old code worked only due to some issue in urwid.

geier:
Thanks!

PR #530, opened 2016-11-24
vdir calendars: use color from config if it's not defined in the vdir

mkobel:
With the following config, the color setting did not work:

```
[[my_calendar]]
path = ~/.calendars/my_calendar
type = discover
color = dark blue
```

Therefore I modified the code to use the color setting as fallback if there is no color defined in the directory.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -26,3 +26,4 @@ Max Voit - max.voit+dvkh [at] with-eyes [dot] net
 Taylor L Money - - http://taylorlmoney.com
 Troy Sankey - sankeytms [at] gmail [dot] com
 Paweł Fertyk - pfertyk [at] openmailbox [dot] org
+Moritz Kobel - moritz [at] kobelnet [dot] ch - http://www.kobelnet.ch
--- a/khal/settings/utils.py
+++ b/khal/settings/utils.py
@@ -175,7 +175,7 @@ def config_checks(
         if config['calendars'][calendar]['type'] == 'discover':
             vdir = get_all_vdirs(config['calendars'][calendar]['path'])
             vdirs += vdir
-            if len(vdir) >= 1:
+            if len(vdir) >= 1 and 'color' in config['calendars'][calendar]:
                 vdir_colors_from_config[vdir[0]] = config['calendars'][calendar]['color']
             config['calendars'].pop(calendar)
     for vdir in sorted(vdirs):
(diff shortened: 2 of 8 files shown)


geier:
This does look like a good idea! Could you perhaps add a test (have a look at `tests/settings_test.test_config_checks())?  Also, please add yourself to AUTHORS.txt.

And sorry that you had to fix those blank line errors.

hobarrera:
This sounds a bit redundant. I'd rather get rid of the config variable, and only use the colour from the vdir metadata.

geier:
@hobarrera I would agree, were it not for the fact that only hex colors are allowed in the vdir's color files (for obvious reasons).

hobarrera:
@geier Are you sure?

```
$ cat .local/share/calendars/502708f4-4b94-4963-b042-f7340c829bdf/color
#8ab6d2
```

This syncs with my caldav server file (using vdirsyncer) and works with khal, todoman and my mobile apps (both fastmail's and iOS's calendar).

mkobel:
@hobarrera This does not work with read-only caldav resources.

hobarrera:
@mkobel Do you mean that colours don't work for those?

mkobel:
@hobarrera I can create a color file for those, but it is not synchonized. Therefore my khal configuration is shared (workstation, notebook) to ensure the same colors.

hobarrera:
@mkobel Well, yeah, metadata for read-only calendar won't be sync'd, of course.

I still don't like the idea of overriding per-calendar metadata in the settings file, just so users can specify metadata for read-only calendars, and then sync the configuration through other means.

It's a feature that's _only_ useful in very specific, and complex scenarios.

geier:
@hobarrera I meant that *only* #abcdef0 work, but not color names.

mkobel:
By enabling the metadata sync of the colors of all our company calendars I saw an additional problem: There are cases where I would like to overwrite the color of the calendar server. 

To have the custom calendar colors synced on several computers I think defining the color in the configuration is an appropriate way (My config files are synced using Seafile).

geier:
@mkobel sorry it took me so long to have a look at the code. I believe I found a little bug, that would only make the first discovered calendar without a color file use the default color. Please have a look at [branch mkobel_master](https://github.com/pimutils/khal/tree/mkobel_master), and see if the changes I made are what you intended.

mkobel:
@geier Thank you for the review and the changes. They look good for me.

geier:
I've pushed those commits, but because I rebased, you won't get an email that the PR has been merged. You should still be properly attributed, e.g., see https://github.com/pimutils/khal/commit/2290b6612eeb44c10940607f7517d990300184b2

PR #543, opened 2016-12-19
Reproduce 529: etags which should change are not changing

geier:
Only really slow drive and busy drives (spinning rust) I could reproduce Debian's [issue 844081](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844081).

What seems to be the issue is, that the mtime fstat reports might not have changed yet, if the file changes are still in some kernel buffer. A call to os.sync() before checking a vdir's etag makes sure all changes have been written to disk. I found a similar issue with vdir.get_etag_from_file(), where an os.fsync() before getting a file's etag remedies the situation.

Calling os.sync() on every call of khal is probably is not ideal, but the only alternative I currently see would be making sure that all files in all vdirs are all synced (by calling fsync() on them all), which would be easy for these tests but impossible to enforce for other programs manipulating vdirs, e.g. vdirsyncer.

### DIFF
--- a/khal/khalendar/vdir.py
+++ b/khal/khalendar/vdir.py
@@ -56,28 +56,31 @@ def _generate_href(uid=None, safe=SAFE_UID_CHARS):
 
 
 def get_etag_from_file(f):
-    '''Get mtime-based etag from a filepath or file-like object.
+    '''Get mtime-based etag from a filepath, file-like object or raw file
+    descriptor.
 
     This function will flush/sync the file as much as necessary to obtain a
     correct mtime.
     '''
+    close_f = False
+    if hasattr(f, 'read'):
+        f.flush()
+        f = f.fileno()
+    elif isinstance(f, str):
+        flags = 0
+        if os.path.isdir(f):
+            flags = os.O_DIRECTORY
+        f = os.open(f, flags)
+        close_f = True
+
     # assure that all internal buffers associated with this file are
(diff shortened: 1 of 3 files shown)


geier:
@untitaker if you have the time, please have a look at the above changes. I'd like to hear your opinions on this.

khal/khalendar/khalendar.py
```diff
@@ -235,6 +235,7 @@ def update_db(self):
 
         should be called after every change to the vdir
         """
+        os.sync()
```
untitaker: This should never be necessary and also has severe perf impacts.

geier:
I know that os.sync() isn't a great approach. And no, without that os.sync() the tests fail. The problem is, even when we write to the vdir, we don't enforce writing to disks.

For the purpose of the tests, the os.sync() can just go into the tests (and perhaps it better should). In practice, one probably doesn't call khal that shortly after editing the vdir, so it shouldn't really be a problem.

untitaker:
Try fsyncing the parent directory.

geier:
The above commit seems to work, might need some refactoring.

untitaker:
For explanation: IIRC to fsync metadata of a file, you need to fsync the parent directory

geier:
I'm merging this, because it's at least better then what we currently have, we can get rid of the some more kinks in master.

PR #551, opened 2017-01-13
If DTSTART == DTEND, consider event is 1h instead of 1d. Fix #444

### LINKED ISSUE #444: iCal, DTSTART == DTEND
Currently, when DTSTART == DTEND khal shows things like:

```
Today:
20:00→ : "blah foo bar"
Tomorrow:
  → 20:00: "blah foo bar"
```

Google Agenda assume that event last 1 hour.
Neither behavior is perfect and both are better than omitting the event.
But khal output seems more confusing and by assuming that event last 24h give the visual impression that 2 distinct event happens (especially if other events happens in the timespan).

Assuming up the end-of-the-day, or, like Google 1 hour, could be better trade-off.

nim65s:
See discussions in #444


geier:
looks good, thanks! Could you please add a warning to the user?

hobarrera:
LGTM

geier:
Thanks and merged.

nim65s:
Thanks :)

PR #566, opened 2017-02-06
Use the system locale's formats as a default

hobarrera:
By default, use the system locale for date and time formats. This should (generally) provide defaults with which the user is more familiar, without affecting reconfigurability at all.

This _should not_ change anything for users that have specified a format, and does not limit and configuration options at all.


hobarrera:
Bump?

If this sounds like a bad idea, a single config flag like `user_system_locale` might cover that activates this behaviour might be a good option too.

geier:
please add an entry to the CHANGELOG

hobarrera:
Oops, I'd pushed this to the wrong branch.

hobarrera:
Bump!

geier:
I haven't forgotten about this and will merge this eventually. I am still hesitant to merge this at the moment, as it it breaks the current defaults. Same goes for #585.

geier:
this is blocked by #651

PR #567, opened 2017-02-06
Remove superfluous space from tox.ini

hobarrera:
This space made environment generation bugging (leading space), and running `tox` (with no env specified) failed (due to a path being generated with the leading space in it).


untitaker:
Hmm, that might be a bug in tox. At least I can't believe this is expected behavior.

hobarrera:
Probably, though I don't think they'd support envs with spaces.

geier:
thanks

PR #570, opened 2017-02-11
[RFC] fix `new DATE TIMEDELTA`

### LINKED ISSUE #568: khal new $DATE $TIMDELTA creates events that are one day to long
Adding what I thought was going to be a 3 day event:

      khal new -a test 2017/03/10 3d Visit

produces a 4 day event:

    BEGIN:VCALENDAR
    VERSION:2.0
    BEGIN:VEVENT
    SUMMARY:Visit
    DTSTART;VALUE=DATE:20170310
    DTEND;VALUE=DATE:20170314
    DTSTAMP;VALUE=DATE-TIME:20170206T144724Z
    UID:4UL8DT8GO43ZLZHTQRXJ7U4E2WV9MUEP59MP
    END:VEVENT
    END:VCALENDAR

(note: DTEND is exclusive for allday events)

geier:
fixes #568

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -39,7 +39,7 @@ import textwrap
 
 from khal import utils, calendar_display
 from khal.khalendar.exceptions import ReadOnlyCalendarError, DuplicateUid
-from khal.exceptions import InvalidDate, FatalError
+from khal.exceptions import FatalError
 from khal.khalendar.event import Event
 from khal.khalendar.backend import sort_key
 from khal import __version__, __productname__
@@ -88,7 +88,7 @@ def calendar(collection, agenda_format=None, notstarted=False, once=False, dater
         day_format = conf['view']['agenda_day_format']
 
     term_width, _ = get_terminal_size()
-    lwidth = 25
+    lwidth = 25  # TODO add two if weeknumbers = right
     rwidth = term_width - lwidth - 4
 
     start, end = start_end_from_daterange(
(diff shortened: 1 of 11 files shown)


khal/controllers.py
```diff
@@ -328,11 +328,16 @@ def new_from_string(collection, calendar_name, conf, info, location=None,
...
+    try:
+        info = utils.eventinfofstr(
+            info, conf['locale'], default_timedelta="60m", adjust_reasonably=True, localize=False)
+        new_from_args(
+            collection, calendar_name, conf, format=format, env=env,
+            location=location, categories=categories, repeat=repeat,
+            until=until, alarms=alarms, **info
+        )
+    except FatalError:
+        sys.exit(1)
```
untitaker: I dislike this. Can we introduce a new exception type instead and catch that further up?

khal/settings/utils.py
```diff
@@ -49,6 +50,13 @@ def is_timezone(tzstring):
         raise VdtValueError("Unknown timezone {}".format(tzstring))
 
 
+def is_timedelta(string):
```
untitaker: Why is this called `is_timedelta` if it doesn't return a boolean? Something like `parse_timedelta` would be more appropriate.

khal/utils.py
```diff
@@ -370,15 +376,7 @@ def guessrangefstr(daterange, locale, default_timedelta=None, adjust_reasonably=
...
-                raise
-                #start = datetime_fillin(end=False)
-
-            elif start.lower() == 'week':
-                    today_weekday = datetime.today().weekday()
-                    start = datetime.today() - \
-                        timedelta(days=(today_weekday - locale['firstweekday']))
-                    end = start + timedelta(days=7)
-                    return start, end, True
+                raise  #  used to be: start = datetime_fillin(end=False)
```
untitaker: Why keep the comment?

khal/settings/utils.py
geier: consistency with `is_timezone()`, but I agree, this isn't great naming.

PR #575, opened 2017-02-14
Fix tests on systems with second-precision mtime

hobarrera:
Some systems have an mtime with precision in seconds, so tests what expect changes in mtime fail.

Update these tests to sleep longer on those systems (but still sleep 0.01son faster systems).

Closes #574

PR #577, opened 2017-02-15
Tweak sleep_time to be as precise as possible

hobarrera:
Tweak sleep_time to be as precise as possible, and also account for possible rounding error when our sleep is equal to the maximum precision (basically, for rounding in our favour).

See [this discussion](https://github.com/pimutils/khal/pull/575#pullrequestreview-22009618).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -70,7 +70,7 @@ def sleep_time(tmpdir_factory):
         return getattr(stat, 'st_mtime_ns', stat.st_mtime)
 
     i = 0.00001
-    while True:
+    while i < 100:
         # Measure three times to avoid things like 12::18:11.9994 [mis]passing
         first = touch_and_mtime()
         sleep(i)
@@ -81,3 +81,9 @@ def sleep_time(tmpdir_factory):
         if first != second != third:
             return i * 1.1
         i = i * 10
+
+    # This should never happen, but oh, well:
+    raise Exception(
+        'Filesystem does not seem to save modified times of files. \n'
+        'Cannot run tests that depend on this.'
+    )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -54,6 +54,8 @@ def never_echo_bytes(monkeypatch):
 @pytest.fixture(scope='session')
 def sleep_time(tmpdir_factory):
     """
+    Returns the filesystem's mtime precision
+
     Returns how long we need to sleep for the filesystem's mtime precision to
     pick up differences.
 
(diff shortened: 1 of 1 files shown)


untitaker:
It appears your last value of `1.1` is still not enough. I'd try starting with `0.001` and adding `0.001` until the mtime changes.

hobarrera:
This new change should be safer.

untitaker:
Hm, now this will return None on second-precision fs I think

untitaker:
Suggestion:

    i = 0.000001
    while True:
        a = touch_and_mtime()
        sleep(i * 1.1)
        b = touch_and_mtime()
        if a != b:
            return i * 1.1
        i *= 10

hobarrera:
Returning `i * 10` sounds safer because we're sure our sleeps are an order of magnitude above the timestamp precision. 
This is 10ms on most modern systems, so the entire test suite doesn't even increase in a single second.

hobarrera:
As another alternative, we can `mock` the result of `os.stat`. This sounds like the best idea, and doesn't involve any sleeping.

untitaker:
>Returning i * 10 sounds safer because we're sure our sleeps are an order of magnitude above the timestamp precision.

And it's 10 seconds on a system with second-precision

geier:
I don't like mocking this, I'd go with untitaker's solution and wait till it fails somewhere again.

hobarrera:
Back to this PR then, [this](https://github.com/pimutils/khal/pull/577#issuecomment-280055820) suggestion won't work.

Assume that precision is 100ms:
* Current time is 11:42:20.99999
* Record `a`
* Sleep 11ms
* Record `b`
* We noticed the `mtime` differ
* Okay, we return 11ms (WRONG!)

The chances of this happening are actually about 10%, so it's something worth considering.

This should work though:

```python
i = 0.000001
while True:
    a = touch_and_mtime()
    sleep(i)
    b = touch_and_mtime()
    sleep(i)
    c = touch_and_mtime()
    if a != b != c:
        return i * 1.1
    i *= 10
```

This fixes the above scenario. Chances of failure because of the current time are practically inexistant.

---

Note (mostly for the fun of it): This can fail on [leap seconds](https://en.wikipedia.org/wiki/Leap_second). Seriously, I don't care this means the loop has to be run *at an exact time* (with sub-second precision) in a window that only available once every year.

untitaker:
Good point and I agree that your solution is excellent.

hobarrera:
Thanks! PR updated.

untitaker:
We probably should raise an exception if `i > 10`

geier:
Thank you both! 

Am 18. Februar 2017 18:53:34 MEZ schrieb Hugo Osvaldo Barrera <notifications@github.com>:
>Merged #577.
>
>-- 
>https://github.com/pimutils/khal/pull/577#event-968070359

-- 

PR #584, opened 2017-02-20
Slight documentation clarifications.

### LINKED ISSUE #582: show more events in calendar and agenda
There are times when I only have one or two events a week, so I only use khal very infrequently. Nevertheless, I'd like to be reminded about future events, even when they're far away, so I don't miss them :-) Also, navigating around the agenda is very cumbersome, you have to give multiple `agenda <date>` to see all of them, because normally, the `calendar` and `agenda` views show very few events, mostly only for "today" and "tomorrow". Instead, they could show the next _n_ events in the future, and choose _n_ to be a number that fills the height of the calendar in `calendar` view, or around ~10 (a configurable number maybe?) in `agenda` view.

hobarrera:
Closes #582


geier:
Perhaps we need to move the formating stuff out of view, some general restructering might be in order..

geier:
I didn't mean to create that merge commit here, I merged by hand.

geier:
@hobarrera just so that you also can forget about this: please run `make html` after changing khal.spec and check in the resulting rst. If you are able to figure out how to do it on readthedocs so that we don't need that file checked in, I'd be very happy :)

hobarrera:
Right, that totally slipped my mind.

AFAIK, you can't run custom script, but you *can* provide a [custom `conf.py`](https://readthedocs.org/dashboard/todoman/advanced/). Since it's a python script, you can just make it do the same thing that `generate_config.py` does.

This since requires tinkering on the readthedocs side, I'll experiment on a fork a bit and let you know (I'd also love to do this for `todoman` 😉 ).

PR #588, opened 2017-02-21
Auto-generate configspec.rst

hobarrera:
FWIW, this is the only difference between the previous `rst` and the one generated on-demand now.

```
$ git diff doc/source/configspec.rst
diff --git a/doc/source/configspec.rst b/doc/source/configspec.rst
index 9d6173f..ea4b466 100644
--- a/doc/source/configspec.rst
+++ b/doc/source/configspec.rst
@@ -1,10 +1,3 @@
-..
-    PLEASE DO NOT EDIT THIS FILE
-    ############################
-
-    This file gets generated from `khal/khal/settings/khal.spec`. Please edit
-    that file instead and then run `make` in `khal/doc/'.
-

 The [calendars] section
 ~~~~~~~~~~~~~~~~~~~~~~~

```

See #582


geier:
Great! it even builds on rtd, what current master doesn't (even though I restarted it twice already).

Issue #486, opened 2016-08-25
khal at broken in 0.8.2


milouse:
khal at raise an error when used with a complete datetime

```
$ khal at "27/08/2016"

(works as expected)
$ khal at "09:00"

(works as expected)
$ khal at

(works as expected)
$ khal at "27/08/2016 09:00"
Traceback (most recent call last):
  File "/usr/bin/khal", line 11, in <module>
    load_entry_point('khal==0.8.2', 'console_scripts', 'khal')()
  File "/usr/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/lib/python3.5/site-packages/khal/cli.py", line 462, in at
    dtime, _ = aux.guessdatetimefstr(dtime_list, locale)
  File "/usr/lib/python3.5/site-packages/khal/aux.py", line 207, in guessdatetimefstr
    dtstart = fun(dtime_list, dtformat)
  File "/usr/lib/python3.5/site-packages/khal/aux.py", line 71, in datetimefstr
    dtime_list.pop(0)
IndexError: pop from empty list
```

I fix this issue by adding the following lines in /usr/lib/python3.5/site-packages/khal/aux.py at line ~71

```python
if not dtime_list:
    continue
```

The same in context (from line 57)

```python
def datetimefstr(dtime_list, dtformat):
    """
    converts a datetime (as one or several string elements of a list) to
    a datetimeobject

    removes "used" elements of list

    :returns: a datetime
    :rtype: datetime.datetime
    """
    parts = dtformat.count(' ') + 1
    dtstring = ' '.join(dtime_list[0:parts])
    dtstart = datetime.strptime(dtstring, dtformat)
    for _ in range(parts):
        if not dtime_list:
            continue
[...]

PR #600, opened 2017-03-07
Be more precise about the at command #486

milouse:
I finally found time to edit the man page.

Do whatever you want with this. I found it usefull, thus maybe it could be for other :)


geier:
Thank you for the fix!

PR #602, opened 2017-03-09
More pythons and CI stuff

hobarrera:
Test out ~pypy3~ and python3.7 (the latter won't make CI fail, it's mostly informative so we keep an eye on forward compatibility).

Also make sure docs build on CI so we never merge broken docs into master.

Finally, reduce the CI matrix a bit, because otherwise with these changes it would be huge, while adding little value.

**EDIT:** Nope, khal does not seem to work on pypy3. [CI results](https://travis-ci.org/hobarrera/khal/jobs/209199316)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
     - 3.4
     - 3.5
     - 3.6
+    - 3.7-dev
 
 env:
     - BUILD=tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ python:
     - 3.7-dev
 
 env:
-    - BUILD=tests
-    - BUILD=tests-negtz
+    - BUILD=py
+    - BUILD=negtz
 
 matrix:
     include:
@@ -26,4 +26,4 @@ install:
         - "pip install tox"
 
 script:
-    - "tox -e py-$BUILD"
+    - "tox -e $BUILD"
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,11 @@ python:
 env:
     - BUILD=tests
     - BUILD=tests-negtz
-    - BUILD=style
+
+matrix:
+    include:
+      - python: 3.6
+        env: BUILD=style
 
 addons:
     apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ matrix:
     include:
       - python: 3.6
         env: BUILD=style
+      - python: 3.6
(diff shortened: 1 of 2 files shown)


hobarrera:
Sadly, I had to remove pypy3, because it doesn't work. If anybody want to actually work on making it work, the relevant diff is merely this:

```diff
diff --git a/.travis.yml b/.travis.yml
index f07a30c..6d0563b 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ python:
     - 3.5
     - 3.6
     - 3.7-dev
+    - pypy3.3-5.2-alpha1

 env:
     - BUILD=tests
```

Note that newer pypy3 versions are not available in travis. Though manual installation is possible, I feel like it makes all the CI setup too complex needlessly.

hobarrera:
This is ready for review, BTW (not sure if I should tag it `ready`).

geier:
Thanks, looks good.
I never used the ready tag myself, I gets auto-added by waffle (in my name).

Two notes:

* does python 3.7 ever break because of their problems? If not, let's not allow failures, so we catch errors earlier than the python 3.6 bug in khal (or do we still get emails if the test does fail?).
* I don't think the negtz tests, because we never actually read $TZ, all timezone settings are either explicitly set or set by tzlocal, which, AFAIK, only reads /etc/localtime.

hobarrera:
`tzlocal` actually does read `$TZ`. I tested it using this script:

```python
import os
from datetime import datetime

from dateutil.tz import tzlocal

print(os.environ.get('TZ'))
print(datetime.now(tz=tzlocal()))
```

Example 1:

```$ python test.py
UTC
2017-03-21 03:27:53.938673+00:00
```
Example 2:

```
$ python test.py
None
2017-03-21 00:28:06.584555-03:00
```

hobarrera:
Regarding python3.7, nope, I haven't seen it break, so I've removed it from `allow_failures`.

geier:
While we are not using dateutil's timezones but pytz's and the tzlocal package, it behaves the same. My bad. Still not sure if we need this, because I don't think we let it choose the timezone itself in any tests, it won't cause any harm.

Thanks!

hobarrera:
Back when we introduced it, it detected some bugs assuming that "a certain day has started if it's past midnight UTC".

This is true for all timezones east of UTC, but failed during a few hours a day for the rest. IMHO, I'd prefer to keep it to make sure we never regress on that again. :)

PR #607, opened 2017-03-21
Fix viewing/editing logic in ikhal

ssiegel:
Test case for the unexpected behavior which is fixed by this commit:

Show the first event of some day, then
- with event_view_always_visible = False:
  press Escape, Up, Down, Enter
- with event_view_always_visible = True:
  press Up, Down

The expected result is to just view the event, but previously the edit form was opened.

### DIFF
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -913,12 +913,12 @@ class EventColumn(urwid.WidgetWrap):
                 if self.delete_status(self.focus_event.recuid):
                     self.pane.window.alert(('alert', 'This event is marked as deleted'))
                 self.edit(self.focus_event.event)
+            elif key in self._conf['keybindings']['external_edit']:
+                self.edit(self.focus_event.event, external_edit=True)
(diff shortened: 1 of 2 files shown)


khal/ui/__init__.py
```diff
@@ -904,23 +907,18 @@ def keypress(self, size, key):
 
         rval = super().keypress(size, key)
         if self.focus_event:
-            if key in self._conf['keybindings']['view'] or \
+            if key in self._conf['keybindings']['view'] and \
+                    prev_shown == self.focus_event.recuid:
+                # the event in focus is already viewed -> edit
+                if self.delete_status(self.focus_event.recuid):
+                    self.pane.window.alert(('light red', 'This event is marked as deleted'))
```
geier: please change 'light red' to 'alert', even though this doesn't (and didn't) do anything.

ssiegel: Of course you're right, this slipped through when going from latest release to current master 😳

geier:
Thanks and merged.

Issue #609, opened 2017-03-23
Draw khal with urwid as well


hobarrera:
So this is an odd proposal, I think it merit some discussion, but could result in some big [positive] changes.

I've been fiddling around with `urwid`s event loop, alarms, alike, and I've found some hacks that would maybe allow us to draw `khal` using urwid as well.

The dumbest and simplest example I've gotten working is have urwid call `os._exit(0)` right after drawing, which just leaves the initial draw on screen.
A sophisticated approach would be to write a dummy screen that prints the first redraw and forces an exit.

Before actually moving into this, I think it warrants some discussion (also, I've to ask _"would this get merged?"_).

### Pros

* De-dupe a lot of code. Rather than having one code block to draw calendars in khal and another in ikhal, share all that code. The de-duplication could be huge.
* *A lot* less code to maintain
* Forget about fine-tuning columns, colour overflows and alike for `khal`.

### Cons

* My dumb approach using `unittest.mock.patch` doesn't work with `less`, or anything else that pipes due to screen width being misinterpreted. (a dummy screen approach would work there, AFAIK, though maybe we can force urwid to believe it has a different approach).
* It's a lot of effort. Especially the less hacky approaches.

### Oportunities

* Consistency: We're sure that both look very much alike. We can tune one to have less months, etc, but the fine details will be identical on both.

### Risks

* If we take the hackier approaches (touching the event hooks, or alarms), we risk this breaking with newer `urwid` versions if something changes (because we're basically exploiting it). A dummy screen wouldn't have this risk.

---

I'd like to leave this open for a place of some brief discussion. I'd really like to work on this if the idea sound acceptable. If I'm not being very clear, just ask (it's 1am, so maybe I'm making less sense that I think).

geier:
I really like this idea for the **Pros** and don't think it would be that much work, my initial approach would be to call .render() on the outermost widget, but haven't really looked into it to see if that is a viable strategy.

added to **Cons**: potentially slower, I really don't want to see khal even slower than it currently is, so if it would be significantly slower, I wouldn't want to merge it, but even than the effort wouldn't be wasted in my opinion, as this would be a valuable tool for testing ikhal (and obviously other urwid applications).

Also, as I take my business strategy tools from "Silicon Valley", shouldn't this be Strengths, Weaknesses, Opportunities, and Threats?

hobarrera:
Good points on the pro and con.

[...]

PR #611, opened 2017-03-24
WIP: Use urwid to draw `khal calendar`

WhyNotHugo:
This is a work in progress/preview of what's been discussed on #609.

We should be able to use `CanvasTranslator` (gee, what an awful name!) to test some other urwid components, probably.


geier:
closing for lack of activity

PR #612, opened 2017-03-24
Remove outdated references to configspec

hobarrera:
Time to delete the auto-generated file, since there's no point in keeping it around (it's now just an intermediate build-time file).

Also remove outdated doc reference, since this hasn't been true for a while.


hobarrera:
This is ready for review.

PR #615, opened 2017-03-26
Streamline config wizard

takluyver:
I really like khal, but I thought the initial setup was a bit fiddly. I guess most experienced users aren't going through the config wizard very often, so I thought I'd have a go at simplifying it.

- In master, a typical setup took 8 prompts and 15 keypresses
- The first commit here removes some of the intermediate confirmation prompts (if you get it wrong, you can just start over), and changes the final confirmation prompt to default to yes.
  - 6 prompts, 10 keypresses
- The second commit (which might be more controversial) condenses picking the date format into a single set of options I guess are most likely: `%Y-%m-%d, %d/%m/%Y, %m/%d/%Y`
  - 5 prompts, 8 keypresses

I think the first step is a clear improvement. The second is a trade-off between simplicity and flexibility. I think it's a better user experience with one fewer choice to think about, and users who really want another format can still set that in the config file.

Here's what it looks like after both commits:

```
$ khal configure
What ordering of year, month, date do you want to use?
[0] year-month-day (today: 2017-03-26)
[1] day/month/year (today: 26/03/2017)
[2] month/day/year (today: 03/26/2017)
Please choose one of the above options: 1
Date format: %d/%m/%Y (today as an example: 26/03/2017)

What timeformat do you want to use?
[0] 24 hour clock (recommended)
[1] 12 hour clock
Please choose one of the above options [0]:
Time format: %H:%M (current time as an example: 13:13)

If you use vdirsyncer to sync with CalDAV servers, we can try to load its config file and add your calendars to khal's config.
Should we try to load vdirsyncer's config? [Y/n]: n

Do you want to create a local calendar? (You can always set it up to synchronize with a server in vdirsyncer later). [y/N]: y
Created new vdir at /home/takluyver/.local/share/khal/calendars/private111
Do you want to write the config to /home/takluyver/.config/khal/config? (Choosing `No` will abort) [Y/n]:
Successfully wrote configuration to /home/takluyver/.config/khal/config
```

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -30,3 +30,4 @@ Paweł Fertyk - pfertyk [at] openmailbox [dot] org
 Moritz Kobel - moritz [at] kobelnet [dot] ch - http://www.kobelnet.ch
 Guilhem Saurel - guilhem [at] saurel [dot] me - https://saurel.me
 Stefan Siegel - ssiegel [at] sdas [dot] net
+Thomas Kluyver - thomas [at] kluyver [dot] me [dot] uk
--- a/khal/configwizard.py
+++ b/khal/configwizard.py
@@ -47,48 +47,42 @@ def validate_int(input, min_value, max_value):
 
 def choose_datetime_format():
     """query user for their date format of choice"""
-    while True:
-        ordering_choices = [
-            ('year month day', ['%Y', '%m', '%d']),
-            ('day month year', ['%d', '%m', '%Y']),
-            ('month day year', ['%m', '%d', '%Y']),
-        ]
-        separator_choices = ['-', '.', '/']
-        validate = partial(validate_int, min_value=0, max_value=2)
+    ordering_choices = [
+        ('year month day', ['%Y', '%m', '%d']),
+        ('day month year', ['%d', '%m', '%Y']),
+        ('month day year', ['%m', '%d', '%Y']),
+    ]
+    separator_choices = ['-', '.', '/']
+    validate = partial(validate_int, min_value=0, max_value=2)
 
-        print("What ordering of year, month, date do you want to use? "
-              "(You can choose the separator in the next step)")
-        print('\n'.join(
-            ['[{}] {}'.format(num, one) for num, (one, _) in enumerate(ordering_choices)]))
-        ordering_no = prompt("Please choose one of the above options", value_proc=validate)
-        print()
-        print("Now, please choose a separator")
-        print('\n'.join(['[{}] {}'.format(num, one) for num, one in enumerate(separator_choices)]))
-        prompt_text = "Please choose one of the above options"
-        separator = separator_choices[prompt(prompt_text, value_proc=validate)]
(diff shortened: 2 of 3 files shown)


untitaker:
It would be nice if I could enter a date format OR an integer selecting a format from the examples. The difference can be told by checking whether `int(my_input)` crashes.

takluyver:
I thought about that, but I don't want to clutter up the selection with information about the date format. It could be a hidden feature? Or there could be an option `[4] Custom format`, so it only shows the info if the user selects that.

untitaker:
I didn't think about showing info for the date format, since I thought it'd be intuitive enough with the examples being shown. But yeah, a fourth option might be better.

takluyver:
How about this:

```
$ khal configure
What ordering of year, month, date do you want to use?
[0] year-month-day (today: 2017-03-26)
[1] day/month/year (today: 26/03/2017)
[2] month/day/year (today: 03/26/2017)
[3] Custom
Please choose one of the above options: 3
Common fields for date formatting:
Year     Month     Day        
%Y=2017  %m=03     %d=26      
%y=17    %B=March  %a=Sun     
         %b=Mar    %A=Sunday  
More info: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior
Make your date format: %a %d %B
Date format: %a %d %B (today as an example: Sun 26 March)
```

untitaker:
Yeah looks good!

On 26 March 2017 15:49:19 CEST, Thomas Kluyver <notifications@github.com> wrote:
>How about this:
>
>```
>$ khal configure
>What ordering of year, month, date do you want to use?
>[0] year-month-day (today: 2017-03-26)
>[1] day/month/year (today: 26/03/2017)
>[2] month/day/year (today: 03/26/2017)
>[3] Custom
>Please choose one of the above options: 3
>Common fields for date formatting:
>Year     Month     Day        
>%Y=2017  %m=03     %d=26      
>%y=17    %B=March  %a=Sun     
>         %b=Mar    %A=Sunday  
>More info:
>https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior
>Make your date format: %a %d %B
>Date format: %a %d %B (today as an example: Sun 26 March)
>```
>
>-- 
>https://github.com/pimutils/khal/pull/615#issuecomment-289285792

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

geier:
I like this, the configuration wizard does indeed not see too much love (for reasons you outlined above). Could you please add yourself to AUTHORS.txt (and fix the flake8 errors)?

takluyver:
Done.

I'm also interested in improving how the wizard deals with vdirsyncer - I think it would be valuable to be able to set up khal with a new calendar and vdirsyncer at the same time. That's going to take a bit more thought, though, so I wanted to get this relatively simple change in first.

geier:
Thanks for this and I'd love to see some better integration with vdirsyncer in the wizard.

PR #621, opened 2017-04-04
Fix/pytz201702

geier:


.travis.yml
```diff
@@ -9,8 +9,8 @@ python:
     - 3.7-dev
 
 env:
-    - BUILD=py
-    - BUILD=negtz
+    - BUILD=pytz201702
+    - BUILD=pytz201610
```
hobarrera: Personally, I'd just include a new entry below in matrix/include.
```yaml
- python: 3.6
  env: BUILD=pytz201610
```
We don't need to cross every pytz version with EVERY python version, just running it with py36 would do (I'd prefer to keep the job list as small as possible).

TBH, I don't know why I didn't do the same with `negtz`.

geier: sounds reasonably, I'll give it a try.

PR #629, opened 2017-04-13
Integrate sync into config wizard

takluyver:
This follows on from #615, to make first-time setup more user friendly. It isn't 100% where I want it to be, but it shows how I think this should work. Instead of y/n questions, we present the user with another menu of options:

```
[0] Create a new calendar on this computer
[1] Use a calendar already on this computer (vdir format)
[2] Sync a calendar from the internet (CalDAV format)
```

Selecting 0 creates a new calendar, ignoring vdirsycner. Selecting 1 attempts to find vdirsyncer config to identify where a calendar might be; the user can reject this and enter a location manually if they prefer. Selecting 2 creates a new calendar and tries to write vdirsyncer config to sync it, prompting the user to enter details of a caldav server.

Things that aren't right yet:

a. I've currently assumed that khal's and vdirsyncer's config point to the same calendar directory, but looking at the config on my machine, I think vds actually points one level above, at a directory containing a collection of calendars. Is this correct?
b. When setting up vdirsyncer (for option 2), I'd like to set it up to run regularly, e.g. adding it to cron on Linux. Do you know of any good tools to do this kind of thing cross-platform?

### DIFF
--- a/khal/configwizard.py
+++ b/khal/configwizard.py
@@ -23,9 +23,11 @@
 import datetime as dt
 import logging
 from functools import partial
+import json
 from itertools import zip_longest
-from os import makedirs
-from os.path import exists, expanduser, expandvars, isdir, join, normpath
+from os import makedirs, environ
+from os.path import expanduser, expandvars, join, normpath, exists, isdir
+from subprocess import call
 
 import xdg
 from click import Choice, UsageError, confirm, prompt
@@ -125,21 +127,16 @@ def choose_default_calendar(vdirs):
 
 def get_vdirs_from_vdirsyncer_config():
     """trying to load vdirsyncer's config and read all vdirs from it"""
-    print("If you use vdirsyncer to sync with CalDAV servers, we can try to "
-          "load its config file and add your calendars to khal's config.")
-    if not confirm("Should we try to load vdirsyncer's config?", default=True):
-        return None
     try:
         from vdirsyncer.cli import config
         from vdirsyncer.exceptions import UserError
     except ImportError:
-        print("Sorry, cannot import vdirsyncer. Please make sure you have it "
-              "installed.")
+        print("Couldn't load vdirsyncer to discover its calendars.")
         return None
     try:
         vdir_config = config.load_config()
     except UserError as error:
-        print("Sorry, trying to load vdirsyncer failed with the following "
+        print("Sorry, trying to load vdirsyncer config failed with the following "
               "error message:")
         print(error)
         return None
@@ -153,21 +150,31 @@ def get_vdirs_from_vdirsyncer_config():
             path += '*'
             vdirs.append((storage['instance_name'], path, 'discover'))
     if vdirs == []:
-        print("No usable collections were found")
+        print("No calendards found from vdirsyncer.")
         return None
     else:
-        print("The following collections were found:")
-        for name, path, _ in vdirs:
-            print(f'  {name}: {path}')
         return vdirs
 
 
+def find_vdir():
+    print("The following collections were found:")
+    synced_vdirs = get_vdirs_from_vdirsyncer_config()
+    if synced_vdirs:
+        print("Found {} calendars from vdirsyncer")
+        for name, path, _ in synced_vdirs:
+            print('  {}: {}'.format(name, path))
+        if confirm("Use these calendars for khal?", default=True):
+            return synced_vdirs
+
+    vdir_path = prompt("Enter the path to a vdir calendar")
+    return [('private', vdir_path, 'calendar')]
+
 def create_vdir(names=None):
+    """create a new vdir, make sure the name doesn't collide with existing names
+
+    :param names: names of existing vdirs
+    """
     names = names or []
-    if not confirm("Do you want to create a local calendar? (You can always "
-                   "set it up to synchronize with a server in vdirsyncer "
-                   "later)."):
-        return None
     name = 'private'
     while True:
         path = join(xdg.BaseDirectory.xdg_data_home, 'khal', 'calendars', name)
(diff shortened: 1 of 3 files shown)


geier:
The UX looks good (I haven't looked at the actual code).

a) vdirsyncer's configuration will probably point to a directory of actual vdirs, but khal can deal with that if `type = discover` 

b) no, I don't even know if cron works reliably on OS X anymore, or if modern Linux/systemd uses it anymore.

Feel free to ask more questions (or join us on irc #pimutils@Freenode).

takluyver:
Thanks :-)

Systemd has 'timers', which are a more powerful way to do cron-like things, but it also still runs traditional cronjobs - I have vdirsyncer set up in cron. So far, I've not come across any suggestion that distros are going to drop cron, but it might happen one day.

I think for now I'll get it to print out a message about setting up regular runs of vdirsyncer. Then we can add code later to try to automate it, and fall back to the message for cases we can't handle.

takluyver:
Tests are passing, and I think I'm ready for this to be reviewed. :-)

geier:
@untitaker could you have a look at the vdirsyncer side of things here?

khal/configwizard.py
```diff
@@ -166,6 +187,131 @@ def create_vdir(names=[]):
...
+    print("Please set up your system to run 'vdirsyncer sync' periodically, "
+          "using cron or similar mechanisms.")
+
+
+def choose_vdir_calendar():
+    """query the user for their preferred calendar source"""
+    choices = [
+        ("Create a new calendar on this computer", create_vdir),
+        ("Use a calendar already on this computer (vdir format)", find_vdir),
+        ("Sync a calendar from the internet (CalDAV format)", create_synced_vdir),
```
geier: perhaps add something like (vdirsyncer required)?

takluyver: Done

geier:
I really dropped the ball on this PR. Sorry @takluyver!

If I remember correctly, back then untitaker wasn't sure if this was the right way forward because he wanted the freedom to keep changing vdirsyncer's configuration format if necessary. 

As vdirsyncer's development has largely stabilized, I don't really think this is a problem.

@WhyNotHugo What do you think? Can we merge this?

WhyNotHugo:
> a. I've currently assumed that khal's and vdirsyncer's config point to the same calendar directory, but looking at the config on my machine, I think vds actually points one level above, at a directory containing a collection of calendars. Is this correct?

Indeed, `vdirsyncer` will point to a directory containing multiple directories/collections. Khal points to collections directly. For example, my vdirsyncer config has:

    path = "~/.local/share/calendars/"

While khal has:

    path = ~/.local/share/calendars/*

They're very close tho, and appending `/*` should work on [almost?] every case.

> a) vdirsyncer's configuration will probably point to a directory of actual vdirs, but khal can deal with that if type = discover

I was entirely unaware of this. `vdirsyncer showconfig` is also handy if you need to programatically access vdirsyncer's config, though not in the latest stable, IIRC.

---

Either approach has no objections from my end. I haven't looked at the code here too much tho.

I would recommend adding a prominent note on this new feature on the README and CHANGELOG. It would be very useful, and I'd make sure it's not a hard-to-discover feature.

geier:
I rebased the PR onto master and force pushed into this branch. Unfortunately, the vdirsyncer setup fails for me atm.

khal/configwizard.py
```diff
@@ -113,7 +125,7 @@ def choose_time_format():
 def choose_default_calendar(vdirs):
     names = [name for name, _, _ in sorted(vdirs or ())]
     print("Which calendar do you want as a default calendar?")
-    print("(The default calendar is specified, when no calendar is specified.)")
+    print("(The default calendar is used as default, when no calendar is specified.)")
```
WhyNotHugo, line 128: `The default calendar is used as default` sounds a bit recursive. Like the dictionary definition:

Recursive: see recursive.

khal/configwizard.py
```diff
@@ -153,21 +160,38 @@ def get_vdirs_from_vdirsyncer_config():
...
+    """Use one or more existing vdirs on the system.
+
+    Tries to get data from vdirsyncer if it's installed and configured, and
+    asks user to confirm it. If not, prompt the user for the path to a single
+    vdir.
+    """
+    print("The following collections were found:")
+    synced_vdirs = get_vdirs_from_vdirsyncer_config()
+    if synced_vdirs:
+        print(f"Found {len(synced_vdirs)} calendars from vdirsyncer")
```
WhyNotHugo, line 179: Only print `"The following collections were found:"` is calendars are actually found:

```suggestion
    synced_vdirs = get_vdirs_from_vdirsyncer_config()
    if synced_vdirs:
        print("The following collections were found:")
        print(f"Found {len(synced_vdirs)} calendars from vdirsyncer")
```

There's mixed usage of "calendar" and "collection". Please stick to just one term and use that consistently.

khal/configwizard.py
```diff
@@ -185,6 +209,132 @@ def create_vdir(names=None):
...
+    There may or may not already be a file at the returned path.
+    """
+    fname = environ.get('VDIRSYNCER_CONFIG', None)
+    if fname is None:
+        fname = normpath(expanduser('~/.vdirsyncer/config'))
+        if not exists(fname):
+            xdg_config_dir = environ.get('XDG_CONFIG_HOME',
+                                         normpath(expanduser('~/.config/')))
+            fname = join(xdg_config_dir, 'vdirsyncer/config')
+    return fname
```
WhyNotHugo, line 251: I wonder if it wouldn't be best to expose this functionality on vdirsyncer, rather than duplicate this here.

A bit out of-scope tho, so this is not a blocking comment.

geier: probably, if we are at it, it would also be really nice if it could print all synchronized collections (and their type).

khal/configwizard.py
geier: my fault.

khal/configwizard.py
geier: also fixed

WhyNotHugo (commented):
Code looks good. Just output suggestions here.

khal/configwizard.py
```diff
@@ -185,6 +209,132 @@ def create_vdir(names=None):
...
+    caldav_url = prompt('CalDAV URL')
+    username = prompt('Username')
+    password = prompt('Password', hide_input=True)
+
+    vds_config = vdirsyncer_config_path()
+    if exists(vds_config):
+        # We are adding a pair to vdirsyncer config
+        mode = 'a'
+        new_file = False
+        pairno = get_available_pairno()
```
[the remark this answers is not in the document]
geier: let's leave it as it is for the moment, can alwlays clean that up.

khal/configwizard.py
```diff
@@ -113,7 +125,7 @@ def choose_time_format():
 def choose_default_calendar(vdirs):
     names = [name for name, _, _ in sorted(vdirs or ())]
     print("Which calendar do you want as a default calendar?")
-    print("(The default calendar is specified, when no calendar is specified.)")
+    print("(The default calendar is used, when no calendar is specified.)")
```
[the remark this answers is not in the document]
geier: fixed

takluyver:
Thanks for picking this up again after so long! :slightly_smiling_face:

geier:
Thanks for sending it int the first place and sorry again for ignoring it for several years.

takluyver:
No problem - I'm a maintainer on various open source projects myself, so I know how things can drop through the cracks.

PR #636, opened 2017-04-29
update_birthday: Augment sqlite3.IntegrityError with additional info

madduck:
khal wasn't actually expecting a sqlite3.IntegrityError, so this
converts it to an UpdateError, while also providing some more
information about the addressbook record for which we're creating
a birthday event.

Signed-off-by: martin f. krafft <madduck@madduck.net>

### DIFF
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -269,6 +269,7 @@ class SQLiteDb:
         for key in vcard.keys():
             if key in ['BDAY', 'X-ANNIVERSARY', 'ANNIVERSARY'] or key.endswith('X-ABDATE'):
                 date = vcard[key]
+                uuid = vcard.get('UID')
                 if isinstance(date, list):
                     logger.warning(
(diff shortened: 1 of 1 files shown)


madduck:
Hm, `BUILD=style` failed, and I think this must be due to the very long string. I am not sure how else you'd prefer it, and I actually consciously put it all into one line to not weigh too heavy on the code otherwise.

Please advise…

madduck:
I'd be happy to try to provide a test-case too, but first I need to understand the issue.

Is it that the assumption is that a birthday doesn't change, and so `INSERT` is fine? What happens in the case of typos?

madduck:
There, fixed the long line after finding out that Python indeed supports `("foo"\n"bar")` concatenation.

geier (approved):

geier:
only 4 years later: let's get this merged, better error messages are always good. @madduck do you want to rebase this or should I?

Issue #638, opened 2017-05-01
[Question] Performance of `khal list`


samdmarshall:
Hi, so i've been trying to integrate using `khal` into my terminal prompt to give myself a heads-up about upcoming events and such. I've already implemented a similar behavior with `notmuch` for when i have new email waiting for me. Major difference is in execution time, which, is not unexpected as `notmuch` is written in C, and `khal` is written in Python. Given that, I was wondering if there was anything that could be done around there performance around getting the number of upcoming events. Right now I'm measuring it at about 1/3 to 2/5 of a second to return the data, if I can get that fast it would be much appreciated.

Thanks.

hobarrera:
Actually, the issue it not because of python itself: it's the entry point that setuptools generates.

Maybe we could adapt https://github.com/pimutils/todoman/pull/88/files to khal, which trimmed off quite a bit for todoman.

samdmarshall:
My only hesitation behind the patch you linked to is that Python 2 uses `python` by default, whereas `python3` does not, and would not be feasible to patch these things on many systems, otherwise this sounds great.

hobarrera:
The script is mainly targetted towards packagers, so they can patch it accordingly (including weird python paths, etc).

Another optimization I've done locally (I think this is worth documenting), is running `khal > /dev/null` right after `vdirsyncer sync`. This makes sure the cache is up to date after syncs, and reduces startup time.

samdmarshall:
My khal config is set to use `interactive` by default, is there a particular command that you are running to update the cache?

hobarrera:
`khal list` should update the cache.

samdmarshall:
Hmmm, I dunno if i'm missing something about this then. I have a background task that runs every 15 minutes that invokes `vdirsyncer sync` so that my calendars and todo lists get updated automatically. When I run `khal` after an update i'm not noticing any measurable difference between invocations (before and after the cache is updated). Both `khal` and `todo` take approximately a third of a second to run regardless, so i'm not sure if there is a measurable perf change in the patch you linked to.

hobarrera:
That might depend a bit on your scenario. One of my devices/servers (not sure which) tend to "touch" lots of events, so during every `sync`, about half of them are needlessly updated, making the cache invalid.

geier:
Can you [perhaps generate a profile](https://github.com/pimutils/khal/issues/172#issuecomment-78962657) so we can actually see what actually uses up the CPU time?

[...]

PR #639, opened 2017-05-01
Bundle alternative entry point files

hobarrera:
The ones generated by setuptools are quite slow (+1s on a modern system), so include alternative entry points file for downstream packagers to use.

Based on pimutils/todoman#88. There's also some mention of discarded alternatives there, if there's interest in them.

Related to #638.


hobarrera:
This is pretty much ready for review. It's mostly a hack until setuptools improves this, but for packagers, it lets them create packages that avoid that ~1s overhead.

geier:
thanks.

PR #646, opened 2017-05-04
Set prodid to khal/icalendar

geier:


khal/khalendar/event.py
```diff
@@ -311,7 +311,9 @@ def _create_calendar():
         """
         calendar = icalendar.Calendar()
         calendar.add('version', '2.0')
-        calendar.add('prodid', '-//CALENDARSERVER.ORG//NONSGML Version 1//EN')
+        calendar.add(
+            'prodid', '-//PIMUTILS.ORG//NONSGML khal / icalendar //EN'
```
hobarrera: Out of curiosity, what does `NONSGML` mean here?

geier: Non Standard Generalized Markup Language, we could really put anything in there, but somehow everybody seems to use this format.

PR #652, opened 2017-05-15
Import from stdid if no filename is provided

hobarrera:
Import from stdin if not filename is provided. This is especially useful for integrating khal with mutt.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,8 @@ unreleased
 * FIX don't crash on updated vcards
 * FIX checking for RRULEs we understand
 
+* CHANGE import will read from stdin if not filenames are provided.
+
 * NEW new entry points recommended for packagers to use.
 * NEW support keyword `yesterday` for querying and creating events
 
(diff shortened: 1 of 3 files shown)


geier:
doesn't work for me
```
cat test.ics | khal import                                            ~/workspace/khal
27.01. 08:00-27.01. 10:00 ⟳ :: Foobar
Which calendar do you want to import to? (unique prefixes are fine)
events(0), freundin(1), home(2), meetup(3), work(4) [home]: Aborted!
```
I guess because click than tries to read from stdin.

hobarrera:
Oh, right, this worked for me: `cat 5SYRVOCXNEN98720VJF8JEQYE7N0QANS4VFX.ics | khal import -a work --batch`.

I guess we need to run in `batch` mode when reading from `stdin`.

hobarrera:
Oh, I've a default calendar set up, you don't. I guess that needs to work too.

I'll do some testing with mutt itself to see how much interactivity can work -- if possible, I'll try to keep that working.

geier:
if we are reading from stdin, we will only ever read one .ics "file", so we should be able to rewire stdin to /dev/tty.

geier:
I just pushed an example of what I mean to your branch, feel free to squash it.

geier:
superseded by #658

PR #654, opened 2017-05-15
increment width of the month column. fix #653

### LINKED ISSUE #653: Broken design for months that have long abreviated names
With a french locale, this is what ikhal looks like (note the blank lines for juil. and others) :  
khal

nim65s:


geier:
I'd prefer to retrieve the appropriate length, who knows if 5 is enough for all locales?

geier:
Does anyone know if there is a better way than `max([len(calendar.month_abbr[i]) for i in range(1, 13)])`?

geier:
Also, why is that *really* slow?

nim65s:
hum, I did not found it really slow

geier:
I guess it depends on what you call slow:

```python
In [1]: import calendar

In [2]: %timeit max([len(calendar.month_abbr[i]) for i in range(1, 13)])
10000 loops, best of 3: 45.5 µs per loop

In [3]: tt = [''] + [calendar.month_abbr[i] for i in range(1, 13)]

In [4]: %timeit max([len(tt[i]) for i in range(1, 13)])
1000000 loops, best of 3: 1.88 µs per loop
```

I'm mostly wondering where this factor of 20 versus a list lookup comes from.

nim65s:
```python
In [5]: %timeit [calendar.month_abbr[i] for i in range(1, 13)]
35.2 µs ± 178 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

nim65s:
Yes, `calendar.month_abbr[1]` is almost 3μs, so times 12…

geier:
Thanks! (I bug reports with accompanied fixes are really my favorite kind of bug reports)

PR #665, opened 2017-06-07
Fix lack of space in fatal error ("tofile an issue")

strk:
Also reduces duplication by storing the error message in a variable

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -638,15 +638,10 @@ class LocalizedEvent(DatetimeEvent):
         try:
             starttz = getattr(self._vevents[self.ref]['DTSTART'].dt, 'tzinfo', None)
         except KeyError:
-            logger.fatal(
-                "Cannot understand event {} from calendar {},\n you might want to"
-                "file an issue at https://github.com/pimutils/khal/issues"
-                "".format(kwargs.get('href'), kwargs.get('calendar'))
-            )
+            msg = "Cannot understand event {} from calendar {}, you might want to file an issue at https://github.com/pimutils/khal/issues".format(kwargs.get('href'), kwargs.get('calendar'))
+            logger.fatal(msg)
             raise FatalError(  # because in ikhal you won't see the logger's output
-                "Cannot understand event {} from calendar {},\n you might want to"
-                "file an issue at https://github.com/pimutils/khal/issues"
-                "".format(kwargs.get('href'), kwargs.get('calendar'))
+                msg
             )
 
         if starttz is None:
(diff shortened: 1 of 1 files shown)


untitaker:
FWIW you can preserve the line wrapping, since all that is missing is a space at the end of each line:

    "lorem ipsum"
    "dolor"

becomes:

    "lorem ipsum "
    "dolor"

khal/khalendar/event.py
```diff
@@ -638,15 +638,13 @@ def __init__(self, *args, **kwargs):
         try:
             starttz = getattr(self._vevents[self.ref]['DTSTART'].dt, 'tzinfo', None)
         except KeyError:
-            logger.fatal(
-                "Cannot understand event {} from calendar {},\n you might want to"
-                "file an issue at https://github.com/pimutils/khal/issues"
-                "".format(kwargs.get('href'), kwargs.get('calendar'))
-            )
+            msg = "Cannot understand event {} from " \
+                  "calendar {}, you might want to file an issue at " \
```
hobarrera: Intentation should remain the same (also, as a convention, indentation should always be four spaces).

khal/khalendar/event.py
```diff
@@ -638,15 +638,13 @@ def __init__(self, *args, **kwargs):
         try:
             starttz = getattr(self._vevents[self.ref]['DTSTART'].dt, 'tzinfo', None)
         except KeyError:
-            logger.fatal(
-                "Cannot understand event {} from calendar {},\n you might want to"
-                "file an issue at https://github.com/pimutils/khal/issues"
-                "".format(kwargs.get('href'), kwargs.get('calendar'))
-            )
+            msg = "Cannot understand event {} from " \
```
hobarrera: Prefer parenthesis to slashes to join lines like these.

khal/khalendar/event.py
```diff
@@ -638,15 +638,13 @@ def __init__(self, *args, **kwargs):
...
+                  "calendar {}, you might want to file an issue at " \
+                  "https://github.com/pimutils/khal/issues" \
+                  .format(kwargs.get('href'), kwargs.get('calendar'))
+            logger.fatal(msg)
             raise FatalError(  # because in ikhal you won't see the logger's output
-                "Cannot understand event {} from calendar {},\n you might want to"
```
hobarrera: Was there any downside to simply adding the space here?

strk:
Downside: duplicated string, would have had to fix in two places.

See how you like 4f1063c136a99a28210cf736bf7c259c793cf544

geier:
Thank you!

hobarrera:
Duh, sorry, the diff confused my and I failed to noticed the duplication. Cool, thanks!

PR #673, opened 2017-06-19
Fix arg1 of dt.datetime.combine ()

pdav:
call to: ``dt.datetime.combine (dt.date, dt.time(dt.datetime.now().hour))``
fails with:
``TypeError: combine() argument 1 must be datetime.date, not type``

Test with: start ikhal, then press the ``n`` key to add a new event.

### DIFF
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -73,7 +73,6 @@ from .calendarwidget import CalendarWidget
 ALL = 1
 INSTANCES = 2
 
-
 class DateConversionError(Exception):
     pass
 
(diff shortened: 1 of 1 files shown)


geier:
Ouch, I messed that one up during the datetime import refactor last week.

khal/ui/__init__.py
```diff
@@ -870,7 +869,7 @@ def new(self, date, end=None):
             self.pane.window.alert(('alert', 'No writable calendar.'))
             return
         if end is None:
-            start = dt.datetime.combine(dt.date, dt.time(dt.datetime.now().hour))
+            start = dt.datetime.combine(dt.date.today (), dt.time(dt.datetime.now().hour))
```
geier: This should actually be `dt.datetime.combine(date, dt.time(dt.datetime.now().hour())`, date is a variable given to `new()` here.

pdav:
Whops! New commit added.

geier:
thank you!

Issue #685, opened 2017-07-18
usability improvement: put month name on the first row where the first day is in the month


pdav:
In the calendar view (left panel), the month name is on located on the first row where a day of the month appears. Example for Jun 2017:

         Mo Tu We Th Fr Sa Su
          ....
    Jun  29 30 31  1  2  3  4
          5  6  7  8  9 10 11
          ....

Instead, it should be located on the first row where the first day (either monday or sunday or another day, depending upon the `firstweekday` configuration item) is inside the new month. Example:

         Mo Tu We Th Fr Sa Su
          ....
         29 30 31  1  2  3  4
    Jun   5  6  7  8  9 10 11
          ....

This would be a great readability improvement.

geier:
Personally I prefer the way it is now, but if many people agree with your suggestion I'm open to have it changed.

pdav:
May I suggest to make a configuration option?

 Example:
    month_display = first_day | plain_week

geier:
While I generally dislike the amount of configurability we already have,
as this one doesn't affect function and is easy to test, it probably is
a good option to make it configurable.


Quoting Pierre DAVID (2017-07-19 16:26:37)

geier:
this was closed by #686

PR #686, opened 2017-07-19
Add [view]monthdisplay = firstday | plainweek

pdav:
Issue #685


geier:
Looks good. I'm not sure if `plainweek` is self-documenting so. I'd prefer something like `firstfullweek`, but I guess that is too long. Perhaps somebody else has some ideas?

pdav:
I'm ok with `firstfullweek`.

geier:
Thanks!

PR #696, opened 2017-08-10
Emphasise and align editable fields

### LINKED ISSUE #695: Remove default 'new event' title for new events (ikhal) or make it configurable
I'm using khal 0.9.6. When I open ikhal and type `n` (new event) I get the default event title `new event`. I almost never want to create an event with such a title, so every time I have to remove it. I'm pretty sure that previous versions of khal used an empty title by default.

Can this default `new event` value be removed? If not, could we make it configurable (`default_new_event_title: ''`)?

pfertyk:
Fixes #695.


geier:
Is this done and/or do you need any help with it?

pfertyk:
Just added a comment to the original issue.

geier:
Looks much better than before, thanks!

PR #703, opened 2017-08-23
Set new date in the future if year is inferred

### LINKED ISSUE #697: default year should be such that event is in the future
So far the default year is always the current year. Wouldn't it be more handy if khal understands by default that I mean an event in the future?   If, in December 2017, I add a new event for the 20.04. , it should understand that I mean 20.04.2018, not 2017.

pfertyk:
Fixes #697.

Note: I still need to add tests for this.

### DIFF
--- a/tests/utils_test.py
+++ b/tests/utils_test.py
@@ -395,6 +395,12 @@ test_set_format_de = _create_testcases(
         'DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20131025T180000',
         'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20141026T200000')),
 
+    # only start date given (no year, past day and month)
+    ('25.01. 18:00 20:00 Äwesöme Event',
+     _create_vevent(
+        'DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20150125T180000',
+        'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20150125T200000')),
+
     # date ends next day, but end date not given
     ('25.10.2013 23:00 0:30 Äwesöme Event',
      _create_vevent(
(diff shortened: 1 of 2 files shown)


pfertyk:
Existing tests should be fixed, new test was added.

tests/utils_test.py
```diff
@@ -390,11 +390,16 @@ def test_construct_daynames():
         'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20131026T000000')),
 
     # start and end date same, explicit end date (but no year) given
-    # XXX FIXME: if no explicit year is given for the end, this_year is used
     ('25.10.2013 18:00 26.10. 20:00 Äwesöme Event',
      _create_vevent(
         'DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20131025T180000',
-        'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20131026T200000')),
+        'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20141026T200000')),
```
geier: I don't think this should generate an event lasting for more than a year (and the FIXME is outdated any way).

geier:
It probably should work like this:
* if only a time is given, the next time that time occurs is meant
* if a date without a year is given, the next time that date is occurs is meant *unless* it follows after a date including a year, than the next time the second date occurs after the first date is meant

we probably need some more tests for guessdatetimefstsr as well.

geier:
I have merged these changes into #715 (and hopefully soon into master). Thanks again!

pfertyk:
I'm sorry that I didn't finish the task, I was very busy recently :disappointed:. I'd like to implement the functionality you described, but I can't say exactly when it will be possible.

geier:
No need, perhaps some more cleaning up to do, but I think we are good.
I applied a minor patch to your PR and added some tests and merged it
[0].  Thanks for taking care of this!


[0] https://github.com/pimutils/khal/pull/715/commits

Quoting pfertyk (2017-10-02 15:43:03)

PR #743, opened 2017-12-03
Add khald executable for complete overview

### LINKED ISSUE #725: Day-by-day summary view
I coded myself a report showing a compact day-by-day view of my appointments, which I find useful for doing short to medium term planning. The code is a quick and dirty ripoff of the existing agenda report, and the holidays are hardcoded as the Italian ones. I find it utterly useful, and so here I share it. Feel free to add it to khal if you like it, consider it released under the public domain:

```py3
def do_summary(collection, dates=None, encoding='utf-8', show_all_days=False, full=False,
           week=False, bold_for_light_color=True, **kwargs):
    from click import echo
    term_width, _ = get_terminal_size()
    event_column = get_summary(collection, dates=dates, width=term_width,
                              show_all_days=show_all_days, full=full, week=week,

                              bold_for_light_color=bold_for_light_color, **kwargs)
    # XXX: Generate this as a unicode in the first place, rather than
    # casting it.
    echo('\n'.join(event_column))

def is_holiday(day):
    import dateutil.easter
    from datetime import date, timedelta
    if day.weekday() == 6:
        return True
    if (day.month, day.day) in (
            (1, 1), (1, 6), (4, 25),
            (5, 1), (6, 2), (8, 15),
            (11, 1), (12, 8), (12, 25),
            (23, 26)):
        return True

    easter = dateutil.easter.easter(day.year)
    if day == easter + timedelta(days=1):
        return True
    return False

def get_summary(collection, locale, dates=None, firstweekday=0, days=None, events=None, width=45,
               week=False, full=False, show_all_days=False, bold_for_light_color=True,):
    """returns a list of events scheduled for all days in daylist

    included are header "rows"
    :param collection:
    :type collection: khalendar.CalendarCollection
    :param dates: a list of all dates for which the events should be return,
                    including what should be printed as a header
    :type collection: list(str)
    :param show_all_days: True if all days must be shown, event without event
    :type show_all_days: Boolean
    :returns: a list to be printed as the agenda for the given days
    :rtype: list(str)

    """
    from datetime import date, timedelta
    from click import style, echo

    event_column = list()

    if days is None:
        days = 2

    if dates is None or len(dates) == 0:
        dates = [date.today()]
    else:
        try:
            dates = [
                aux.guessdatetimefstr([day], locale)[0].date()
                if not isinstance(day, date) else day
                for day in dates
            ]
        except InvalidDate as error:
            logging.fatal(error)
            sys.exit(1)

    if week:
        dates = [d - timedelta((d.weekday() - firstweekday) % 7)
                 for d in dates]
        days = 7

    if days is not None:
        daylist = [day + timedelta(days=one)
                   for one in range(days) for day in dates]
        daylist.sort()

    last_day = None
    for day in daylist:
        if last_day is None or day.month != last_day.month:
            event_column.append(style(day.strftime("%B"), bold=True))

        last_day = day
        events = sorted(collection.get_events_on(day))
        if is_holiday(day):
            prefix = style(day.strftime("%d*%a: "), bold=True)
        else:
            prefix = style(day.strftime("%d %a: "), bold=False)
        if not events:
            event_column.append(prefix)
            continue
        for event in events:
            lines = list()
            items = event.relative_to(day, full).splitlines()
            for item in items:
                lines += textwrap.wrap(item, width)
            lines = [colored(line, event.color, bold_for_light_color=bold_for_light_color) for line in lines]
            event_column.append(prefix + lines[0])
            prefix = " " * 8
            event_column += [prefix + line for line in lines[1:]]

    if event_column == []:
        event_column = [style('No events', bold=True)]
    return event_column
```

ariep:
Intends to close #725 by adding a new command that gives a complete but compact overview of all events in a given period.

The code is due to @spanezz ; I have updated it here and there to work with khal's current master.

It is implemented at first as a separate executable, based on the main cli program. If acceptable, it should probably be reworked into the main cli program as a subcommand.

### DIFF
--- a/khal/daycal.py
+++ b/khal/daycal.py
@@ -1,0 +1,370 @@
+#!/home/arie/programs/khal/venv-git/bin/python3
+import logging
+import sys
+import textwrap
+from shutil import get_terminal_size
+
+try:
+    from setproctitle import setproctitle
+except ImportError:
+    def setproctitle(x):
+        pass
+
+import click
+import pytz
+
+from khal import controllers, khalendar, __version__
+from khal.parse_datetime import guessdatetimefstr
+from khal.settings import get_config, InvalidSettingsError
+from khal.settings.exceptions import NoConfigFile
+from khal.exceptions import FatalError
+from khal.terminal import colored
+
+
+logger = logging.getLogger('khal')
+
+days_option = click.option('--days', default=None, type=int,
+                           help='How many days to include.')
+week_option = click.option('--week', '-w',
+                           help=('Include all events in one week.'), is_flag=True)
+events_option = click.option('--events', default=None, type=int,
+                             help='How many events to include.')
+dates_arg = click.argument('dates', nargs=-1)
+
+
+def time_args(f):
+    return dates_arg(events_option(week_option(days_option(f))))
+
+
+def _multi_calendar_select_callback(ctx, option, calendars):
+    if not calendars:
+        return
+    if 'calendar_selection' in ctx.obj:
+        raise click.UsageError('Can\'t use both -a and -d.')
+    if not isinstance(calendars, tuple):
+        calendars = (calendars,)
+
+    mode = option.name
+    selection = ctx.obj['calendar_selection'] = set()
+
+    if mode == 'include_calendar':
+        for cal_name in calendars:
+            if cal_name not in ctx.obj['conf']['calendars']:
+                raise click.BadParameter(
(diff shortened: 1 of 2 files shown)


geier:
Hi, I'm sorry that nobody ever looked at this in over 4 years.

I will close this, because of its age (would need to be rebased) and also, because it would need a lot of work. The new functionality should be included into cli.py and controller.py instead of mostly copying functions from the existing code base.

PR #754, opened 2018-02-05
There's now an official Arch package

WhyNotHugo:
There's now an official Arch package in the [community] repositories.

Updated the docs to reflect this.


geier:
Nice, thanks!

Issue #763, opened 2018-03-13
all-day events with edit command


aleprovencio:
Hello,
when using the `edit` command, I'm unsure of the correct way of setting the datetime for all-day events.

When using the `new` command and as its argument just the date (without time and end date) or use a keyword (today, tomorrow, etc), the event will show correctly as an all-day event (i.e. not displaying times associated with it). This also happens when marking `all-day` with `ikhal`.

However, when I use `edit`, choose to set the `datetime` and input the all-day event as I used with `new`, the event will display its time at `00:00-24:00`.

geier:
This is probably a bug...

murilocamargos:
Hi @geier, I've just added a PR #769 with the fix for this. Can you take a look, please?

jlamoure:
khal 0.9.10, apologies if this has been fixed or else doing something wrong. Using Debian Buster's package.

I, too, cannot find a command arrangement that allows me to change the date of an all-day event without changing it into a time-ranged event. Likewise, I cannot change an event from time-ranged to all-day.

PR #769, opened 2018-04-07
Fix/763

murilocamargos:
This is a fix for #763. When adding an event, if `allday` flag is `True`, only the date is taken from datetime object; when updating an event, this flag wasn't being considered.

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -441,7 +441,7 @@ def edit_event(event, collection, locale, allow_quit=False, width=80):
             value = prompt("datetime range", default=current)
             try:
                 start, end, allday = parse_datetime.guessrangefstr(value, locale)
-                event.update_start_end(start, end)
+                event.update_start_end(start, end, allday)
                 edited = True
             except:  # noqa
                 echo("error parsing range")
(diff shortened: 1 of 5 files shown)


khal/khalendar/event.py
```diff
@@ -171,7 +171,7 @@ def __lt__(self, other):
         except TypeError:
             raise ValueError('Cannot compare events {} and {}'.format(start, other_start))
 
-    def update_start_end(self, start, end):
+    def update_start_end(self, start, end, allday=False):
```
geier: this doesn't look right, update_start_end should figure out by itself if start and end are of time date and datetime

khal/controllers.py
```diff
@@ -441,7 +441,7 @@ def edit_event(event, collection, locale, allow_quit=False, width=80):
             value = prompt("datetime range", default=current)
             try:
                 start, end, allday = parse_datetime.guessrangefstr(value, locale)
-                event.update_start_end(start, end)
+                event.update_start_end(start, end, allday)
```
geier: this conversion either needs to happen here, or more likely, there is a bug somewhere that makes start and end of time datetime even though it should be a date

geier:
Hi, are you still interested in this PR?

geier:
Feel free to send a new PR if you still want to fix those issues

Issue #775, opened 2018-05-04
khal config documentation should explain colors / metasync


amandabee:
I was really struggling to figure out how to colorize khal and circuitously stumbled on an issues (#542) which clued me into metasync. 

If this paragraph, from the [config documentation](https://khal.readthedocs.io/en/latest/configure.html): 
> If the color is set to auto (the default), khal tries to read the file color from this calendar’s vdir, if this fails the default_color (see below) is used. If color is set to ‘’, the default_color is always used. 

just said something like "if you want to use `type = discover` you can manually set colors for each calendar by creating a file at {path}/color with a color name in it, or you can run `vdirsyncer metasync` to synchronize colors with your caldav server" I would have found what I wanted a whole lot faster.

geier:
Thank you for pointing out where the documentation is currently failing and making a good suggestion. Do you want to send a PR yourself? If not, may I take your suggestion as is and add you to the AUTHORS.txt file? (In that case, what entry would you like?)

amandabee:
I can create a pull request with my suggestion but I feel like it probably needs an edit pass?

untitaker:
@amandabee what do you mean by edit pass?

amandabee:
Is that the best language to use? Is it the most accurate way to phrase it? 

On May 5, 2018 9:09:00 AM PDT, Markus Unterwaditzer <notifications@github.com> wrote:
>@amandabee what do you mean by edit pass?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/pimutils/khal/issues/775#issuecomment-386816083

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

untitaker:
Ah, sure, it'll be reviewed if you open a PR

amandabee:
Where does the text of the manual live? A search for matching phrases turns up only this page: 

https://github.com/pimutils/khal/blob/master/khal/settings/khal.spec
 
But that doesn't directly correspond to https://khal.readthedocs.io/en/latest/configure.html

geier:
I was sure I had answered your question, sorry @amandabee 

configure.html gets automatically generated from khal.spec, therefore editing khal.spec is the right way to go.

d7415:
Closed in #784.

PR #784, opened 2018-05-16
flag `vdirsyncer metasync` option on "color"

amandabee:
Tweaking the text to make it more clear where and how color can be set. Addresses #775


geier:
Thank you for the PR!

PR #792, opened 2018-06-03
Control Codes

### LINKED ISSUE #737: Newlines in --format argument impossible
Hi,

i am using khal 0.9.8. I try to style the printics command output with the format agument:

```
khal printics --format '{start}\n{title}' file.ics
1 events found in file.ics
2017-11-10 11:00\nMeeting
```

I can see that backslashes are escaped before the format string is used. How to put newlines in the format? Maybe a format attribute {nl} could be added?

Thanks,
Florian

### LINKED ISSUE #765: Using displayname doesn't ignore the newline character
I am using a calendar of type `discover` and it uses the `displayname` file to display the name of the calendars. However, if that file happens to end by a newline, a few things happens:

* It is no longer possible to set a default calendar (you get `critical: in section [default] DisplayName1 is not valid for 'default_calendar', must be one of ['DisplayName1\n', 'DisplayName2\n']`).
* The "pop-up" to change the calendar of an event in interactive mode has gaps.

I believe having a newline at the end of the `displayname` isn't a problem and thus should be ignored by khal. Some text editors don't even allow to remove it.

I am able to make a PR to fix this, I just need to know if it's okay to modifiy this behaviour.

daryltucker:
# Control Codes
## Overview

Adding control codes, `\n`, `\t`, and `\a`, which provide newline, tab and bell respectively.

Requests for newlines:

* Closes #737

Out of the available [control characters](https://en.wikipedia.org/wiki/Control_character), I felt that tab and bell would also be beneficial.  Please let me know if you disagree.

These control codes are now explained in `usage.rst`, similar to how colors are documented.

### Validation/Manual Testing

    $ cat ~/.config/khal/config
    ...
    agenda_event_format = {calendar-color}{cancelled}{start-time-full}{to-style}{end-time-full} {title}{repeat-symbol}{nl}{tab}{location}{reset}
    ...

    $ clear && khal [--color] list
    
    Tomorrow, 2018-06-04
    09:00-9:30 Repeating Meeting ⟳
            https://link-to-meeting/
    1:30-2:00 Single Meeting
            https://link-to-meeting/

Also tested with `{bell}`, which is working, but this example is closer to what I or someone else might actually want.

### `khal` self-test

    $ py.test tests/event_test.py 
    ==================================================================================== test session starts =====================================================================================
    platform linux -- Python 3.6.5, pytest-3.6.0, py-1.5.3, pluggy-0.6.0
    rootdir: /mnt/data/src/khal, inifile:
    collected 44 items                                                                                                                                                                           
    
    tests/event_test.py ............................................                                                                                                                       [100%]
    
    ================================================================================= 44 passed in 0.29 seconds ==================================================================================

When attempting to test khal as a whole, `dateutils` is showing errors.
There are known-issues similar to this, so I'm assuming this is expected right now.
If there is something I can do to get these tests working properly, or an alternative, please let me know.  Otherwise, I'll monitor Travis.

    $ py.test test
    ...
                if (self._dtstart.tzinfo is not None) != (self._until.tzinfo is not None):
                    # According to RFC5545 Section 3.3.10:
                    # https://tools.ietf.org/html/rfc5545#section-3.3.10
                    #
                    # > If the "DTSTART" property is specified as a date with UTC
                    # > time or a date with local time and time zone reference,
                    # > then the UNTIL rule part MUST be specified as a date with
                    # > UTC time.
                    raise ValueError(
    >                   'RRULE UNTIL values must be specified in UTC when DTSTART '
                        'is timezone-aware'
                    )
    E               ValueError: RRULE UNTIL values must be specified in UTC when DTSTART is timezone-aware
    
    /usr/lib/python3.6/site-packages/dateutil/rrule.py:464: ValueError
    ====================================================================== 14 failed, 265 passed, 5 xfailed in 3.85 seconds ======================================================================

## Notes

1. These control characters are best with the `--list` output variation.
    * Using `agenda_event_format` in `config` and `--list` argument accomplish this.
1. Fixed a typo in `usage.rst`.

## Unrelated Notes

1. In addition to `pytest`, `freezegun` is required for testing.
    * ref: https://khal.readthedocs.io/en/latest/hacking.html
1. Should `.pytest_cache/*` be added to `.gitignore`?

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -599,6 +599,10 @@ class Event(object):
             for c in ["black", "red", "green", "yellow", "blue", "magenta", "cyan", "white"]:
                 attributes[c] = attributes[c + '-bold'] = ''
 
+        attributes['nl'] = '\n'
+        attributes['tab'] = '\t'
+        attributes['bell'] = '\a'
+
         attributes['status'] = self.status
         attributes['cancelled'] = 'CANCELLED ' if self.status == 'CANCELLED' else ''
         return format_string.format(**dict(attributes)) + attributes["reset"]
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -161,7 +161,10 @@ Several options are common to almost all of :program:`khal`'s commands
    is also `reset`, which clears the styling, and `bold`, which is the normal
    bold.
 
-   For example the below command with print the title and description of all events today.
+   A few control codes are exposed.  You can access newline (`nl`), 'tab', and 'bell'.
+   Control codes, such as `nl`, are best used with `--list` mode.
+
+   Below is an example command which prints the title and description of all events today.
 
    ::
 


daryltucker:
Removed mistaken reference to unrelated ticket.

khal/khalendar/event.py
```diff
@@ -599,6 +599,11 @@ def format(self, format_string, relative_to, env={}, colors=True):
             for c in ["black", "red", "green", "yellow", "blue", "magenta", "cyan", "white"]:
                 attributes[c] = attributes[c + '-bold'] = ''
 
+        if "control codes":
```
geier: What is this if statement doing? "control codes" will always be True.

geier:
Thank you for your PR and sorry for the long delay in me looking after it.

Please have a look at the single comment I left, otherwise this looks good to me, but you might want to add a test (feel free to ask for help, if you don't know how). 

Your unrelated notes: your are right on both occasions, feel free to open a PR.

khal/khalendar/event.py
daryltucker: This was done to keep the code at the same tab level as the other `attributes` modifications, as well as to allow you to extend this to be toggle-able in the future (ie: allow some users to turn it off for whatever reason).

I'd be happy remove the logic if you'd prefer, no problem.

daryltucker:
Please let me know if I should look into these failed builds, or if this is something you're expecting.  I've re-performed the tests mentioned in the PR, and have the same working results.

geier:
looks good, thanks!

PR #797, opened 2018-07-17
Support Python 3.7

WhyNotHugo:
The 3.7 stable release is out, so let's run tests with it and mark it as supported (no actual changes are required to support it).


geier:
Thanks, looks good!

Issue #526, opened 2016-11-10
Have calendars with "low priority coloring"


HappyTetrahedron:
Hello

I've been using khal for a while now and I love it, but there's one little feature I'd love to have. 

I'd like to be able to specify a calendar to have "low priority coloring", which means that the events get a color in the agenda, but that color doesn't cause the "multiple" color to appear when an event from this calendar conflicts with events from the other. In other words, the colors of some calendars should have precedence over the color of other calendars.

I think this is probably better explained with an example: I have my lecture schedule in one calendar, and several other calendars with appointments for various groups. Naturally, these appointments are more important since my schedule hardly ever changes anyway. Now, even when I specify no color for the schedule calendar, the "multiple" color shows up on the calendar when an appointment is on a day where I also have a lecture. I'd like the color of the appointment to have precedence, so that I can immediately see from which calendar it is, without having to entirely ignore the schedule calendar using the -i option.

Maybe this could be generalized into priorities for the calendars. The color of a calendar with higher priority suppresses the color of one with lower priority, but when two calendars with equal priority have a conflict, the "multiple" color shows up.

geier:
This sounds like a very interesting enhancement! If anyone is interested in implementing this, I'll gladly help over any bumps, but I probably won't have time for this in the near future to do it myself.

PR #812, opened 2018-09-18
Add coloring priority system as per #526

RcColes:
Adds a priority property to each calendar, which controls how day coloring is done. The day is colored based on the color of the calendar with the highest priority, falling back to the current system in the case of multiple calendars having the same highest priority.

I've updated the tests, and as far as I can tell all the tests pertaining to the code are passing (which is somewhat tricky with the current ```dateutil``` problems). 

Currently there don't appear to be any tests related to coloring (as far as I can tell from codecov), so I thought I'd just check that there wasn't a reason for that before adding some.

As for the documentation, because it's a config change the autogenerated documentation for the [calendars] has been updated. I'm not sure if this should also be mentioned in other sections as well however. I'd also appreciate somebody looking over the wording of the documentation since I'm not entirely happy with how it describes the options.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -34,3 +34,4 @@ August Lindberg
 Thomas Kluyver - thomas [at] kluyver [dot] me [dot] uk
 Tobias Brummer - hallo [at] t0bybr.de - https://t0bybr.de
 Amanda Hickman - amanda [at] velociraptor [dot] info
+Raef Coles - raefcoles [at] gmail [dot] com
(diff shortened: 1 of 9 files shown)


geier:
Hi and thanks for this PR! Can you please add one or two tests for the output or at least `get_color_list`? Looks good to me otherwise!

RcColes:
Right, there's the tests added. Unfortunately it seems to now be failing the print_ics_command test, which is strange because I don't think I touched any code related to that (and also it started failing between cc09ce3 and bd0e9c0 which was a trivial change). I can pretty easily fix it so it passes the test, but I'd rather know why it's suddenly started failing. Any ideas?

geier:
You are right, those issues are probably not connected at all.

Thanks for your work!

PR #819, opened 2018-10-30
In printformats, use time 21:45 to distinguish hours/minutes and AM/PM formats.

sciurius:
To clearly show how the time is formatted, printformats should use a sample time like 21:45 instead of 10:09. This distinguishes hours from minutes, and 12/24 hour clocks.


geier:
Agreed, much better example, thanks for the PR!

geier:
(failing tests are not related)

PR #839, opened 2019-01-29
Add 'khal export' command

tpikonen:
I needed to create a webcal .ics file from a calendar and could not figure out an easy way to do it. So, I ended up adapting 'khal list' to output the raw ics strings of events, with the headers of the intermediate events trimmed out. The result is this PR implementing 'khal export'. 

The code is not tested much, but works for me. If this has a chance of being accepted, I can work on it some more and add tests, documentation etc.

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -460,6 +460,45 @@ def _get_cli():
 
     @cli.command()
     @multi_calendar_option
+    @click.option('--notstarted', help=('Export only events that have not started.'),
+                  is_flag=True)
+    @click.option('--outfile', '-o', help='The output file [defaults to stdout]',
+                  type=click.File(mode='w'), default='-', metavar='OUT')
+    @click.argument('DATERANGE', nargs=-1, required=False,
+                    metavar='[DATETIME [DATETIME | RANGE]]')
+    @click.pass_context
+    def export(ctx, include_calendar, exclude_calendar, daterange, notstarted, outfile):
+        """Export to an .ics file (or stdout).
+
+        Includes all events between a start (default: today) and (optional)
+        end datetime."""
+        try:
+            events = controllers.khal_export(
+                build_collection(
+                    ctx.obj['conf'],
+                    multi_calendar_select(ctx, include_calendar, exclude_calendar)
+                ),
+                daterange=daterange,
+                notstarted=notstarted,
+                conf=ctx.obj['conf'],
+                env={"calendars": ctx.obj['conf']['calendars']}
+            )
+            if events:
+                sl = events[0].raw.split('\n')
+                header = '\n'.join(sl[:3]) + '\n' # 3 first lines of first event
+                footer = '\n'.join(sl[-2:]) # last line of first event
+                outfile.write(header)
+                for e in events:
+                    sl = e.raw.split('\n')
+                    outfile.write('\n'.join(sl[3:-2]) + '\n')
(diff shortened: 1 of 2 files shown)


matt-snider:
I think this is an important thing to have! The issue referenced above is one good use case.

A few thoughts:

* A `--uuid` (or `--uid` to stay consistent) filter to specify a single event unambiguously 
* The date range filters seem more fitting as options than arguments, since as with `--uid`, more may be added later, or you may not want to filter by date at all, but by description, etc. 
* Make the outfile the only argument. If it is omitted it should print to stdout

matt-snider:
Another thing to consider, is whether it may make more sense to just extend the `search` command with this functionality. This command is actually just a way to filter events and output them in a particular format.

I think this would entail adding at least the following to `search`:

* `--output [ics, string]`
* `--date-range ...`

Adding `--uid` as described in the previous comment might also be useful.

chess-seventh:
I think having either the `--uid` parameter passed in the `search` or the `export` command would be great. Having this attribute for scripting (FWIW) would be indeed an asset.

Is there something to do/add in order to get this PR in merged? I'll gladly try to help! 
pinging new maintainer @ephase :)

ephase:
@chess-seventh I'm agree with you about `--uid` for search or export. Code seems good but as I'm newbie python developer, I think @d7415 should look into it.

@tpikonen could you add an option for uid export, if not I could try.

d7415:
> I think @d7415 should look into it. I'll try to test it this evening.

This one's on my list to look at. Probably next on my list. I don't know when that'll be though.

d7415:
We have an option for output. Currently this will overwrite the file if one exists.
Options:
- Document it, leave it as-is
- Prompt the user if the file exists
- Always output to stdout

I'm leaning to the first option, because it's easy. The second might be nicer, but then we'd need to go interactive, which limits scripting options. I only expect to use it with stdout (if at all), but the flexibility is nice.

WhyNotHugo:
Fourth option is to only overwrite it `--force` is passed. Safe, and script friendly.

d7415:
> Fourth option is to only overwrite it `--force` is passed. Safe, and script friendly.

True. It's a bit messy needing two commands just to write a file though. And "force" seems a bit strong as a verb.

madduck:
But `--force` is the "industry-standard" and it's what pretty much all other commands use in this situation.

d7415:
I mostly agree - I'd say industry standard is probably "just overwrite it" though!

But this is why I raised it here before doing starting work in any direction :)

madduck:
Yeah, I guess it depends on whether you clobber by default, or not. I think it's better to use `--force` in this case.

d7415:
Looks like a consensus.

Next question - Should we default other messages to stderr if we're outputting to stdout? Somewhat out of scope, but should we do that anyway? Or just recommend it in the docs using `-l`/`--logfile`?

WhyNotHugo:
For non-fatal output, we should use `stderr`, since one might want verbose scripts, but the output to be piped into something else.

For fatal output... I've no strong opinion either way. Probably writing to `stderr` reduces the changes of unexpected misbehaviour when piping, but people _should_ use stuff like `set -o pipefail`, or checking exit codes when writing scripts that pipe in general.

geier:
As nobody continued the work on this, I'll close it for now.

Issue #836, opened 2019-01-22
Today's date ignores local_timezone


ryannjohnson:
My config has the following:

```conf
[locale]
default_timezone = America/Los_Angeles
local_timezone = America/Los_Angeles
```

My computer's timezone is set to UTC. It's displaying `2019-01-22 06:30:00` right now.

In Los Angeles, the time is currently `2019-01-21 22:30:00`.

When I run khal, this is what I get:

```
$ khal
    Su Mo Tu We Th Fr Sa     Today, 2019-01-22
Jan 30 31  1  2  3  4  5     17:00-20:00 Meeting
     6  7  8  9 10 11 12     Tomorrow, 2019-01-23
    13 14 15 16 17 18 19     Thursday, 2019-01-24
    20 21 22 23 24 25 26     17:00-20:00 Meeting
Feb 27 28 29 30 31  1  2     Friday, 2019-01-25
     3  4  5  6  7  8  9     16:30-18:30 Meeting
    10 11 12 13 14 15 16     Saturday, 2019-01-26
    17 18 19 20 21 22 23     
Mar 24 25 26 27 28  1  2     
     3  4  5  6  7  8  9     
    10 11 12 13 14 15 16     
    17 18 19 20 21 22 23     
    24 25 26 27 28 29 30     
Apr 31  1  2  3  4  5  6
```

I expect `Today, 2019-01-22` to be the 21st of January, not the 22nd of January.

The event times show up correctly, but they are truncated if their day is not shown.

```
$ khal calendar 2019-01-21
    Su Mo Tu We Th Fr Sa     Monday, 2019-01-21
Jan 30 31  1  2  3  4  5     23:00-24:00 Today!
     6  7  8  9 10 11 12     Today, 2019-01-22
    13 14 15 16 17 18 19     17:00-20:00 Meeting
    20 21 22 23 24 25 26     Tomorrow, 2019-01-23
Feb 27 28 29 30 31  1  2     Thursday, 2019-01-24
     3  4  5  6  7  8  9     17:00-20:00 Meeting
    10 11 12 13 14 15 16     Friday, 2019-01-25
    17 18 19 20 21 22 23     16:30-18:30 Meeting
Mar 24 25 26 27 28  1  2     
     3  4  5  6  7  8  9     
    10 11 12 13 14 15 16     
    17 18 19 20 21 22 23     
    24 25 26 27 28 29 30     
Apr 31  1  2  3  4  5  6
```

---

[...]

PR #840, opened 2019-02-13
Refactor icalendar helpers into their own module

ryannjohnson:
This is a step towards resolving https://github.com/pimutils/khal/issues/836.

I'd created a function in `khal/utils.py` that I wanted to use in `khal/parse_datetime.py`, but I couldn't because of the circular dependency that would create.

```py
# khal/utils.py

import khal.parse_datetime as parse_datetime  # TODO get this out of here
```

I saw this TODO comment and decided to just go for it.

---

I've additionally included a small fix for the failing test in `tests/parse_datetime_test.py`. It now parses string categories on khal's side instead of letting [icalendar try to iterate through them](https://github.com/collective/icalendar/blob/f340b761aa3a4a6f6b456833cc29ef1b05cd9c14/src/icalendar/prop.py#L273).

Adding categories via `khal new --categories a,b,c ...` still seems to be broken, same as before.


geier:
Thanks for taking the time to improve khal and sorry for my inactiveness. The refactoring looks sensible and lots of good smalf fixes, if you could rebase it against master, I would merge.

ryannjohnson:
Done!

geier:
Thank you!

PR #843, opened 2019-03-05
Adding whitespace to event format self.status

whnr:
Same whitespace handling as for the cancelled status. The {status} field in the event format needs an additional whitespace, if it is present. Otherwise the status is either right next to the next item, or there are two spaces if there is no status.


whnr:
Reading the hacking notes first and making flake8 pass in the next try

PR #861, opened 2019-04-06
'/dev/tty' is not and should not be a regular file

wictory:
When I run `khal import` using `|` from mutt, khal aborts because there's a problem with `/dev/tty`. However, there's no problem with `/dev/tty` but khal thinks it should be a regular file instead of a special char device file.

This patch changes khal to expect a special char device file instead.

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -22,6 +22,7 @@
 import datetime as dt
 import logging
 import os
+import stat
 import sys
 import textwrap
 from shutil import get_terminal_size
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
This seems to work for me, and makes sense.

geier:
Thank you for taking the time to improve khal!

Could you explain this a bit for people (like me) who don't know enough about unix programming to make the difference more obvious?

wictory:
In the docs of [os.path.isfile](https://docs.python.org/3/library/os.path.html?highlight=isfile#os.path.isfile) it says:

> Return True if path is an existing regular file. This follows symbolic links, so both islink() and isfile() can be true for the same path.

In unix, a regular file is "just a file containing data on the disk", while `/dev/tty` is a character special file. I couldn't find any built-in python function to directly check whether a file a special file, so I decided to use [os.stat](https://docs.python.org/3/library/os.html#os.stat) which performs the [stat(2)](http://man7.org/linux/man-pages/man2/stat.2.html) system call. Using the result from [stat(2)](http://man7.org/linux/man-pages/man2/stat.2.html) it's possible to figure out the unix-filetype.

Is this the kind of answer you were looking for, or did I miss the point? :)

geier:
No, all good. Thanks again and merged!

PR #863, opened 2019-04-11
adding a .desktop file

teto:
Couldn't really test it because of this error
```
LookupError: setuptools-scm was unable to detect version for '/build/khal'.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
```
but should be ok


geier:
Thanks!

PR #874, opened 2019-08-13
document removal of default_command

Konfekt:


geier:
That was quick, thanks!

PR #878, opened 2019-08-21
Add support for other VCARD dates (e.g. X-ANNIVERSARY)

### LINKED ISSUE #243: Parse out other events similar to 'birthday'
I see `X-ANNIVERSARY` tags at least. There are probably others in use too.

d7415:
Fixes #243

2 problems at the moment (which Travis will report on shortly)

1) The definition line for `update_vcf_dates()` is too long (101 chars). I thought about `update_vcfdates()` or similar, but figured I'd push first and ask later.
2) There's a UNIQUE constraint violation. I can't see how the old function got around it. I'll keep looking, but help is welcome!

Other notes:
* I've changed the function and variable names to be more generic and added a docstring.
* I modified the `href` for the vcard-related events to be `href+vcard_fieldname`. This was because khal as a whole is set up for one event per file, and contacts may have e.g. a birthday and an anniversary. This might have caused (2) above...
* I might have a little time tomorrow, otherwise it will probably be a week or so before I can get back to this. Feel free to carry on in my absence or wait.
* Other feedback (style/approach/etc) welcome.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -37,3 +37,4 @@ Amanda Hickman - amanda [at] velociraptor [dot] info
 Raef Coles - raefcoles [at] gmail [dot] com
 Nito Martinez - nito [at] qindel [dot] com - http://qindel.com http://theqvd.com
 Florian Wehner - florian [at] whnr [dot] de
+Martin Stone
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,13 @@ may want to subscribe to `GitHub's tag feed
 <https://github.com/geier/khal/tags.atom>`_.
 
 
+0.10.2
+======
+not released
+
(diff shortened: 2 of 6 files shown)


geier:
I had a quick look through your changes and I like the approach. Thanks for taking the time to do this! If I find some more time, I'll see if I can help you with that unique problem.

d7415:
Ok. I still don't know why it used to work, but I think it should work now.

The `test_birthdays_update` test passes, the reason it passes makes sense, but I'm still a little concerned that I don't know how I broke it, so I might have missed something.

d7415:
Just to confirm that (now that I've stopped adding commits) I think this is ready for review/merge.

d7415:
> Just to confirm that (now that I've stopped adding commits) I think this is ready for review/merge.

Let's try that again.

geier:
Seems to work fine, thanks for your work!

PR #882, opened 2019-09-14
Fix wording in warning message

ddast:
Fix wording in warning message (since I'm seeing this quite often recently).


geier:
Thank you!

Issue #884, opened 2019-09-23
Khal new example doesn't work


p-h:
I copy pasted `khal new 25.10. 18:00 24:00 Another Event :: with Alice and Bob` from the [examples](https://khal.readthedocs.io/en/latest/usage.html) and it failed with the following error:

    critical: Could not parse "25.10. 18:00 24:00 Another Event".
    critical: Please check your configuration or run `khal printformats` to see if this does match your configured [long](date|time|datetime)format.
    critical: If you suspect a bug, please file an issue at https://github.com/pimutils/khal/issues/

d7415:
I'd guess it's the trailing dot in the date, but I can't test (and, if it is, either the docs or code still need updating).

If it's still pending at the end of the week I'm happy to take a look.

geier:
What's the output of `khal printformats`? Does it match `25.10. 18:00 24:00`?

d7415:
@p-h Do you have any update on this?

p-h:
I added

```
timeformat = %H:%M
dateformat = %d.%m.
longdateformat = %d.%m.%Y
datetimeformat =  %d.%m. %H:%M
longdatetimeformat = %d.%m.%Y %H:%M
```

to my configuration file and now it works. Not really a fix but it works now.

PR #885, opened 2019-09-27
Add note about date formats to `khal new` examples.

d7415:
Up to the maintainers whether this is necessary - Date formats are already mentioned in the description of `khal new` but it might help to cover off #884.


geier:
Thanks!

Issue #805, opened 2018-08-13
Ability to define default duration


madduck:
Creating an event and specifying only a single timestamp results in an event with a default duration of 1 hour. My default timeslot is actually 50 minutes, and I'd really like to be able to override this. Could this please be exported to the configuration file?

Thank you!

Jikhai:
Hi, we have two possible fixes for this issue, one adds a variable in the config file that sets the default duration of all events, the other allows to define a different default duration for each calendar. We however have yet to complete the unit tests. Which of the two options would be preferable ?

links to our fixes are respectively :
https://github.com/Khalzone/khal/tree/issue805
https://github.com/Khalzone/khal/tree/issue805b

madduck:
Hey @Jikhai, the simpler solution (config file) will do just fine for me. I am not sure I can fathom a use-case where the additional complexity would be worthwhile.

Jikhai:
Right now, we're waiting for @geier  's opinion on the matter before making a pull request.
We're also considering merging our two fixes so that both options can be used.
However, we'd like to know if you need this option only in `khal`  (command line) or in `ikhal` too (interface).

We've got most of it ready on hand, implementing it properly for `ikhal` will take just a bit more work, though.

madduck:
Sounds good, @Jikhai. I am personally a fan of small steps, so merging the config file option first, then creating a pull request for per-calendar, and then one for `ikhal` would be my preferred way forward, especially if I was the project maintainer. It's just easier to work through chunks at a time, rather than one big convult.

Jikhai:
Hey @madduck just a heads up,following your advice, and after going through all the testing, we made a pull request with just the config file fix ready, last Friday.
It is now pending for reviews.

PR #890, opened 2019-10-12
#805 : Add ability to define default duration

ephase:
For float and allday event, this can be set on config file : 
```
[default]
# Define the defaut duration for an allday event
default_duration_date = 1d

# Define the default duration for an event
default_duration_datetime = 1h
```

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -38,3 +38,8 @@ Raef Coles - raefcoles [at] gmail [dot] com
 Nito Martinez - nito [at] qindel [dot] com - http://qindel.com http://theqvd.com
 Florian Wehner - florian [at] whnr [dot] de
 Martin Stone
+Maxime Ocafrain
+Axel Danguin
+Yorick Barbanneau - git [at] epha [dot] se - https://xieme-art.org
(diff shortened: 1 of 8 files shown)


madduck:
I just had a look at this, and tried it out, and it's working perfectly.

The only feedback I have is:

* I am unsure of the use case to override the duration of an all-day event.
* I find `default_duration_datetime` to be confusing, possibly exposing too much implementation detail. How about `default_event_duration` instead?

Jikhai:
hey,
about all-day events, we thought it would be useful for uses like an on-call schedule, where you might want to schedule someone on a post for 36 or more hours instead of 24,
while keeping the other field for shorter events.

As for variable names, you are absolutely right, your idea sounds way better, we will adjust the PR to use ```default_event_duration``` and  ```default_dayevent_duration```

Jikhai:
Hi, 
We've taken care of the changes you requested, in both ```khal.spec``` and ```CHANGELOG.rst```.
We've also added our names to the Authors file (of all 5 of us, since it's a group effort).
All these changes are in the PR, ready for review.

geier:
Thank you all then!

PR #893, opened 2019-10-21
Add Python 3.7 version for testing

ephase:
As Python 3.7 is the default version for distribution like Arch or Manjaro, add it to `tox.ini`. I run all tests with this version and all work like a charm.


d7415:
Did this in #924 to fix the conflict.

PR #898, opened 2019-11-08
ikhal: Add 2 row in left pane when `frame` is `color` or `width`. Fixes #897

### LINKED ISSUE #897: ikhal: frame around leftpane breaks indent
Hi!

I'm using `khal, version 0.10.1`.

It seems that `ikhal` suffers from a ui bug on the left pane, when I go to the
right the current line loses its left indentation. This can be repeated with
the following lines. I think that a picture could help ;)

ikhal_frame_bug

- First I'm on the day "déc. 30" (any day on the last column);
- Then I press `right` or `l` or `space` and… the entire line go to the left;
- `right`, `down`, `right`, `down`… repeat this on several lines.

When I go to the left, the line appears correctly.

I tried some configurations, the bug is related to the `[view]` frame options
when set to `width` or `color`. Globally when there is a complete frame around
the calendar.

ephase:
This is a simple fix for the issue. I found that left pane calculation  was wrong if we add a frame around it. We must add 2 more char in `self.lwidth` when `cont['view']['frame']` contain `'width'` or `'color'`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,8 +20,10 @@ not released
 * FIX No warning when importing event with Windows timezone format
 * FIX Launching an external editor no longer crashes `ikhal`
 * UPDATED DEPENDENCY urwid>=1.3.0
+* FIX Wrong left pane width calculation in ikal when `frame` is `width` or 
+   `color` in configuration.
 
- 0.10.1
+0.10.1
 ======
 2019-03-30
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,7 @@ not released
 * FIX Launching an external editor no longer crashes `ikhal`
 * UPDATED DEPENDENCY urwid>=1.3.0
 
-0.10.1
+ 0.10.1
 ======
 2019-03-30
 
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -1061,6 +1061,8 @@ class ClassicView(Pane):
             elistbox.set_focus_date_callback = lambda _: None
         self.calendar = ContainerWidget(calendar)
         self.lwidth = 31 if self._conf['locale']['weeknumbers'] == 'right' else 28
+        if self._conf['view']['frame'] in ["width", "color"]:
+            self.lwidth += 2
         columns = NColumns(
             [(self.lwidth, self.calendar), self.eventscolumn],
             dividechars=0,


CHANGELOG.rst
```diff
@@ -20,8 +20,10 @@ not released
 * FIX No warning when importing event with Windows timezone format
 * FIX Launching an external editor no longer crashes `ikhal`
 * UPDATED DEPENDENCY urwid>=1.3.0
+* FIX Wrong left pane width calculation when `frame` is `width` or `color` in
+   configuration.
 
-0.10.1
+ 0.10.1
```
d7415: Extra space

CHANGELOG.rst
```diff
@@ -20,8 +20,10 @@ not released
 * FIX No warning when importing event with Windows timezone format
 * FIX Launching an external editor no longer crashes `ikhal`
 * UPDATED DEPENDENCY urwid>=1.3.0
+* FIX Wrong left pane width calculation when `frame` is `width` or `color` in
+   configuration.
```
d7415: It's probably worth mentioning that this is specific to `ikhal`
```suggestion
* FIX Wrong left pane width calculation in `ikhal` when `frame` is `width` or
   `color` in configuration.
```

khal/ui/__init__.py
```diff
@@ -1061,6 +1061,8 @@ def __init__(self, collection, conf=None, title='', description=''):
             elistbox.set_focus_date_callback = lambda _: None
         self.calendar = ContainerWidget(calendar)
         self.lwidth = 31 if self._conf['locale']['weeknumbers'] == 'right' else 28
+        if self._conf['view']['frame'] == "width" or self._conf['view']['frame'] == "color":
```
d7415: I think `in` would be easier to read here, and possibly (from very brief testing) faster.
```suggestion
        if self._conf['view']['frame'] in ["width", "color"]:
```

ephase:
@d7415 Thanks for replying, I need to work on Python specificities. I made the changes

ephase:
The bug only occurs on French locale, this is because french abbreviated month name is longer than 3 chars. Then we have 2 options :
 - add 2 row on left column when locale is french
 - force month name to 3 chars

d7415:
> The bug only occurs on French locale, this is because french abbreviated month name is longer than 3 chars.

Ooh. Interesting..

> Then we have 2 options :+1: 

It's your locale, but the 3 char limit looks like it would make june/july awkward. Is the default 5? Another option would be to add a number based on the length of the string. Thinking about it, does the "add 2 characters" option work with `mai`?

d7415:
This has been here a while now, so I'm going to merge it tomorrow unless someone beats me to it or raises an issue before then.

ephase:
@d7415 I'm ok with that :).

PR #899, opened 2019-11-15
Implementation for the issue809 (only khal)

Moustakick:
Added a solution for the issue #809 where you can define a specific character to put before the single digit numbers in khal, such as "0" to get the output to be  `01, 02, 03, etc` instead of `1, 2, 3, etc`.

The parameter I added can be found in "khal/settings/khal.spec" under the [view] section, named `date_format_view = string(default=' ')`.
Also added some "test cases" for `01, 02, 03, etc` output.

### DIFF
--- a/khal/calendar_display.py
+++ b/khal/calendar_display.py
@@ -51,6 +51,7 @@ def getweeknumber(date):
     return dt.date.isocalendar(date)[1]
 
 
+
 def get_calendar_color(calendar, default_color, collection):
     """Because multi-line lambdas would be un-Pythonic
     """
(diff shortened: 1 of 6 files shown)


geier:
Thank you for taking the time to improve khal! (And sorry for being several years late in having a look at this).

I don't think it's a good idea to implement #809 only for khal. I'm also not entirely sure that making even more things configurable is a good idea in the first place. @WhyNotHugo @d7415  @ephase  What's your feeling?

Also, this PR needs a bit of work (tests that test passing something else then "_") and also a bit of cleaning up. I'll close this for now, if you are still interested in getting this merged (and implementing the same functionality for ikhal) please re-open and I'll give it a proper review.

Issue #633, opened 2017-04-24
White text on bright yellow background in interactive mode


mortie:
The selected day in interactive mode is highlighted by having white text with a yellow background, like so: ![scr-3029579](https://cloud.githubusercontent.com/assets/3728194/25333148/7c2bc678-28e9-11e7-91e0-9a6ef83cfff9.png)

With many color schemes, this will probably look okay, but with others, like mine, the contrast is way too low. I don't think khal should assume that white text on a bright yellow background will be readable.

kflak:
I have the same issue, it is very close to impossible to read the highlighted text...

geier:
Should be fixed with 7b44b59, further refinement should be easier after #705

rslindee:
As of v0.9.10, the old "white text on bright-yellow background" is still the color scheme for date selected. Is there a way to override/modify this without changing the colors.py located in my /usr/lib/.../khal/ui directory?

rslindee:
As of v0.9.10, the colors appear to still be unreadable "white on yellow background" for "date selected". Is there any way to override/modify this without having to change /usr/lib/.../khal/ui/colors.py?

Or, a simple hotfix in the next release would certainly work.

geier:
can you send a screenshot of how it looks on your machine?

rslindee:
![screenshot-2019-04-01-21:24:46](https://user-images.githubusercontent.com/3173046/55376348-db3c5c00-54c4-11e9-956a-545453a56b1f.png)

This is with st term with what I understand to be no color/theme modifications.

geier:
With the merging of https://github.com/pimutils/khal/pull/1279 this can now be fixed in your config file. The how should be [documented](https://khal.readthedocs.io/en/master/configure.html#the-palette-section).  You probably want to override     `date header`, `date header focused` and `date header selected`. If you find something that works well for you and also improves the default theme, please send a PR.

Issue #395, opened 2016-04-08
ikhal interface text and highlight same color


ryanmorillo:
Fresh install after wiping my ~/khal directory left the same issue.  Days and Months are white on white. Haven't been able to figure out where the colors come from.  Day highlight is black on white, and selected day is black on aqua.  Help is blue on white.

geier:
Your terminal  emulator's color scheme probably doesn't play well with the chosen theme. Try setting `theme = light` (in section [view]). If that doesn't help try editing one of the color schemes in `ui/colors.py`.

bmarcote:
The default Dark theme scheme (khal/ui/colors.py) shows:
('dayname', 'light gray', 'white')
('monthname', 'light gray', 'white')
causing that the names of the days and months are invisible.

In my opinion, it should be changed and combine dark/light colors (e.g. dark gray, white; or the opposite) to allow the default theme to show properly the names.

geier:
@bmarcote thank you for looking into this.

jmoses:
The `interactive` mode for dark/today has yellow bg and white fg, and it's just this eye shattering blur.  I can't tell by the docs how to change that via config, so I just updated the `colors.py` data.

geier:
@jmoses: sadly I can't be done via the config right now.

geier:
as of https://github.com/pimutils/khal/pull/1279 themes can be overriden. Better defaults would still be nice, but this allows to fix specific issues such as this one.

PR #900, opened 2019-12-06
date_header_selected color modified for dark theme

pizzacoca:
Hi, Here's a little change for the #633 and that old #395 one.
Is that ok for you ?


geier:
Sorry for never getting back to you, can you perhaps show some screenshots of before and after? I can hardly see a difference with any of the dark presets in iterm2.

Issue #903, opened 2020-01-15
Test `test_popup()` on  `test_editor.py` failed with urwid v. 2.1.0


ephase:
With the latest urwid version, `test_popup()` failed with the message :

```
self = <[AttributeError("'CalendarWidget' object has no attribute '_wrapped_widget'") raised in repr()] CalendarWidget object at 0x7f043271eb50>
```
When I downgrade the version of urwid to 2.0.1 on my venv, this particular test file work like a charm.

ephase:
I found the commit that introduced this behavior on `urwid` code with git bisect : https://github.com/urwid/urwid/commit/018509d1b6af0cba9a76b2eea5648e63b46e5fd3

d7415:
I've found a dirty workaround. I'll see if I can figure out what's actually going on.

ephase:
I though to change the test by using render()method but i don't had time to test it these days.

PR #913, opened 2020-02-20
Add `{uid}` property to template options in `--format`. Fixes #891

### LINKED ISSUE #891: Return UID of new events
When creating an event from the command line, especially in batch mode, it would be amazing to have it provide some sort of unique identifier that can be used subsequently from a script to access that event, e.g. to export it. Maybe just printing the UID to `stdout` when a specific flag is passed?

d7415:
A quick fix that adds the `{uid}` as an option for use in formats.

I couldn't decide whether a test for this option was necessary as the current tests seem to test formatting more generically.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ not released
 * NEW Add ability to change default event duration with
    `default_event_duration` and `default_dayevent_duration` for an day-long 
    event
+* NEW Add `{uid}` property to template options in `--format`
 
 0.10.1
 ======
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -73,6 +73,12 @@ Several options are common to almost all of :program:`khal`'s commands
    description
         The description of the event.
 
+   description-separator
+        A separator: " :: " that appears when there is a description.
+
+   uid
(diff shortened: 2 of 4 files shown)


ephase:
Failing test is relative to #903

geier:
looks good! do you want to add a test?

d7415:
> looks good! do you want to add a test?

I've added it to the existing `test_print_ics_command` that seems to test most of the template options. Actual UID values are tested elsewhere. I could add a test like the `print_ics` one that tests for an exact output if that's desirable, but there's nothing like that at the moment for similar options.

geier:
Looks good to me! 

Am 25. Februar 2020 14:56:25 MEZ schrieb Martin Stone <notifications@github.com>:
>> looks good! do you want to add a test?
>
>I've added it to the existing `test_print_ics_command` that seems to
>test most of the template options. Actual UID values are tested
>elsewhere. I could add a test like the `print_ics` one that tests for
>an exact output if that's desirable, but there's nothing like that at
>the moment for similar options.
>
>-- 
>https://github.com/pimutils/khal/pull/913#issuecomment-590878876

-- 

PR #914, opened 2020-02-20
Force CalendarWidget to selectable. Fixes #903.

### LINKED ISSUE #903: Test `test_popup()` on  `test_editor.py` failed with urwid v. 2.1.0

d7415:
I'd like to know more about what's going on, but for now:
 - This fixes the test (#903)
 - "Playing around" on urwid 2.1.0 I couldn't see any problems, nor any other widgets that seemed to need the forced `selectable` property.
 - This was copied from the `EventEditor` class, but the same or similar is used on other widgets.

### DIFF
--- a/khal/ui/calendarwidget.py
+++ b/khal/ui/calendarwidget.py
@@ -643,6 +643,10 @@ class CalendarWidget(urwid.WidgetWrap):
     def reset_styles_range(self, min_date, max_date):
         self.walker.reset_styles_range(min_date, max_date)
 
+    @classmethod
+    def selectable(cls):
+        return True
+
     @property
     def focus_date(self):
         return self.walker.focus_date


d7415:
Ok, it fixed the test in my dev environment. I'll try harder :P

d7415:
That last change actually *did* fix it, but probably means my other change isn't required. Unfortunately I can only test with Travis. The current failure is on a `dateutil` build that's unrelated to this.

d7415:
For reference, the remaining failure seems to be https://github.com/spulec/freezegun/issues/335

PR #916, opened 2020-02-23
Remove testing for dateutil 2.6. Fixes #915.

### LINKED ISSUE #915: dateutil2.6 tests fail
[`freezegun` no longer supports `dateutil` < 2.7](https://github.com/spulec/freezegun/issues/333)

As far as I can see it, we can:

- [Drop testing for dateutil 2.6](https://github.com/pimutils/khal/pull/916)
- [Force the `dateutil26` set of tests to use an older version of `freezegun`.](https://github.com/pimutils/khal/pull/917)

PRs above.

d7415:
Option 1. As it's only testing that changes (it's `freezegun`'s dependency, not ours) I haven't added it to the changelog.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,6 @@ matrix:
         env: BUILD=pytz201702
       - python: 3.6
         env: BUILD=pytz_latest
-      - python: 3.6
-        env: BUILD=dateutil26
-      - python: 3.6
-        env: BUILD=dateutil_latest
       - python: 3.7
         dist: xenial
         sudo: true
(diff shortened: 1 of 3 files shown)


d7415:
Added #914 to make the tests happy.

ephase:
Since Debian stable use python-dateutils 2.7, i think this test is now deprecated and can be removed.

d7415:
> Since Debian stable use python-dateutils 2.7, i think this test is now deprecated and can be removed.

That was roughly my thinking (and `freezegun` has dropped support) but it was easy enough to fix that I figured I'd add #917 too and let @geier decide. Or allow an opportunity for comment, anyway :)

d7415:
I plan to merge this on Monday unless I hear any objections.

PR #921, opened 2020-03-08
Remove warning about missing timezones in repeated events. Fixes #830. Fixes #920.

### LINKED ISSUE #830: Erroneous warning about lack of timezone on UNTIL
I ran this command:

`$ khal new -r weekly '2018-12-01 9:00' 1h 'Test Event'`

It immediately printed this warning (the filename in question is the test event file it just created):

`warning: In event 9KQQRAX8X7C9ZII4QB4VKKE9T4PB0O10F7Q5.ics, DTSTART has a timezone, but UNTIL does not. This might lead to errenous repeating instances (like missing the last intended instance or adding an extra one).`

But there is no `UNTIL` value at all, never mind one without a timezone! The same warning happens if the `RRULE` has a `COUNT` value.

### LINKED ISSUE #920: Repeating events create warnings
I believe I'm having the same issue described in #830 .

Running gives

`warning: In event KRIYRYTUB902QB90AAM3T9LX22GANROFIN6X.ics, DTSTART has a timezone, but UNTIL does not. This might lead to errenous repeating instances (like missing the last intended instance or adding an extra one).`

for many different events. They all seem to be repeating ones. I also notice a correlation between how slow the program is and how many of these warnings are produced.

I really want to start using khal as my only calendar software, but the crashes make it close to unusable, so I'm just looking for an update on the issue. Does anyone have an idea of where the issue starts? I'll fix it myself if I'm able.

Here's the Dockerfile I'm running with.

```
FROM archlinux

ENV TZ='America/Chicago'

RUN pacman -Sy khal --noconfirm
RUN mkdir /khal
RUN mkdir /khal/routine
RUN mkdir /khal/home
```

d7415:
Looks like there are two issues here:

1) The simpler, more obvious problem that this error is raised even if `UNTIL` is not set. That's an easy fix (I have a branch with that if this check is still useful)

2) If a timezone not set on the command line (and, as far as I can tell, it can't be set from `ikhal`), the timezone of `DTSTART` is assigned implicitly (in `eventinfofstr()`, called by `new_from_string()`), before this check. This is not done for `UNTIL` and, as far as I can see, renders this check pointless as `DSTART` will always have a timezone regardless of the command line used, so we have no information about whether a timezone was passed with `DTSTART` to compare with `UNTIL`.

This patch removes the check, as it appears to have insufficient information to warn appropriately.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,9 @@ not released
 * UPDATED DEPENDENCY urwid>=1.3.0
 * FIX Wrong left pane width calculation in ikal when `frame` is `width` or 
    `color` in configuration.
+* CHANGE Remove check for timezones in `UNTIL` that aren't in `DTSTART` and 
+   vice-versa. The check wasn't fulfilling its purpose and was raising warnings
+   when no `UNTIL` value was set.
 
 0.10.1
 ======
--- a/khal/icalendar.py
+++ b/khal/icalendar.py
@@ -244,18 +244,6 @@ def expand(vevent, href=''):
         dtstart = vevent['DTSTART'].dt
         if events_tz:
             dtstart = dtstart.replace(tzinfo=None)
-        if events_tz and 'Z' not in rrule_param.to_ical().decode():
-            logger.warning(
-                "In event {}, DTSTART has a timezone, but UNTIL does not. This "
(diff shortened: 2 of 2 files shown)


d7415:
I plan to rebase and merge this tomorrow after #898 unless anyone objects.

PR #924, opened 2020-03-12
Add testing for Python 3.7. Resolves #893.

d7415:


d7415:
That test failure seems to be the test failing. I guess something changed its handling of assertions or `or` since the last build...

d7415:
The main change is so minor and I've included the test fix, so I'm going to merge this now.

PR #926, opened 2020-03-12
Drop support for Python 3.4

d7415:
No rush yet, but Python 3.4 is out of support, so we can probably drop it. That said we could leave it until it breaks, especially as this patch would block setup.py on 3.4.


geier (approved):

geier:
This probably needs a rebase, but I think it's time to merge this.

PR #928, opened 2020-03-19
Stop and start the whole urwid loop instead of the screen when launch…

### LINKED ISSUE #827: Khal crashes after external edit
I am having an issue where whenever I go make changes via the external edit  option in interactive mode (ikhal) no further input is registered and I have to stop the program forcefully.

d7415:
…ing an external editor. Fixes #827. Workaround for urwid/urwid#285.

I've added a version requirement for urwid >= 1.3.0 because the workaround requires it. Accordingly I've removed the FAQ entry about upgrading to urwid >= 1.1.0. 1.3.0 was released in 2014, so that's probably safe.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,12 +12,14 @@ may want to subscribe to `GitHub's tag feed
 ======
 not released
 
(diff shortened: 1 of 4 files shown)


d7415:
It's a simple commit so I'll aim to merge tomorrow.

PR #931, opened 2020-03-28
705

pizzacoca:
theme choice added for config file.

### DIFF
--- a/khal/settings/utils.py
+++ b/khal/settings/utils.py
@@ -57,6 +57,21 @@ def is_timedelta(string):
     except ValueError:
         raise VdtValueError("Invalid timedelta: {}".format(string))
 
+def is_theme(option):
+    """checks if *theme* is a valid value
+
+    :param option: the option the user set in the config file
+    :type option: str
+    :returns: off, left, right
+    :rtype: str/bool
+    """
+    option = option.lower()
+    if option in ['dark', 'light']:
+        return 'user'
+    else:
+        raise VdtValueError(
+            "Invalid value '{}' for option 'theme', must be one of "
+            "'dark' or 'light' ".format(option))
 
 def weeknumber_option(option):
     """checks if *option* is a valid value
--- a/khal/settings/utils.py
+++ b/khal/settings/utils.py
@@ -62,8 +62,8 @@ def is_theme(option):
 
     :param option: the option the user set in the config file
     :type option: str
-    :returns: off, left, right
-    :rtype: str/bool
+    :returns: user
+    :rtype: str
     """
     option = option.lower()
     if option in ['dark', 'light']:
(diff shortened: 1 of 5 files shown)


pizzacoca:
while coding I realized that my method is_theme in utils.py could parse colors.py nor khal.spec to check correct value.
Is that(s desirable ? For this time I use the same way other check methods.

geier:
Thank you for this PR and sorry for not looking at this for > 2 years.

I don't think we need the test for checking for valid color schemes as configobj already does this for us. Am I missing something?

fullstopslash:
Any chance on merging this?

khal/settings/utils.py
```diff
@@ -57,6 +57,21 @@ def is_timedelta(string):
...
+def is_theme(option):
+    """checks if *theme* is a valid value
+
+    :param option: the option the user set in the config file
+    :type option: str
+    :returns: user
+    :rtype: str
+    """
+    option = option.lower()
+    if option in ['dark', 'light']:
```
WhyNotHugo, line 69: **Nitpick**: we should actually check if this is defined in `colours.py`, rather than keep a copy of the theme names here.

This is non-blocking: I'm okay with addressing this in a separate PR.

khal/settings/utils.py
```diff
@@ -57,6 +57,21 @@ def is_timedelta(string):
...
         raise VdtValueError("Invalid timedelta: {}".format(string))
 
+def is_theme(option):
+    """checks if *theme* is a valid value
+
+    :param option: the option the user set in the config file
+    :type option: str
+    :returns: user
+    :rtype: str
+    """
```
WhyNotHugo, line 67: Use type hints instead:

```suggestion
def is_theme(option: str) -> str:
    """checks if *theme* is a valid value

    :param option: the option the user set in the config file
    :returns: user
    """
```

WhyNotHugo:
```
> git fetch origin pull/705/head
fatal: couldn't find remote ref pull/705/head

> git fetch origin pull/1005/head
From github.com:pimutils/khal
 * branch            refs/pull/1005/head -> FETCH_HEAD
```

I can't pull / checkout this specific PR to rebase/merge locally.

geier:
> Any chance on merging this?

What problem does this PR solve for you?

fullstopslash:
I thought this PR adds  the ability  to define colorscheme in the config file. That would be rather  snappy, and generally accepted as standard feature among command-line apps. It's kind of a pain to manually edit my colours.py every time I update ikhal.

WhyNotHugo:
This PR sounds like a first step for what you want -- it still doesn't allow defining new themes in user-defined locations, just picking one of the buit-in themes via the config file.

tgy:
I think this can be closed in favour of https://github.com/pimutils/khal/pull/1279

PR #938, opened 2020-05-16
First implementation for date detection in string_search

ephase:
This is my very first working work for search improvement. In this version, you can specify a date range for limit search results.

Usage (for now) is `<start> <end> <search_string>`, you can't indicate `end` date without `start` exemple :

```
khal search 01.01.2020 Dentist
```

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -545,7 +545,7 @@ def _get_cli():
     @multi_calendar_option
     @click.option('--format', '-f',
                   help=('The format of the events.'))
-    @click.argument('search_string')
+    @click.argument('search_string', type=str, nargs=-1)
     @click.pass_context
     def search(ctx, format, search_string, include_calendar, exclude_calendar):
         '''Search for events matching SEARCH_STRING.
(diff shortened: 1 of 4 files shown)


ephase:
I have another idea for add date and others element in search. It could be easy and simple to add keyword with double column. for example, search events with dentist before 2020.01.24 : 

```
khal search before::2020.01.24 dentist
```
Other example, search appointment with friends between 2019.06.04 and 2019.07.23 : 

```
khal search before::2019.07.23 after::2019.06.04 Friends
```

It could be easy to add some search element like `location::`, `summary::`, etc.

geier:
I like this, but perhaps we should make this more explicit by giving start and end as an argument, e.g. `search --start 2022-05-01 Denstist`. Then one could also give start without end.

geier:
I like this, but perhaps we should make this more explicit by giving start and end as an argument, e.g. `search --start 2022-05-01 Denstist`. Then one could also give start without end.

geier:
Please re-open if there is any interest in working on this.

PR #943, opened 2020-06-17
Remove  `tuple=` in sql_ex(), fix #942

### LINKED ISSUE #942: Error with mypy event in master branch
When I try to run mypy on khal project I've got this error 
```
tests/ui/test_calendarwidget.py:6: error: Need type annotation for 'on_press' (hint: "on_press: Dict[<type>, <type>] = ...")
khal/khalendar/backend.py:190: error: Incompatible default for argument "stuple" (default has type "object", argument has type "Tuple[Any, ...]")
tests/cli_test.py:405: error: Not all arguments converted during string formatting
```

For the `backend.py` file, the source of problem is `sql_ex()`, the `stuple: tuple=Union[tuple, str]` parameter. Removing the `tuple=` part avoid the problem.

ephase:
Fix mypy error for `backend.py`, other errors no need to be fixed and are present for a long time

### DIFF
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -28,7 +28,7 @@ from enum import IntEnum
 import logging
 import sqlite3
 from os import makedirs, path
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import icalendar
 import pytz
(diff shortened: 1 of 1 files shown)


d7415:
Ah. I hit that failure in the export PR...

ephase:
I'll merge this PR on Friday

PR #954, opened 2020-09-12
Option to use the `multiple` color only when not all calendar colors can be displayed

agschaid:
Sorry. No tests. Didn't find any that were concerned with coloring so I wasn't sure where to begin.

Not too happy that one and the same color-management logic is duplicated (in quite different ways) for `ikhal` and `khal calendar`. Those calendar columns should always look the same for both cases, right? Might make sense to refactor and unify that piece of logic. That would make it easier to test too. But I wasn't brave enough to do that just yet.


agschaid:
closed in favor of https://github.com/pimutils/khal/pull/957 which is created from its own feature branch

PR #957, opened 2020-09-14
Option to use the `multiple` color only when not all calendar colors can be displayed

agschaid:
Sorry. No tests. Didn't find any that were concerned with coloring so I wasn't sure where to begin.

Not too happy that one and the same color-management logic is duplicated (in quite different ways) for `ikhal` and `khal calendar`. Those calendar columns should always look the same for both cases, right? Might make sense to refactor and unify that piece of logic. That would make it easier to test too. But I wasn't brave enough to do that just yet.


ephase:
Hi @agschaid and thank you for your contribution. It seems useful but I think you should remove the 4e0de92 commit as every khal contributor have his own way to work with `venv`

agschaid:
yeah. Right. Will do that. 
Thought about moving that to my global gitignore but forgot . .

agschaid:
@ephase 
dropped that one commit as requested.

geier:
Thank you! And sorry for being so late in merging this.

agschaid:
No problem. To be honest I forgot about that ;)

PR #958, opened 2020-09-14
Filter from highlighting

### LINKED ISSUE #956: Discussion: filter calendar widget
Hi,

I'd like to propose a feature and also offer to implement it. The idea is to filter events or whole calendars from the day highlighting in the calendar while keeping the agenda untouched.

For that matter I would introduce a calendar-level config `filter_from_highliting` that can either be a boolean or a string and defaults to `False`

- when set to `True` the whole calendar would not be considered for day highlighting. A usecase would be birthday calendars that I myself want to be featured in the agenda in a prominent color (thus preventing me from just setting an empty calendar color) but don't want to show up in the calendar widget.
- when set to a string it would considered to be a (python-re-style) regex that filters events from the calendar whose title matches the regex. For example on my business calendar I would use the regex `"(^Daily)|(^Away:)"` which would filter both my Scrum dailies and away-notifications from colleges so a day in the calendar would only be highlighted in the color of the business calendar if I really have a meeting.

I had a deep dive into the relevant code parts yesterday. So I have a fair idea what I am getting myself into (including changes to sqlite schemata) and how to attack that problem.

Would that feature be interesting for you? Any changes you would suggest?

However as I already observed [in a merge request](https://github.com/pimutils/khal/pull/954) the code implementing the day highlighting is pretty much duplicated for `ikhal` and `khal calendar`. So before making it even more complicated I would like to refactor it into a single code base for both usages that is actually testable and decoupled from the "dirty business of really applying the colors".

agschaid:
fixes #956 

that was not as easy as I thought. And much dirtier too . . .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ dist/
 htmlcov
 doc/source/configspec.rst
 .mypy_cache/
+.venv
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -165,6 +165,7 @@ def build_collection(conf, selection):
                     'color': cal['color'],
                     'priority': cal['priority'],
                     'ctype': cal['type'],
+                    'filter_from_highlighting': cal['filter_from_highlighting']
                 }
         collection = khalendar.CalendarCollection(
             calendars=props,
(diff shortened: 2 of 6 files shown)


agschaid:
that change was dirty to begin with. Now there are conflicts too. And I don't think I will get this into shape anytime soon. Hope it's ok if I just close.

PR #969, opened 2020-10-26
Change default format of printics to event_format

### LINKED ISSUE #941: Sensible default for printics
printics currently uses agenda_event_format as default:

https://github.com/pimutils/khal/blob/a401ef28076954779cc66885a05dd1fa7da46fe3/khal/controllers.py#L625

Therefore, the date of the event is missing, when it is used to quickly check the contents of an ICS file, kind of missing its purpose. If I do not miss another use case where a date context is somehow given, I suggest to use event_format as format.

Or - in my opinion even better - define an own config value for printics_format that is used and can be tweaked for own purposes without affecting other parts of khal.

ddast:
It seems more reasonable to use `event_format` instead of `agenda_event_format` as the default format for the `printics` command, since the date ist not clear through context.

Fixes #941

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -622,7 +622,7 @@ def import_event(vevent, collection, locale, batch, format=None, env=None):
 
 def print_ics(conf, name, ics, format):
     if format is None:
-        format = conf['view']['agenda_event_format']
+        format = conf['view']['event_format']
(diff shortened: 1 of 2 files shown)


d7415 (approved):
Sorry for the delay. Looks good to me.

PR #972, opened 2020-11-04
[icalendar] Support rrules that ended in the past

david-caro:
Currently we were throwing an UnsupportedRecurrenceError when there
was an rrule that happened to have been finished (RRULE:UNTIL) in the
past.

This just returns an empty list of events when expanding instead.

This also adds freezegun (through pytest-freezegun) to the
khalendar_utils_test in order to be able to test this case, as most
tests were using old dates.


ephase:
Seems good to me, can you add yourself in `AUTHORS.rst`? Then I'll merge your PR by the end of the week if there is no objections.

ephase:
I merged then revert back because of problem when I tested. I lost all information about recurred event in the past. All of this need to be discuss with others maintainers.

d7415:
Ok, so looking at this properly, @david-caro's conclusion appears to be incorrect. khal fails it the RRULE UNTIL is before the DTSTART, which seems reasonable to me. This PR is not required.

Example of an event that works with an UNTIL well in the past:
```
BEGIN:VEVENT
SUMMARY:An Event
DTSTART;VALUE=DATE-TIME:19801109T193000
DTEND;VALUE=DATE-TIME:19801109T203000
RRULE:FREQ=DAILY;COUNT=10;UNTIL=19801119T203000Z
DTSTAMP;VALUE=DATE-TIME:20140401T234817Z
UID:V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOV
END:VEVENT
```

david-caro:
This is happening to me with several events imported from Google calendar, I'll try to get some examples and verify the behaviour and the cause.

d7415:
Having found 3 issues on the same subject, I've added a warning in #978 (not yet merged). khal still won't display them, but I don't think it would with this patch either.

PR #973, opened 2020-11-05
Attendee widget

Schubisu:
Hi everyone,
I've edited the khal `editor.py` to include an attendee widget. Attendees are read from and written to icalendar files through the `Event` class in `event.py`.

As of now, attendees can be specified as comma separated email addresses. I think it could be convenient to add an option to either resolve the email addresses from address books (e.g. khard) internally, and/or improve the attendee widget to browse and search addressbooks for contacts.

I've read the contribution guidelines and I'm sorry I didn't discuss this feature before creating this PR. Since it didn't need many code changes and there was no reaction on #pimutils, I thought it would be easiest to propose this feature this way.

Please let me know what you think; I'll be happy to add documentation, tests and also improve the feature further, if it meets your approval.

Current status of implementation:
- The attendee widget is implemented as simple text field, that accepts comma separated email addresses (input is not checked)
- Attendees are read and written from and to .ics files
- Attendees are added as `icalendar.props.vCalAddress` with parameter `ROLE = vText("REQ-PARTICIPANT")`. Adding further params for better calendar integration would be possible and could be implemented per widget or settings 
- Currently, no CN information is added. Events with existing attendee information that do have CN specified will have the CN information (and additional params) stripped on change.

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -502,7 +502,7 @@ class Event:
                         vCalAddresses.append(old_attendee)
                         unchanged_attendees.append(attendee)
             for attendee in [a for a in attendees if a not in unchanged_attendees]:
-                item = icalendar.prop.vCalAddress('MAILTO:{}'.format(attendee))
+                item = icalendar.prop.vCalAddress(f'MAILTO:{attendee}')
                 item.params['ROLE'] = icalendar.prop.vText('REQ-PARTICIPANT')
                 item.params['PARTSTAT'] = icalendar.prop.vText('NEEDS-ACTION')
                 item.params['CUTYPE'] = icalendar.prop.vText('INDIVIDUAL')
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -968,7 +968,7 @@ class EventDisplay(urwid.WidgetWrap):
         if event.attendees != '':
             lines.append(urwid.Text('Attendees:'))
             for attendee in event.attendees.split(', '):
-                lines.append(urwid.Text('  - {}'.format(attendee)))
+                lines.append(urwid.Text(f'  - {attendee}'))
 
         # start and end time/date
         if event.allday:
--- a/tests/event_test.py
+++ b/tests/event_test.py
@@ -3,7 +3,8 @@ import datetime as dt
 import pytest
 import pytz
 from freezegun import freeze_time
-from icalendar import vRecur, vText, vCalAddress, Parameters
+from icalendar import Parameters, vCalAddress, vRecur, vText
+
 from khal.khalendar.event import (AllDayEvent, Event, FloatingEvent,
                                   LocalizedEvent, create_timezone)
 


ephase:
Hi @Schubisu and thank you for your contribution.

There is one syntax error with Python version below 3.5, as Python 3.5 reached end-of-life (since august 2017) I'm not sure this is important (Maybe we can remove support for this version in khal).

You need to add yourself in contributor and if you can add some tests and documentation.

Schubisu:
Thanks @ephase, I have added some tests and updated the `CHANGELOG.txt`. I also changed the f string to a `str.format()` to also support older versions of Python. I didn't add any documentation, yet, since I believe there are no docs for the interactive editor. I could however also implement the cli options to add attendees and document that accordingly.

tbabej:
This would be a great addition to khal. @Schubisu Are you using this in your local install?

tbabej (approved):
@geier I tested the patchset on the current master. Works to me functionality-wise, and the code looks OK, however, I'd suggest adding the information about the attendees into the `EventDisplay` too, so that it shows in `ikhal`, something like:
```
diff --git a/khal/ui/__init__.py b/khal/ui/__init__.py
index ddf987e..b6ffcfc 100644
--- a/khal/ui/__init__.py
+++ b/khal/ui/__init__.py
@@ -960,6 +960,12 @@ class EventDisplay(urwid.WidgetWrap):
         if event.categories != '':
             lines.append(urwid.Text('Categories: ' + event.categories))
 
+        if event.attendees != '':
+            lines.append(urwid.Text('Attendees:'))
+            for attendee in event.attendees.split(', '):
+                lines.append(urwid.Text(f'  - {attendee}'))
+
+
```

Schubisu:
@tbabej yes, I'm using that locally every day, it works fine for me. I noticed one issue though
an empty `ATTENDEES` line would be added to exported ics files even if event.attendees were empty, because the parsed argument was `['']` instead of an empty list, I fixed that in b97eb78.

Thanks for the suggestion in `EventDisplay`, I tried it and it's a nice improvement, so added this as well in f25b560 .

geier:
@Schubisu Thank you for this PR!

This is a good first step towards supporting invitations!

I see that this currently removes all parameters it doesn't know about (it only knows about email addresses). Meaning, if one edits an event and then syncs it again, data loss can occur.

I'm unsure if it's a good idea to merge this, without at least adding a big warning or even enable it only after the user has set something in their config.

What are the oppinions of other users and developers of khal? @ephase @WhyNotHugo

Schubisu:
@geier, @tbabej  thanks for the positive feedback. Let me try to improve this, so there will be no data loss. I'm not an expert with `VEVENTS` but I think I should manage to do that.
I would also like to add an option to the `cli` to allow adding attendees from command line, but I guess that should better be a second PR.

Schubisu:
@geier I've now updated the code, so `update_attendees` will copy `vCalAddress` objects of attendees that were not changed. This way their `Parameters` should be preserved.

I still plan to add cli support, and also would like to add an email address parser, so the widget can accept formats like
- abc@gmail.com
- Firstname Lastname <abc@gmail.com>
- "Lastname, Firstname" <abc@gmail.com>

and also add a `CN` prop if found. But for now, this should not cause any loss of data anymore.

geier:
@Schubisu thanks for your hard work and sorry for my very late reply (again).

I rebased the PR and merged the branch, then force-pushed into your branch believing the PR would then also be merged, but now it's just closed. Sorry for that. The code is commited anyway.

Issue #790, opened 2018-05-27
RRULE with UNTIL, again?


sciurius:
Using the git version, the attached importfail.ics crashes with UnsupportedRecurrenceException.
Attached importok.ics imports with a warning.
The difference is that in importfail.ics, one of the EXDATE exceptions falls after the RRULE UNTIL.

It is my understanding that this was fixed (issue #591). Am I overlooking something?

[importfail.ics](https://github.com/pimutils/khal/files/2042350/importfail.txt)
[importok.ics](https://github.com/pimutils/khal/files/2042351/importok.txt)

david-caro:
Might be fixed in #972

ephase:
Hi,

Is it correct to have an EXDATE after the event end of he event? I don't find any information about that in RFC5545 but this case seem a bit bizarre.

sciurius:
I think such a situation can result from the following scenario:
1. Create an event that repeats until the end of the year (2020-12-31)
2. Cancel one of the future occurrences (e.g. 2020-12-04). This adds an EXDATE.
3. Change the end date to before the cancelled event (e.g. 2020-11-27). RRULE UNTIL is now set before the EXDATE.

This is the resultant ics from the Nextcloud calendar client:
````
BEGIN:VCALENDAR
PRODID:-//IDN nextcloud.com//Calendar app 2.1.2//EN
CALSCALE:GREGORIAN
VERSION:2.0
BEGIN:VEVENT
CREATED:20201105T083243Z
DTSTAMP:20201105T083648Z
LAST-MODIFIED:20201105T083648Z
SEQUENCE:4
UID:f3ad5798-1461-4dab-a728-e6ea0a979712
DTSTART;TZID=Europe/Amsterdam:20201106T100000
DTEND;TZID=Europe/Amsterdam:20201106T110000
RRULE:FREQ=WEEKLY;BYDAY=FR;UNTIL=20201127T000000
EXDATE;TZID=Europe/Amsterdam:20201204T100000
END:VEVENT
BEGIN:VTIMEZONE
...
END:VTIMEZONE
END:VCALENDAR
````

Given that there is no clear information on what to do in this case (i.e. it is not strictly forbidden) I assume most clients just leave it as is.

ephase:
Thank you, that's clear now. As you said there is not information about this special case in the RFC.

d7415:
[...]

Issue #771, opened 2018-04-10
khal.khalendar.exceptions.UnsupportedRecurrence


dario23:
i have a recurring event in one of my calendars, that both http://ical-validator.herokuapp.com/validate/ and https://icalendar.org/component/com_icalvalidator/ say is basically valid (first one complains about line ending style, but i don't think that's an issue here).

khal fails to parse it with this backtrace:

> Unknown exception happened.
> Traceback (most recent call last):
>   File "/usr/lib/python3/dist-packages/khal/khalendar/khalendar.py", line 303, in _update_vevent
>     update(event.raw, href=href, etag=etag, calendar=calendar)
>   File "/usr/lib/python3/dist-packages/khal/khalendar/backend.py", line 262, in update
>     self._update_impl(vevent, href, calendar)
>   File "/usr/lib/python3/dist-packages/khal/khalendar/backend.py", line 355, in _update_impl
>     dtstartend = utils.expand(vevent, href)
>   File "/usr/lib/python3/dist-packages/khal/khalendar/utils.py", line 112, in expand
>     raise UnsupportedRecurrence()
> khal.khalendar.exceptions.UnsupportedRecurrence
> warning: Skipping cip
> warning: /f25edc51-fac1-4b55-9bdb-78b7207b1d1d.1476835067135.ics: 
> warning: This event will not be available in khal.

as far as i managed to read [RFC 5545](https://tools.ietf.org/html/rfc5545), the RRULE

> RRULE:FREQ=WEEKLY;UNTIL=20161017T095959Z;BYDAY=MO,TU,WE,TH,FR

and those in the VTIMEZONE element

> RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=3
> RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=10

seem to be valid in terms of the rfc.

[ics.txt](https://github.com/pimutils/khal/files/1894674/ics.txt)

bronzehedwick:
I have the same issue. Here's my error output:

```
Unknown exception happened.
Traceback (most recent call last):
  File "/usr/local/Cellar/khal/0.9.10/libexec/lib/python3.7/site-packages/khal/khalendar/khalendar.py", line 303,
in _update_vevent
    update(event.raw, href=href, etag=etag, calendar=calendar)
  File "/usr/local/Cellar/khal/0.9.10/libexec/lib/python3.7/site-packages/khal/khalendar/backend.py", line 262, in
 update
[...]

PR #978, opened 2020-11-09
Add a warning for the special case where RRULE:UNTIL is before DTSTART (#771,#790,#972)

d7415:
#771, #790 and #972 all refer to events where the `RRULE:UNTIL` was before the event's `DTSTART`. This patch will stop khal from attempting to handle such events and will log a warning about it.


d7415:
I'm not entirely sure that this is the best option. Part of me thinks we should still raise `UnsupportedRecurrence`, but add the message for context. Ideas welcome.

ephase (approved):
Seems good for me, tested with my own calendars all work like a charm.

d7415:
Thanks @ephase. Do you have any thoughts about raising the UnsupportedRecurrence? I'm leaning towards no as I've already said `This event will not be available in khal` so a Traceback may just be distracting. I'll probably wait a week or so before merging to allow for other opinions though.

ephase:
I'm ok with you, Exception is not necessary here, your message is enough.

PR #987, opened 2020-11-28
Improve sorting of events

### LINKED ISSUE #976: khal interactive allday events at bottom of day.
`khal list` will show an agenda with  any  all  day events  at the  top  of the day (which seems right), while `khal interactive` shows events listed at the bottom of  the day (which seems counter intuitive to me) . Tried poking around the docs to see if there was some setting I missed and I couldn't seem to find one.

### LINKED ISSUE #981: Proposal: Improve sorting
Currently events are sorted by start date/time only. I would like to change to sort by start, then end, then summary.

Comments welcome.

d7415:
Sort by start, then end, then summary.

Closes #981. Fixes #976.

This may need some refactoring, maybe some comments and probably needs some tests. But the concept should be ready for review.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ not released
 * NEW Possibility to add a blank line before day in `khal` with
    `blank_line_before_day` option
 * FIX `new` keybinding in search no longer crash `ikhal`
+* NEW Improved sorting of events. Sort by `DTSTART`, `DTEND` then `SUMMARY`.
 
 0.10.2
 ======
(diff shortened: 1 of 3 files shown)


geier:
Looks sensible, do you have the time to write a test? If not, I could do it.

d7415:
I'll see what I can manage tomorrow. It's been a while since I've looked at this, but I guess it should just be a set of events to check each level and ideally conflicting types.

d7415:
Rebased on master and added tests.

geier:
Awesome, thanks!

PR #989, opened 2020-12-07
Set microsoft busy status

whit2333:
- The status is set to busy by default for microsoft.
- Need to add option to set value.


geier:
@whit2333 Thank you for helping improve khal.

I am not sure, if setting all calendar entries to busy is the right approach. IMO the busy status should be editable in ikhal and probably in the command line version as well.

geier:
closing and reopening to re-run tests

WhyNotHugo:
See: https://github.com/pimutils/khal/pull/1220

geier:
please re-open if there is any interest in working on this.

PR #1001, opened 2021-01-08
Fix small typo in usage.rst

pianoslum:
I am not a native speaker, but I guess I found a small typo, cf. https://www.merriam-webster.com/words-at-play/keeping-up-with-passed-and-past


d7415:
I'd accept either, but you're technically correct. Thanks!

PR #1005, opened 2021-01-18
All day event process correctly with --notstarted, fix #1004

### LINKED ISSUE #1004: Day events and --notstarted
Not sure if this is intended behaviour, but currently I have a day event for today:

```
❯ khal list
Today, 2021-01-13
 test ⟳
```

and I would expect that by doing:

```
❯ khal list --notstarted now 30m                                                                                                                                                                                    
```

would filter out the already-started event, but the output is the same.

ephase:
When you do `khal list --notstarted`, the "today" all-day event is displayed. This MR avoid this problem.

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -187,7 +187,7 @@ def get_events_between(
         # yes the logic could be simplified, but I believe it's easier
         # to understand what's going on here this way
         if notstarted:
-            if event.allday and event.start < original_start.date():
+            if event.allday and event.start <= original_start.date():
                 continue
             elif not event.allday and event.start_local < original_start:
                 continue


d7415 (approved):
I don't see a problem with this :)

d7415:
Hmm, actually...

d7415 (changes requested):
It took a while, but (like I said in #1004) this would require us to change the documented behaviour. This would also prevent excluding multi-day events (that have started before the chosen day) but showing new events on that day. It's also confusing because in this form `khal list --notstarted` will show events that have started today (or even ended) *unless* they're all day events.

d7415:
Fixing the bug (that `khal list now --notstarted` still shows all-days) without breaking this looks to be a pain.

ephase:
Hi @d7415, so if I understand, I need to create a exception to display multi-day events? For example event begins yesterday and ends in 3 day, `khal list now --notstarted` should display my event for tomorrow and next days?

d7415:
> Hi @d7415, so if I understand, I need to create a exception to display multi-day events? For example event begins yesterday and ends in 3 day, `khal list now --notstarted` should display my event for tomorrow and next days but not for today?

No. In that example the event has started so shouldn't be shown (and the current code won't show it)

My problems with this PR are:

- There should be a way to hide events like your example, while showing all-day events today (e.g. hide that I'm on holiday all week, show that it's my friend's birthday today). The current code does this, but the PR does not.
- This PR results in an inconsistency where running `khal list --nostarted` at midday will hide an all-day event for today, but show an event that happened this morning.

There is still a bug from #1004 where if a time after midnight *is* specified (e.g. with `now`) all day events show in `--notstarted`. But that looks like we'd have to special-case a time (other than midnight) being specified, which is probably why it wasn't done in the first place. (Well, that and showing today's all-day events in `--notstarted` isn't likely to cause a major inconvenience)

PR #1018, opened 2021-03-07
Include readthedoc badge in Readme

geier:


d7415 (approved):
Looks good. I wonder whether linking to https://khal.readthedocs.io/en/latest/ might be more useful, but it would also make sense pointing this to the project page as we link to the documentation itself elsewhere.

geier:
@d7415 makes sense and fixed.

Issue #1020, opened 2021-03-11
Unsupported Recurrence


Petemir:
Hello,

I am just starting with khal, I wanted first to thank all maintainers for their effort in supporting this interesting type of software :) .

That said, I just synced with vsyncdir some Google calendars and I got some errors when trying to start khal (installed via pip, so I have the latest release). Some information has been redacted (cal name, event id, paths, and event name).

I'll add the details as per [here](https://khal.readthedocs.io/en/latest/feedback.html), if there's anything else that you may need let me know. Thanks!

* how you ran khal (please run in verbose mode with -v DEBUG):
"khal", "khal list" , "khal interactive"

This doesn't happen always, so I did it first, found the error, and then trying to run it with -v DEBUG didn't return the error... but then I just did a normal "khal interactive" again and the error appeared again :man_shrugging: .

* what you expected khal to do:
To start without any warnings nor errors

* what it did instead:
It printed an error

* everything khal printed to the screen (you may redact private details):
```
warning: Skipping $CALNAME/$EVENTID.ics: 
warning: This event will not be available in khal.
Unknown exception happened.
Traceback (most recent call last):
  File "~/anaconda3/envs/newbase/lib/python3.8/site-packagesy", line 338, in _update_vevent
    update(event.raw, href=href, etag=etag, calendar=calendar)
  File "~/anaconda3/envs/newbase/lib/python3.8/site-packages, line 240, in update
    self._update_impl(vevent, href, calendar)
  File "~/anaconda3/envs/newbase/lib/python3.8/site-packages, line 356, in _update_impl
    dtstartend = expand_vevent(vevent, href)
  File "~/anaconda3/envs/newbase/lib/python3.8/site-packages72, in expand
    raise UnsupportedRecurrence()
khal.exceptions.UnsupportedRecurrence
```
* in case khal complains about a specific .ics file, please include that as well (or create a .ics which leads to the same error without any private information)

```
BEGIN:VCALENDAR
PRODID:-//Google Inc//Google Calendar 70.9054//EN
VERSION:2.0
CALSCALE:GREGORIAN
X-WR-CALNAME:$CALNAME
X-WR-TIMEZONE:Europe/Zurich
BEGIN:VTIMEZONE
TZID:America/Argentina/Buenos_Aires
X-LIC-LOCATION:America/Argentina/Buenos_Aires
BEGIN:STANDARD
TZOFFSETFROM:-0300
TZOFFSETTO:-0300
TZNAME:-03
[...]

PR #1022, opened 2021-03-11
Add warning when an event with an RRULE will never occur (e.g. #1020) and stop processing.

d7415:


WhyNotHugo (approved):
LGTM. This behaviour aligns with the RFC, and makes perfect sense.

geier:
Agreed. Tests also run locally.

PR #1023, opened 2021-03-12
Let pytest remove all tmp files and folders

geier:
In some tests we set some temporary directories' permissions so that those directories are not accessible anymore. We now reset those permissions afterwards so that they can be cleaned up by pytest after the tests have run.

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -761,12 +761,25 @@ default_calendar = private
     assert '{}/khal/calendars/private1' .format(runner.xdg_data_home) in actual_config
 
 
+def cleanup(paths):
+    """reset permissions of all files and folders in `paths` to 644 resp. 755"""
+    for path in paths:
+        if os.path.exists(path):
+            os.chmod(str(path), 0o755)
+        for dirpath, dirnames, filenames in os.walk(path):
+            os.chmod(str(dirpath), 0o755)
+            for filename in filenames:
+                os.chmod(str(os.path.join(dirpath, filename)), 0o644)
+
+
 def test_configure_command_cannot_write_config_file(runner):
     runner = runner()
     runner.config_file.remove()
     os.chmod(str(runner.xdg_config_home), 555)
     result = runner.invoke(main_khal, ['configure'], input=choices())
     assert result.exit_code == 1
+    # make sure pytest can clean up behind us
+    cleanup([runner.xdg_config_home])
 
 
 def test_configure_command_cannot_create_vdir(runner):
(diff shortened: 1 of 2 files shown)


tests/cli_test.py
```diff
@@ -761,12 +761,25 @@ def test_configure_command_create_vdir(runner):
...
 
 
+def cleanup(paths):
+    """reset permissions of all files and folders in `paths` to 666 resp. 777"""
+    for path in paths:
+        if os.path.exists(path):
+            os.chmod(str(path), 0o777)
+        for dirpath, dirnames, filenames in os.walk(path):
+            os.chmod(str(dirpath), 0o777)
+            for filename in filenames:
```
d7415, line 771: Should this loop through `dirnames` as well?

tests/cli_test.py
```diff
@@ -761,12 +761,25 @@ def test_configure_command_create_vdir(runner):
...
+
+
 def test_configure_command_cannot_write_config_file(runner):
     runner = runner()
     runner.config_file.remove()
     os.chmod(str(runner.xdg_config_home), 555)
     result = runner.invoke(main_khal, ['configure'], input=choices())
     assert result.exit_code == 1
+    # make sure pytest can clean up behind us
+    cleanup([runner.xdg_config_home, runner.xdg_data_home])
```
d7415, line 782: Do we need to cleanup `xdg_data_home`? We didn't change it in this test.

tests/cli_test.py
```diff
@@ -779,6 +792,8 @@ def test_configure_command_cannot_create_vdir(runner):
     )
     assert 'Exiting' in result.output
     assert result.exit_code == 1
+    # make sure pytest can clean up behind us
+    cleanup([runner.xdg_config_home, runner.xdg_data_home])
```
d7415, line 796: Do we need to cleanup `xdg_config_home`? We didn't change it in this test.

d7415 (changes requested):
A bit permissive (600/700 should be sufficient unless I'm missing something), but this should only run against temporary test directories anyway so it doesn't really matter.

A few small queries, but in principle it looks good.

tests/cli_test.py
geier: I don't think so. As os.walk walks down all directories, we catch all directories eventually with line 770.

tests/cli_test.py
geier: removed

tests/cli_test.py
geier: removed

geier:
Changed te permissions 
back to 755 and 644, those seem to be the default ones on the directories created by pyetst.

tests/cli_test.py
d7415: Ah, yes, re-reading it I think you're right.

d7415 (approved):

PR #1024, opened 2021-03-12
Allow darwin locale in calendar tests

geier:
Darwin (Mac OS) has some other locale settings for French and Greek than Linux. This path lets the tests pass on Darwin as well.


d7415 (approved):
I don't have Darwin to test, but it looks ok to me.

PR #1026, opened 2021-03-13
Running flake8 and mypy on github actions.

geier:
This should only run flake8 and mypy. One test is still failing on github actions, I cannot reproduce this locally for the moment.

Template taken from https://hynek.me/articles/python-github-actions/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,36 @@
+---
+name: CI
+
+on:
+  push:
+    branches: ["master"]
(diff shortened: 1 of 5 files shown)


geier:
Yes, there is some superfluous stuff in the workflow file. We could take it out, but we probably want to use it once the tests are running on github actions as well.

d7415 (approved):
Looks ok to me!

WhyNotHugo (approved):

PR #1030, opened 2021-04-08
Use new syntax for validate. Fixes #993.

### LINKED ISSUE #993: Migrate to new configobj validate
[`configobj` 5.1.0+](https://github.com/DiffSK/configobj/blob/master/CHANGES.rst#release-510) has integrated `validate` and will eventually want a new syntax. 5.1.0 has a shim to allow the old syntax to work for now. Until we set a minimum version of `configobj` we'll have to cover both imports.

d7415:
configobj 5.1.0+ has integrated validate and will eventually want a new syntax. 5.1.0 has a shim to allow the old syntax to work for now. Until we set a minimum version of configobj we should cover both imports.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -12,10 +12,15 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import validate
 from configobj import ConfigObj
 
 import khal
+
+try:
+    # Available from configobj 5.1.0
+    import configobj.validate
+except ModuleNotFoundError:
+    import validate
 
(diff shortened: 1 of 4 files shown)


d7415:
There's no rush for this (especially as [5.1.0 hasn't been released yet](https://github.com/DiffSK/configobj/issues/204)) but it was easy enough to implement. It will also be something else to remove when we do set a minimum version (and 5.1.0 has been released).

geier:
I manually tested this with configobj's master branch, works!

Thanks at @d7415!

geier (approved):

PR #1029, opened 2021-04-02
Add an error message if a calendar specified with 'type = discover' is not a directory

d7415:
Closes #991

PR #1031, opened 2021-04-27
Drop support for Python 3.5

d7415:
I should have caught this earlier (before the release), but #1029 apparently broke py3.5 support. I guess it slipped through while we didn't have Travis.

I suspect we could work around it, but as py3.5 has been EOL for 7 months (since 2020-09-30) I propose we just stop supporting it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 os: linux
 
 python:
-    - 3.5
     - 3.6
     - 3.7
     - 3.8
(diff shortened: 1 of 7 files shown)


WhyNotHugo (approved):
LGTM. This should allow making use of other 3.6+ features freely.

geier (approved):

geier:
Looks good to me.

Hooray for f-strings!

PR #1032, opened 2021-04-28
Upgrade syntax for Python 3.6

WhyNotHugo:
This change updates lots of syntax dropping anything that was necessary for Python < 3.6.

Honestly, I just ran:

    pyupgrade --py36-plus $(find . -name "*.py")

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -711,10 +711,9 @@ class LocalizedEvent(DatetimeEvent):
             starttz = getattr(self._vevents[self.ref]['DTSTART'].dt, 'tzinfo', None)
         except KeyError:
             msg = (
-                "Cannot understand event {} from "
-                "calendar {}, you might want to file an issue at "
+                f"Cannot understand event {kwargs.get('href')} from "
+                f"calendar {kwargs.get('calendar')}, you might want to file an issue at "
                 "https://github.com/pimutils/khal/issues"
-                .format(kwargs.get('href'), kwargs.get('calendar'))
             )
             logger.fatal(msg)
             raise FatalError(  # because in ikhal you won't see the logger's output
@@ -777,11 +776,11 @@ class AllDayEvent(Event):
         end = super().end
         if end == self.start:
             # https://github.com/pimutils/khal/issues/129
-            logger.warning('{} ("{}"): The event\'s end date property '
-                           'contains the same value as the start date, '
-                           'which is invalid as per RFC 5545. Khal will '
-                           'assume this is meant to be a single-day event '
-                           'on {}'.format(self.href, self.summary, self.start))
+            logger.warning(f'{self.href} ("{self.summary}"): The event\'s end '
+                           'date property contains the same value as the start '
+                           'date, which is invalid as per RFC 5545. Khal will '
+                           'assume this is meant to be a single-day event on '
+                           f'{self.start}')
             end += dt.timedelta(days=1)
         return end - dt.timedelta(days=1)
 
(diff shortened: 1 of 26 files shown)


khal/khalendar/vdir.py
```diff
@@ -176,7 +176,7 @@ def create(cls, collection_name, **kwargs):
         if not os.path.exists(path):
             os.makedirs(path, mode=cls.default_mode)
         elif not os.path.isdir(path):
-            raise IOError('{} is not a directory.'.format(repr(path)))
+            raise OSError('{} is not a directory.'.format(repr(path)))
```
d7415, line 179:
```suggestion
            raise OSError(f'{repr(path)} is not a directory.')
```

khal/khalendar/khalendar.py
```diff
@@ -341,7 +341,7 @@ def _update_vevent(self, href: str, calendar: str) -> bool:
             if not isinstance(e, (UpdateFailed, UnsupportedFeatureError, NonUniqueUID)):
                 logger.exception('Unknown exception happened.')
             logger.warning(
-                'Skipping {0}/{1}: {2}\n'
+                'Skipping {}/{}: {}\n'
                 'This event will not be available in khal.'.format(calendar, href, str(e)))
```
d7415, line 345:
```suggestion
                f'Skipping {calendar}/{href}: {str(e)}\n'
                'This event will not be available in khal.')
```

d7415 (commented):
I think I learned something from each category of changes here!

My proposed changes are essentially "use more f-strings" for consistency if nothing else. I haven't tested them because I did all of them from GitHub. That may have been an error, but it's done now.

I only changed the ones that were already in this patch, so there might be some more available.

d7415:
> Honestly, I just ran:
> 
> ```
> pyupgrade --py36-plus $(find . -name "*.py")
> ```

Something else I'll have to try to remember!

khal/khalendar/event.py
```diff
@@ -774,7 +774,7 @@ class AllDayEvent(Event):
 
     @property
     def end(self):
-        end = super(AllDayEvent, self).end
+        end = super().end
         if end == self.start:
             # https://github.com/pimutils/khal/issues/129
             logger.warning('{} ("{}"): The event\'s end date property '
```
geier, line 780: what about this one? .format methods in log strings are executed anyway, I dont think it makes much difference between using f-strings and .format.

d7415: I didn't want to look until those changes were merged, but it looks like there are a lot (~90 on a quick grep) of instances of `'{}'.format(x)` that we could change if we want to go the whole hog. It would make it more consistent... I'll try to make a start in a bit. Also fixing style.

d7415: I think that's all of them. If they seem excessive feel free to revert them!

WhyNotHugo: I think these ones are missing because of how the string is split across multiple lines.

Those might have to be looked at manually -- unless you're feeling brave enough to poke into `pyupgrade`.

WhyNotHugo: Re style, I'd just add a pre-commit hook with `black` and `reorder_python_imports`, and forget about dealing with that manually.

d7415: > Those might have to be looked at manually -- unless you're feeling brave enough to poke into `pyupgrade`.

I did a `git grep {}` then looked at them manually. The style fixes were only because I was using GitHub suggestions rather than editing locally :)

geier (approved):
While I like those changes, I find these huge, automatically created PRs quite a pain to review (for rather little gain).

Issue #1025, opened 2021-03-12
khal list output format difficult for other programs to parse


tubaman:
khal [output format options](https://khal.readthedocs.io/en/latest/usage.html#cmdoption-format) are fantastic for humans(colorized, etc)!  However, I feel like I'm jumping through hoops to make khal's list output parsable by other programs.  For example, if I want another program to parse a list events with title, start and description, currently I'm generating a unique field separator that I can use later for splitting:

(in python)
```python
rsep = str(uuid.uuid4())
fsep = str(uuid.uuid4())
fmt = "{title}%s{start-long}%s{description}%s" % (fsep, fsep, rsep)
cmd = ['khal', 'list', '--notstarted', '--day-format', '', '--format', fmt, start, end]
```

Perhaps it would be nice to add a `--json` option to format khal's list output as a JSON array of objects?

geier:
should be fixed with #1308

PR #1037, opened 2021-05-09
Add json output option

PsychicNoodles:
As pointed out in #1025, trying to use output from khal in other programs is difficult with plain string formatting, requiring awkward delimiters in the format string. This PR adds a `--json` option to the at, list, search, and new commands that allows desired JSON fields to be specified. The option uses click's `multiple` feature, like the include and exclude calendar options. Currently it accepts any option available to the default formatter, and all events for a single day are collected into a list (so multi-day queries will have multiple JSON lists consecutively).

There are a couple tweaks that I think were not necessary to implement but could be changed:

* limit to fields that actually contain data (exclude things like color codes)
* combine lists of multiple days


WhyNotHugo:
I actually really like this feature and can see myself rewriting some horrible wrapper scripts that I have.

What do you think of just printing _all_ fields for events, instead of having to specify fields individually? It does seem a bit simpler for end-users, do you think code-wise it would be a simpler approach too?

PsychicNoodles:
Sure, that shouldn't be too hard to add. The only thing is that since you probably don't want, for example, ansi color codes included in the JSON objects there needs to be a filter for valid fields. That isn't a big deal at all to add, but in the future if the list of attributes changes then the filter will need updating as well.

PsychicNoodles:
I added an option so that if the only item given to `--json` is `all` then all valid fields will be included in the output. Depending on your usecase this is probably good enough; unless you have a huge collection of events to process all at once, the output shouldn't get too big even with all fields enabled. That said, I could see adding a special `default` option as well that goes from the config and limits to a specified list of fields, like in `agenda_event_format` and `agenda_day_format`, but I'm not sure if that's necessary at this point.

tests/cli_test.py
```diff
@@ -865,7 +952,7 @@ def test_debug(runner):
     assert not result.exception
 
 
-@freeze_time('2015-6-1 8:00')
+@ freeze_time('2015-6-1 8:00')
```
geier, line 955: small typo

khal/khalendar/event.py
```diff
@@ -656,7 +655,7 @@ def format(self, format_string, relative_to, env={}, colors=True):
 
         attributes['status'] = self.status + ' ' if self.status else ''
         attributes['cancelled'] = 'CANCELLED ' if self.status == 'CANCELLED' else ''
-        return format_string.format(**dict(attributes)) + attributes["reset"]
+        return dict(attributes)
```
geier, line 658: do we need the `dict()` here?

khal/khalendar/event.py
```diff
@@ -692,7 +691,7 @@ def delete_instance(self, instance):
         for key in to_pop:
             self._vevents.pop(key)
 
-    @property
+    @ property
```
geier, line 694: typo

khal/controllers.py
```diff
@@ -280,7 +284,7 @@ def khal_list(collection, daterange=None, conf=None, agenda_format=None,
 
 def new_interactive(collection, calendar_name, conf, info, location=None,
                     categories=None, repeat=None, until=None, alarms=None,
-                    format=None, env=None, url=None):
+                    format=None, json=[], env=None, url=None):
```
geier, line 287: see above

khal/controllers.py
```diff
@@ -344,7 +348,7 @@ def new_interactive(collection, calendar_name, conf, info, location=None,
 
 def new_from_string(collection, calendar_name, conf, info, location=None,
                     categories=None, repeat=None, until=None, alarms=None,
-                    url=None, format=None, env=None):
+                    url=None, format=None, json=[], env=None):
```
geier, line 351: see above

geier (commented):
@PsychicNoodles Thank you for your work and the PR. I haven't really tested anything or had a closer look yet, but in general that approach seems fine to me.

I commented some minor code style issues, which would be great if you could fix them.

Also, I believe we need some more documentation.

The `all` value to `--json` should definitively get mentioned, and probably also a reference to all the allowed fields. Perhaps we could even include an example of how the new --json option works.

khal/controllers.py
```diff
@@ -195,29 +195,33 @@ def get_events_between(
...
             seen.add(event.uid)
 
-    return event_list
+    return formatter(event_list)
 
 
 def khal_list(collection, daterange=None, conf=None, agenda_format=None,
               day_format=None, once=False, notstarted=False, width=False,
-              env=None, datepoint=None):
+              env=None, datepoint=None, json=[]):
```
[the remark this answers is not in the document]
PsychicNoodles: Ooh, didn't know that was a thing in Python

PsychicNoodles:
I noticed when implementing that there are a few template options that are missing from the documentation, so since I was listing them in the json documentation anyway I went ahead and added them.

PsychicNoodles:
Yeah, I noticed that too. I see two ways to solve this, which makes more sense to you?

* (as you suggested) output is an object with keys for each day and corresponding lists of events
* one list combining all events, if the user wants to differentiate they can use `start-date` and/or `end-date` options

I feel like your suggestion is more consistent with the normal, human readable output, but it's maybe unnecessary since if the user needs date information and uses `--json start-date` then it'll be included for every result anyway.

geier:
@PsychicNoodles  I don't really care and would wait for some people seeing themselves using json output to pitch in. With one list I am wondering if it should be deduped, or if it needs a new field "relative-to-date".

PsychicNoodles:
> With one list I am wondering if it should be deduped, or if it needs a new field "relative-to-date".

Ahh, that's a good point, if the dates are getting rolled together then maybe it could just be one entry? But that would be confusing if you aren't expecting a multi-day event and add a `start-date` or `end-date`.

If you'd rather wait for people to use it before making further decisions then is there anything else you think needs to be added?

PsychicNoodles:
@geier 
Is there anything else that you think needs to be done for this PR?

WhyNotHugo (approved):
This LGTM.

I do think that it might be better/simpler to just print _all_ fields rather than expect a list of fields to print. But I won't holding merging this based on that.

PsychicNoodles:
> I do think that it might be better/simpler to just print all fields rather than expect a list of fields to print.
Honestly, I would be fine either way, I just wanted it to be a bit easier to use outside of just scripts since the full output is quite a lot to sift through.

khal/utils.py
```diff
@@ -183,3 +186,74 @@ def relative_timedelta_str(day: dt.date) -> str:
...
+            if f.get('repeat-symbol', '') != '':
+                f["repeat-symbol"] = f["repeat-symbol"].strip()
+            if f.get('status', '') != '':
+                f["status"] = f["status"].strip()
+            if f.get('cancelled', '') != '':
+                f["cancelled"] = f["cancelled"].strip()
+
+            filtered.append(f)
+
+        results = [json.dumps(filtered, ensure_ascii=False)]
```
WhyNotHugo, line 253: **non-blocking suggestion**:

Maybe format this with whitespace to make it a bit more human-readable too?

```suggestion
        results = [json.dumps(filtered, indent=2, ensure_ascii=False)]
```

WhyNotHugo (commented):
I've just one reservation here: why print on array per day, instead of a single array with all the events?

I'm asking mostly because having multiple arrays is a bit trickier to parse: you can't just feed the entire output to a `json` parser, but need to split lines and parse them individually.

WhyNotHugo:
CI failure is unrelated here. Fixed in https://github.com/pimutils/khal/pull/1067

geier:
I accidentally closed this PR, please re-open if you can.

geier:
see https://github.com/pimutils/khal/pull/1146 for now

PR #1038, opened 2021-05-11
Add bugbear flake8 plugin

WhyNotHugo:
Bugbear detects possible code issues. Like having mutable default
arguments in functions, or unnecessary calls to `list`, `set` and `dict`.

I've also moved the `flake8` ignores into the proper `tox` section. The
end result is the same when running `tox -e style`, but this change
allows code editors to pick up these settings too.

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -121,7 +121,7 @@ class Event(object):
         """
         assert isinstance(events_list, list)
 
-        vevents = dict()
+        vevents = {}
         for event in events_list:
             if 'RECURRENCE-ID' in event:
                 if invalid_timezone(event['RECURRENCE-ID']):
@@ -262,21 +262,21 @@ class Event(object):
     @property
     def symbol_strings(self):
         if self._locale['unicode_symbols']:
-            return dict(
-                recurring='\N{Clockwise gapped circle arrow}',
-                range='\N{Left right arrow}',
-                range_end='\N{Rightwards arrow to bar}',
-                range_start='\N{Rightwards arrow from bar}',
-                right_arrow='\N{Rightwards arrow}'
-            )
+            return {
+                "recurring": '\N{Clockwise gapped circle arrow}',
+                "range": '\N{Left right arrow}',
+                "range_end": '\N{Rightwards arrow to bar}',
+                "range_start": '\N{Rightwards arrow from bar}',
+                "right_arrow": '\N{Rightwards arrow}'
+            }
         else:
-            return dict(
-                recurring='(R)',
-                range='<->',
-                range_end='->|',
-                range_start='|->',
-                right_arrow='->'
-            )
+            return {
+                "recurring": '(R)',
+                "range": '<->',
+                "range_end": '->|',
+                "range_start": '|->',
+                "right_arrow": '->'
+            }
 
     @property
     def start_local(self):
@@ -356,7 +356,7 @@ class Event(object):
         return text
         """
         calendar = self._create_calendar()
-        tzs = list()
+        tzs = []
         for vevent in self._vevents.values():
             if hasattr(vevent['DTSTART'].dt, 'tzinfo') and vevent['DTSTART'].dt.tzinfo is not None:
                 tzs.append(vevent['DTSTART'].dt.tzinfo)
@@ -497,12 +497,14 @@ class Event(object):
             recurstr = ''
         return recurstr
 
-    def format(self, format_string, relative_to, env={}, colors=True):
+    def format(self, format_string, relative_to, env=None, colors=True):
         """
         :param colors: determines if colors codes should be printed or not
         :type colors: bool
         """
-        attributes = dict()
+        env = env or {}
+
+        attributes = {}
         try:
             relative_to_start, relative_to_end = relative_to
         except TypeError:
@@ -680,7 +682,7 @@ class Event(object):
 
         # in case the instance we want to delete is specified as a RECURRENCE-ID
         # event, we should delete that as well
-        to_pop = list()
+        to_pop = []
         for key in self._vevents:
             if key == 'PROTO':
                 continue
@@ -853,7 +855,7 @@ def create_timezone(tz, first_date=None, last_date=None):
             last_num = num
             last_tt = transtime
 
-    timezones = dict()
+    timezones = {}
     for num in range(first_num, last_num + 1):
         name = tz._transition_info[num][2]
         if name in timezones:
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -263,19 +263,19 @@ class Event(object):
     def symbol_strings(self):
         if self._locale['unicode_symbols']:
             return {
-                "recurring": '\N{Clockwise gapped circle arrow}',
-                "range": '\N{Left right arrow}',
-                "range_end": '\N{Rightwards arrow to bar}',
-                "range_start": '\N{Rightwards arrow from bar}',
-                "right_arrow": '\N{Rightwards arrow}'
+                'recurring': '\N{Clockwise gapped circle arrow}',
+                'range': '\N{Left right arrow}',
+                'range_end': '\N{Rightwards arrow to bar}',
+                'range_start': '\N{Rightwards arrow from bar}',
+                'right_arrow': '\N{Rightwards arrow}'
             }
         else:
             return {
-                "recurring": '(R)',
-                "range": '<->',
-                "range_end": '->|',
-                "range_start": '|->',
-                "right_arrow": '->'
+                'recurring': '(R)',
+                'range': '<->',
+                'range_end': '->|',
+                'range_start': '|->',
+                'right_arrow': '->'
             }
 
     @property
--- a/khal/khalendar/khalendar.py
+++ b/khal/khalendar/khalendar.py
@@ -67,14 +67,15 @@ class CalendarCollection(object):
                  color: str='',
                  priority: int=10,
                  highlight_event_days: bool=False,
-                 locale: Dict[str, Any]=dict(),
+                 locale: Dict[str, Any]=None,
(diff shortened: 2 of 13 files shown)


tests/cli_test.py
```diff
@@ -320,13 +320,13 @@ def test_invalid_calendar(runner):
 def test_attach_calendar(runner):
     runner = runner(days=2)
     result = runner.invoke(main_khal, ['printcalendars'])
-    assert set(result.output.split('\n')[:3]) == set(['one', 'two', 'three'])
+    assert set(result.output.split('\n')[:3]) == {['one', 'two', 'three']}
```
d7415, line 323: Looks like this one's broken.
```suggestion
    assert set(result.output.split('\n')[:3]) == {'one', 'two', 'three'}
```

tests/cli_test.py
```diff
@@ -320,13 +320,13 @@ def test_invalid_calendar(runner):
...
     assert not result.exception
     result = runner.invoke(main_khal, ['printcalendars', '-a', 'one'])
     assert result.output == 'one\n'
     assert not result.exception
     result = runner.invoke(main_khal, ['printcalendars', '-d', 'one'])
-    assert set(result.output.split('\n')[:2]) == set(['two', 'three'])
+    assert set(result.output.split('\n')[:2]) == {['two', 'three']}
```
d7415, line 329: Same
```suggestion
    assert set(result.output.split('\n')[:2]) == {'two', 'three'}
```

tests/khalendar_test.py
```diff
@@ -107,10 +107,10 @@ def test_list(self, coll_vdirs):
         event = Event.fromString(event_today, calendar=cal1, locale=LOCALE_BERLIN)
         coll.new(event)
         hrefs = sorted(href for href, etag in coll._backend.list(cal1))
-        assert set(str(coll.get_event(href, calendar=cal1).uid) for href in hrefs) == set((
+        assert {str(coll.get_event(href, calendar=cal1).uid) for href in hrefs} == {(
```
d7415, line 110: This one works but changes the meaning.
```suggestion
        assert {str(coll.get_event(href, calendar=cal1).uid) for href in hrefs} == {
```

d7415 (changes requested):
Makes sense and mostly looks good. Looks like it's made a few errors - It can't be expected to understand the code.

khal/khalendar/event.py
```diff
@@ -262,21 +262,21 @@ def increment_sequence(self):
...
-                range_start='\N{Rightwards arrow from bar}',
-                right_arrow='\N{Rightwards arrow}'
-            )
+            return {
+                "recurring": '\N{Clockwise gapped circle arrow}',
+                "range": '\N{Left right arrow}',
+                "range_end": '\N{Rightwards arrow to bar}',
+                "range_start": '\N{Rightwards arrow from bar}',
+                "right_arrow": '\N{Rightwards arrow}'
+            }
```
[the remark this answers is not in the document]
WhyNotHugo: Agreed. Is there any preference so far? I usually stick with `"`, since that's what `black` uses.

khal/ui/calendarwidget.py
```diff
@@ -499,7 +499,7 @@ def _construct_week(self, week):
             attr = None
 
         this_week = [(get_month_abbr_len(), urwid.AttrMap(urwid.Text(month_name), attr))]
-        for number, day in enumerate(week):
+        for _number, day in enumerate(week):
```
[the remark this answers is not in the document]
WhyNotHugo: That's correct. Either leading `_`, or just called `_`.

tests/cli_test.py
WhyNotHugo: Thanks! I did these manually.

tests/khalendar_test.py
WhyNotHugo: It does?

tests/khalendar_test.py
```diff
@@ -298,7 +298,7 @@ def test_search_recurrence_id_only_multi(self, coll_vdirs):
         event = Event.fromString(
             _get_text('event_dt_multi_recuid_no_master'), calendar=cal1, locale=LOCALE_BERLIN)
         coll.new(event, cal1)
-        events = list(sorted(coll.search('Event')))
+        events = list(coll.search('Event'))
```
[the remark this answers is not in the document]
WhyNotHugo: Good catch, I had to drop the `list` here, not `sorted`.

tests/khalendar_test.py
d7415:
```py
>>> set(('uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU'))
{'uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU'}
>>> {('uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU')}
{('uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU')}
>>> {'uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU'}
{'uid3@host1.com', 'V042MJ8B3SJNFXQOJL6P53OFMHJE8Z3VZWOU'}
```

d7415: The first one looks fine, just the one on the following lines (that starts here)

khal/khalendar/event.py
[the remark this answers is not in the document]
d7415: It seems to be pretty mixed at the moment. I usually lean to `"` unless I have a reason (surrounding code, string with a `"` in it). I think I usually use `'` for dict keys too, but I couldn't say why!

tests/khalendar_test.py
```diff
@@ -334,7 +334,7 @@ def test_invalid_timezones(self, coll_vdirs):
         event = Event.fromString(
             _get_text('invalid_tzoffset'), calendar=cal1, locale=LOCALE_BERLIN)
         coll.new(event, cal1)
-        events = list(sorted(coll.search('Event')))
+        events = list(coll.search('Event'))
```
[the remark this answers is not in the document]
WhyNotHugo: Thanks for pointing these last two out, I had to replace `list(sorted(` with `sorted(`, but did the opposite! :facepalm:

tests/khalendar_test.py
WhyNotHugo: Thanks for finding this one too!

WhyNotHugo:
Feel free to squash-merge since the extra commits don't add much value to git-history.

khal/khalendar/khalendar.py
```diff
@@ -67,14 +67,15 @@ def __init__(self,
                  color: str='',
                  priority: int=10,
                  highlight_event_days: bool=False,
-                 locale: Dict[str, Any]=dict(),
+                 locale: Dict[str, Any]=None,
```
geier, line 70: this needs an Optional around it

khal/utils.py
```diff
@@ -50,17 +50,15 @@ def generate_random_uid():
 
 
 def find_last_reset(string):
-    for match in re.finditer(ansi_reset, string):
-        pass
+    *_, match = re.finditer(ansi_reset, string)
```
geier, line 53: this will definitively break

khal/utils.py
```diff
@@ -50,17 +50,15 @@ def generate_random_uid():
...
     try:
         return match.start(), match.end(), match.group(0)
     except UnboundLocalError:
         return -2, -1, ''
 
 
 def find_last_sgr(string):
-    for match in re.finditer(ansi_sgr, string):
-        pass
+    *_, match = re.finditer(ansi_sgr, string)
```
geier, line 61: same here

geier (changes requested):
Personally I prefer `dict()` to `{}`, but it's not a strong preference. Otherwise there are some minor issues that still need to be resolved in here.

khal/utils.py
d7415: It should make `match` the last result of the iterator, like the original. Unless I'm missing something. I did a quick test when I was reviewing, but I didn't check against `re` specifically.

geier: but I believe for that two work, we need at least two matches, or not?

d7415: The syntax seems to work for iterators with length 1. I'll double check `re` for good measure.

d7415: Looks good.

geier: You are right, it totally works with iterators of length 1, learned something new, Thanks!

Unfortunately, it doesn't seem to work if there is no match at all, [some tests are failing therefore](https://travis-ci.com/github/pimutils/khal/jobs/504997519#L293).

d7415: So I guess the options are:
- Revert these two changes
- Move the `try` to the `finditer` line and change the exception accordingly
- Replace the `try`/`except` with an `if match`...`else`

Thoughts?

khal/khalendar/khalendar.py
d7415: Added.

d7415 (approved):
I did a quick commit for @geier's last comment (which was a pre-existing issue), so I think this should be good to merge.

geier (approved):

geier:
Awesome, thanks!

geier:
> I did a quick commit for @geier's last comment (which was a pre-existing issue), so I think this should be good to merge.

I think I'd like to add some more types in general. What's your opinion on type hinting?

d7415:
> What's your opinion on type hinting?

I'm not used to it, but I like it in principle.

WhyNotHugo:
I think it helps quite a bit in figuring out simple errors. But gets in the way for complex projects started without it.

PR #1040, opened 2021-05-11
ikhal: tab moves focus from events to calendar

### LINKED ISSUE #1015: Tab toggle focus
Couldn't find if this had been brought up before but since `<tab>` takes you from the calender to the event list in interactive mode it would seem natural to me if another `<tab>` would bring you back again.

geier:
This commit enables tab (and shift tab) to jump from the right side
(events) to the left (calendar).  With that, we lose the ability to tab
through multiple events on one calendar day.

Tabbing through events was broken beforehand, as jumping from the first to last (or backwards) tabbing didn't work anyway.

fix #1015

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ may want to subscribe to `GitHub's tag feed
 not released
 
 * DROPPED support for Python 3.5
+* CHANGE: ikhal: tab (and shift tab) jump from the events back to the calendar
 
 0.10.3
 ======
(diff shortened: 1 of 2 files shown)


d7415 (approved):
> This commit enables shift (and shift tab)

I assume you mean tab and shift tab?

I guess this might be worth mentioning in the CHANGELOG too. Other than that it seems like a good idea, the code change is straightforward and it works on a quick test :+1:

geier:
@d7415 Thanks for the feedback, should be fixed.

geier:
and yes, automated tests would be nice, but it's rather tedious.

Issue #906, opened 2020-02-05
Uneven tab application per calendar in `khal calendar`


7596ff:
I am using the `{tab}` character in my `agenda_event_format`: `"{calendar-color}{cancelled}{start-end-time-style}{tab}{title}{repeat-symbol} {reset}{calendar}{reset}"`, and here is my output. 
```
[cassie@houndoom ~]$ khal calendar -a Calendar -a Classes
    Su Mo Tu We Th Fr Sa     [2020-02-05] Today
Feb 26 27 28 29 30 31  1     |->  Vivian Visits Calendar
     2  3  4  5  6  7  8     10:00-10:50    PHL-1100-102 LOGIC I (R) Classes
     9 10 11 12 13 14 15     11:00-11:50    RC-2001-120 INTR WRITING ACROSS CURRICULUM (R)
    16 17 18 19 20 21 22     Classes
    23 24 25 26 27 28 29     14:00-14:50    AST-1002-102 INTRO AST II/STAR GALAX (R) Classes
Mar  1  2  3  4  5  6  7     15:30-16:45    PHL-2000-104 PHILOSOPHY, SOCIETY & ETHICS (R)
     8  9 10 11 12 13 14     Classes
    15 16 17 18 19 20 21     17:00-18:00  Pay Ticket Deadline Calendar
    22 23 24 25 26 27 28     22:00-23:50    AST-1002-206 Laboratory (R) Classes
Apr 29 30 31  1  2  3  4     [2020-02-06] Tomorrow
     5  6  7  8  9 10 11     ->|  Vivian Visits Calendar
    12 13 14 15 16 17 18     09:30-10:45    PHL-1501-101 MIND, KNOWLEDGE & REALITY (R) Classes
    19 20 21 22 23 24 25     11:00-14:00  Rivers Street (noodles) (R) Calendar
May 26 27 28 29 30  1  2
```
This is easier to read in a screenshot, [posted here](https://khal.is-pretty.cool/AYxep2j.png).

While the duration formats are the same length on different events, it seems that depending on the calendar there are different tab lengths applied to each event. 

Despite the intended function of the tab control code, I would love to be able to format the list in some sort of column-based output, so that I could have a column each for time, summary, and calendar name. I await the discussion in this issue.

7596ff:
With more calendars in view, there seems to be proceedingly more extra tab characters, in a 1-for-1 relationship.

d7415:
[...]

PR #1041, opened 2021-05-12
Additional test for color_wrap()

geier:
This is an attempt to reproduce #906


geier:
This actually fixes a bug, that somehow never occurred before.

d7415 (approved):

PR #1043, opened 2021-05-17
add alarm symbol for events with at least one alarm

### LINKED ISSUE #674: Show alarm symbol (⏰) in khal
I like that `khal calendar`/`ikhal` shows recurring events with a `⟳` symbol and propose do have a symbol that marks events with any alarm set with a `⏰` (or some similar) symbol.

sdx23:
fixes #674

Would be glad about some feedback on how you want this to be tested (if at all). Note that some tests are failing, but those are the same that have been failing before as well (platform linux -- Python 3.7.3, pytest-3.10.1, py-1.7.0, pluggy-0.8.0).

Edit: note that I chose the notes symbol because that was supported by my default font (Bitstream vera sans mono) in contrast to the other suggestions. They seem to be newer and possibly less supported.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ may want to subscribe to `GitHub's tag feed
 not released
 
 * DROPPED support for Python 3.5
+* NEW Add symbol for events with at least one alarm
 
 0.10.3
 ======
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -112,6 +112,9 @@ Several options are common to almost all of :program:`khal`'s commands
    repeat-symbol
         A repeating symbol (loop arrow) if the event is repeating.
 
+   alarm-symbol
+        An alarm symbol (alarm clock) if the event has at least one alarm.
+
    location
         The event location.
 
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -264,6 +264,7 @@ class Event:
         if self._locale['unicode_symbols']:
             return dict(
                 recurring='\N{Clockwise gapped circle arrow}',
+                alarming='\N{Beamed eighth notes}',
                 range='\N{Left right arrow}',
                 range_end='\N{Rightwards arrow to bar}',
                 range_start='\N{Rightwards arrow from bar}',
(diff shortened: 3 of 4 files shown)


WhyNotHugo (commented):
Does `⏰` not work for you?

PR looks good to me. I won't hold back based on the default icon decision, but I do think it should be customizable.

sdx23:
No, sadly ⏰ doesn't work for me. However, you may well change the icon as you wish.

d7415 (commented):
My main reservation was also the symbol used. FWIW as part of my testing I have pushed a commit with the ⏰ symbol (and deconflicted with latest master) to the `sdx23_pr1043` branch.

WhyNotHugo (approved):

WhyNotHugo:
Merged. Thanks!

d7415:
> it's configurable

Is it..?

sdx23:
Sorry for not following up on this one. After the `sdx23_pr1043` branch was introduced I'd have expected that one to be merged, not the original PR.

> Is it..?

Yes and no. It is configurable **whether** a symbol is shown but not **which**.

WhyNotHugo:
Oh, my bad. I must have mixed up the branch on this PR and `sdx23_pr1043` (I think I read one, but merged the other).

Do you want to open a PR with the additional changes in that one?

d7415:
I think the only main difference was the symbol. It looks like it's also reverted one of the formatting fixes though. I'll go fix that part now. Symbol is up to you - I prefer the clock (usage also says `An alarm symbol (alarm clock)`) but it's not a big deal either way.

WhyNotHugo:
I'd rather have the icon be configurable TBH. If you want to open a PR changing it though, I'm not opposed.

Again, I [mis]merged since I mixed up the branches. :sweat:

sdx23:
There you go: https://github.com/sdx23/khal/commit/6b721df950db7f068c9fed085ea956833850a884

I have no glue how to add that commit to this pr or whether that's possible at all.

Personally I'm in favor of just changing it to '\N{Alarm clock}'. The other symbols are not configurable as well.

PR #1048, opened 2021-05-21
add default alarm configuration option

### LINKED ISSUE #698: Default alarm
It would be nice if khal could default to setting an alarm, say, a configurable number of minutes before a newly-created event.

sdx23:
fixes #698 

Not sure where I put it is the best place to, please comment on that.

Also: I think in the existing options "default_(day)event_duration" the terms "dayevent" and "event" are mixed up. I'd expect "dayevent" to refer to an day-long event and in contrast to that "event" an event of e.g. 1h duration, but the implementation is the other way round. My commit, nevertheless, uses this same (possibly reverse) meaning.


WhyNotHugo:
I agree that the naming looks backwards there. I may be missing context though.

geier:
Thanks and sorry for the 2.5 year delay in taking a look at this.

I have rebased this on the latest master. I also think it's a good idea to have this in ikhal and while it's simple to add a default alarm for a new event, I'm not yet sure on how to convert the alarm if one changes the event from daytime to allday. Perhaps check if the alarm is equal to the default and change it then to the other default?

geier:
> Also: I think in the existing options "default_(day)event_duration" the terms "dayevent" and "event" are mixed up. I'd expect "dayevent" to refer to an day-long event and in contrast to that "event" an event of e.g. 1h duration, but the implementation is the other way round. My commit, nevertheless, uses this same (possibly reverse) meaning.

Good catch and I have no idea how that could be that wrong for so many years. Hopefully fixed.

geier:
> I'm not yet sure on how to convert the alarm if one changes the event from daytime to allday. Perhaps check if the alarm is equal to the default and change it then to the other default?

implemented that way.

octvs:
I was just looking for this option on `man khal`, then ended up here. Was there
a reason why it didn't get documented?

Issue #1050, opened 2021-05-25
edit (especially summary and description) could benefit from readline


allanwind:
The current workflow is to either discard the original content and write what you want, or to copy and paste the text with your mouse so you can edit it.  It would be a more friendly workflow to present the old edit and support readline or similar to allow easy edit.

PR #1052, opened 2021-05-26
Add --every / -e option to the new command

### LINKED ISSUE #999: custom recurrence support type: enhancement
When booking an event with **new** command it would be useful to define custom recurrence; Example - bi-weekly events are quite prevalent. I currently book these kind of events via a python script; But don't have the benefit of managing these events as a single entity.

Existing functionality:
-r, –repeat=RRULE specify if and how this event should be recurring. Valid values for RRULE are daily, weekly, monthly and yearly
-u, –until=UNTIL specify until when a recurring event should run

Add
-e,  -every=FREQUENCY specify 'repeat every' frequency for a recurring event; Ex: every two weeks (bi-weekly).

Usage:
khal new -a work 26.07. Great Event -g meeting -r weekly -e 2

allanwind:
This includes support for interactive setting or editing an event.  Fixes #999

### DIFF
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -412,6 +412,10 @@ Options
 
 * **-u, --until=UNTIL** specify until when a recurring event should run
 
+* **-e, --every=EVERY** specify at what interval a reccurint event
+  should run. For instance, if RRULE is weekly and EVERY is 2, it
+  woulld schedule the event every 2 weeks.
+
 * **--url** specify the URL element of the event
 
 * **--alarms DURATION,...** will add alarm times as DELTAs comma separated for this event,
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -412,7 +412,7 @@ Options
 
 * **-u, --until=UNTIL** specify until when a recurring event should run
 
-* **-e, --every=EVERY** specify at what interval a reccurint event
+* **-e, --every=EVERY** specify at what interval a recurring event
   should run. For instance, if RRULE is weekly and EVERY is 2, it
   would schedule the event every 2 weeks.
 
--- a/doc/source/usage.rst
+++ b/doc/source/usage.rst
@@ -414,7 +414,7 @@ Options
 
 * **-e, --every=EVERY** specify at what interval a reccurint event
   should run. For instance, if RRULE is weekly and EVERY is 2, it
-  woulld schedule the event every 2 weeks.
+  would schedule the event every 2 weeks.
 
 * **--url** specify the URL element of the event
 
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -341,6 +341,9 @@ def _get_cli():
                   help=('Repeat event: daily, weekly, monthly or yearly.'))
     @click.option('--until', '-u',
                   help=('Stop an event repeating on this date.'))
+    @click.option('--every', '-e',
+                  help=('Frequency for repeating events.'),
+                  type=click.INT)
     @click.option('--format', '-f',
                   help=('The format to print the event.'))
     @click.option('--alarms', '-m',
@@ -349,7 +352,7 @@ def _get_cli():
     @click.argument('info', metavar='[START [END | DELTA] [TIMEZONE] [SUMMARY] [:: DESCRIPTION]]',
                     nargs=-1)
     @click.pass_context
-    def new(ctx, calendar, info, location, categories, repeat, until, alarms, url, format,
+    def new(ctx, calendar, info, location, categories, repeat, until, every, alarms, url, format,
             interactive):
         '''Create a new event from arguments.
 
@@ -395,6 +398,7 @@ def _get_cli():
                 repeat=repeat,
                 env={"calendars": ctx.obj['conf']['calendars']},
                 until=until,
+                every=every,
                 alarms=alarms,
                 url=url,
                 format=format,
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -342,7 +342,7 @@ def _get_cli():
     @click.option('--until', '-u',
                   help=('Stop an event repeating on this date.'))
     @click.option('--every', '-e',
-                  help=('Frequency for repeating events.'),
+                  help=('Interval of a repeating event'),
                   type=click.INT)
     @click.option('--format', '-f',
                   help=('The format to print the event.'))
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -420,7 +420,7 @@ def prompt_none(text, default=None, lower=False, type=None):
     if isinstance(default, list):
         default = default[0]
 
-    # icalendar.prop.vRecur return the freq value in upper case,
+    # icalendar.prop.vRecur returns the freq value in upper case,
     # however we only accept lower case values.  For consistency
     # with the interactive mode convert the value for lowercase
     # for display.
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -279,7 +279,7 @@ def khal_list(collection, daterange=None, conf=None, agenda_format=None,
 
 
 def new_interactive(collection, calendar_name, conf, info, location=None,
-                    categories=None, repeat=None, until=None, alarms=None,
+                    categories=None, repeat=None, until=None, every=None, alarms=None,
                     format=None, env=None, url=None):
     try:
         info = parse_datetime.eventinfofstr(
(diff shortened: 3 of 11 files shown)


doc/source/usage.rst
```diff
@@ -412,6 +412,10 @@ Options
 
 * **-u, --until=UNTIL** specify until when a recurring event should run
 
+* **-e, --every=EVERY** specify at what interval a reccurint event
+  should run. For instance, if RRULE is weekly and EVERY is 2, it
+  woulld schedule the event every 2 weeks.
```
WhyNotHugo, line 417:
```suggestion
  would schedule the event every 2 weeks.
```

doc/source/usage.rst
```diff
@@ -412,6 +412,10 @@ Options
 
 * **-u, --until=UNTIL** specify until when a recurring event should run
 
+* **-e, --every=EVERY** specify at what interval a reccurint event
```
WhyNotHugo, line 415:
```suggestion
* **-e, --every=EVERY** specify at what interval a recurring event
```

khal/controllers.py
```diff
@@ -470,7 +471,10 @@ def edit_event(event, collection, locale, allow_quit=False, width=80):
                 until = prompt('until (or "None")', until)
                 if until == 'None':
                     until = None
-                rrule = parse_datetime.rrulefstr(freq, until, locale)
+                interval = prompt('every (or "None")', interval)
+                if interval == 'None':
```
WhyNotHugo, line 475: Empty string should also fall back to `None `.

allanwind: It seem to work with an empty string as input like until above.  This is because, I believe, due to the choice condition that changes the default from None to ''.  I can add a test of empty string, of course, do you want me to do that for until, too?

allanwind: I did so in my last comment, also addressed a couple of defects in editing an existing recurring event.

khal/controllers.py
```diff
@@ -413,6 +413,24 @@ def present_options(options, prefix="", sep="  ", width=70):
...
+def prompt_none(text, default=None, lower=False, type=None):
+    # icalendar.prop.vRecur returns list values for frequency, until and
+    # interval with one element.  However, when creating an event with
+    # new -i they are str values.
+    if isinstance(default, list):
+        default = default[0]
+
+    # icalendar.prop.vRecur return the freq value in upper case,
+    # however we only accept lower case values.  For consistency
+    # with the interactive mode conver the value for lowercase
```
WhyNotHugo, line 425:
```suggestion
    # with the interactive mode convert the value for lowercase
```

WhyNotHugo (approved):

WhyNotHugo:
Looks good! Just a tiny typo.

khal/parse_datetime.py
```diff
@@ -413,12 +413,20 @@ def guessrangefstr(daterange, locale,
...
+def rrulefstr(repeat, until, every, locale):
     if repeat in ["daily", "weekly", "monthly", "yearly"]:
         rrule_settings = {'freq': repeat}
         if until:
             until_dt, is_date = guessdatetimefstr(until.split(' '), locale)
             rrule_settings['until'] = until_dt
+        if every:
+            try:
+                int(every)
+            except ValueError:
```
d7415, line 425: It would be clearer here (and more efficient, not that it really matters) to use `if every.isdigit()` here, as you're not using the cast.

khal/parse_datetime.py
```diff
@@ -413,12 +413,20 @@ def guessrangefstr(daterange, locale,
...
+                logger.fatal("Invalid value for every option. " +
+                             "Must be an integer")
```
d7415, line 427: The "+" isn't required here. Quotes around `every` would make things clearer, to distinguish from "every option must be an integer".
To copy from below:
```suggestion
                logger.fatal("Invalid value for the 'every' option. "
                             "Must be an integer")
```

khal/controllers.py
```diff
@@ -458,20 +476,14 @@ def edit_event(event, collection, locale, allow_quit=False, width=80):
...
-            until = recur["until"] if "until" in recur else ""
-            if not freq:
-                freq = 'None'
-            freq = prompt('frequency (or "None")', freq)
-            if freq == 'None':
-                event.update_rrule(None)
+            freq = prompt_none("frequency", event.recurobject.get("freq"), lower=True)
+            if freq:
+                until = prompt_none("until", event.recurobject.get("until"))
+                interval = prompt_none("interval", event.recurobject.get("interval"), int)
```
d7415, line 482: I also notice you've changed from using the dict syntax to the `.get()` method. As you're not specifying a default I don't see any benefit to this.

d7415 (changes requested):
There are a few things here. It could probably also do with a test.

I'm open to opinions, but I also think we should probably add this to `ikhal` at the same time to prevent events created in `khal` that can't be fully edited in `ikhal`.

khal/cli.py
```diff
@@ -341,6 +341,9 @@ def klist(ctx, include_calendar, exclude_calendar,
                   help=('Repeat event: daily, weekly, monthly or yearly.'))
     @click.option('--until', '-u',
                   help=('Stop an event repeating on this date.'))
+    @click.option('--every', '-e',
+                  help=('Frequency for repeating events.'),
```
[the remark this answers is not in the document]
allanwind: Yeah, I agree, tried a few things different variations to make precise and concise and mixed it up in the final edit. The wording from the spec would be "positive integer representing at which intervals the recurrence rule repeats".  "interval of repeating event"?

khal/controllers.py
```diff
@@ -413,6 +413,24 @@ def present_options(options, prefix="", sep="  ", width=70):
...
 
+def prompt_none(text, default=None, lower=False, type=None):
+    # icalendar.prop.vRecur returns list values for frequency, until and
+    # interval with one element.  However, when creating an event with
+    # new -i they are str values.
+    if isinstance(default, list):
+        default = default[0]
+
+    # icalendar.prop.vRecur return the freq value in upper case,
+    # however we only accept lower case values.  For consistency
```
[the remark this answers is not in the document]
allanwind: Could you spell out what to fix?  It's easy to fix the validation rule, but I thought it was better to enforce the case, then having user enter "weekly" but then in a subsequent edit is now "WEEKLY".

khal/controllers.py
```diff
@@ -413,6 +413,24 @@ def present_options(options, prefix="", sep="  ", width=70):
         return None
 
 
+def prompt_none(text, default=None, lower=False, type=None):
```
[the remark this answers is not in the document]
allanwind: I added the helper to eliminate the duplicate logic for the 3 attributes.  This shrunk the code from 14 to 7 lines not counting the helper.  Planning on using it to fix #1054 which would make it the 4th use, and it's duplicated against for the "else" attribute case and alarm (6 uses).

No love lost on the lower case conversion special case either, but it needs to happen after the type coercing from list to str.  The better alternative is probably to split the function into two: 1. convert value, if list, to str.  Any idea of name for this function?  2. "None" business then handle the case conversion for frequency inline.

In either case, I was looking for input on #1050 to see if we can eliminate the magic "None" value, so this just leaves the type conversion.

khal/parse_datetime.py
allanwind: Yes, I and I later discovered that prompt had a type feature to constraint the input which is a better way of handling it, and later learned that it also have to be positive to I have to change this anyhow.

khal/controllers.py
allanwind: The dict syntax throws a KeyError exception on a missing key while get() returns the default value (None).  Before the code did:

recur = event.recurobject
freq = recur["freq"] if "freq" in recur else ""
if not freq:
  freq = 'None'
freq = prompt('frequency (or "None")', freq)

I thought it was clearer leaving it as None, and handle the magic "None" value in the prompt_none helper as it's really work-around for prompt conflating default values and optional/required values.

khal/parse_datetime.py
d7415: fwiw `.isdigit()` would return false for negatives :)

allanwind: isdigit() ftw :-)

khal/cli.py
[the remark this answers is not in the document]
allanwind: I initially looked at adding/updating tests but found relevant ones to be xfail so not really sure how to proceed.

khal/parse_datetime.py
allanwind: Well, turns out retrieves values are of type vInt so cannot use isdigit() on that, and I need to check for int(every) > 0.  Just a heads up if you wonder why I didn't actually fixed it.

PR #1056, opened 2021-05-27
Remove ANSI escape codes in edit_event. Fixes #1055.

### LINKED ISSUE #1055: edit of datetime range fails with error parsing range
(venv) allan@vent:~/src/khal$ khal new -i
summary: test
datetime range: now
timezone [EST5EDT]:
description (or "None") [None]: test
event saved
Edit?  do[n]e  [s]ummary  [d]escription  da[t]etime range  re[p]eat  [l]ocation  [c]ategories  [a]larm  [D]elete  [u]rl: t                                                                                                                    
datetime range [2021-05-26 20:28 2021-05-26 21:28]:
error parsing range

The main issue is:

1. "error parsing range"

There are a couple of minor issues.

2. The program asks for "datetime range" but I enter a single datetime.  If the ending interval is optional maybe it should be two fields/questions?  It's a bit weird that the "now" maps to 1 hour interval, while "today" is a full day event.

3. For the "datetime range" I entered the value "now", and the program converts that to a pair of timestamps.  It would be better if at least within the new -i session the value that the user entered was retain.  Alternatively, program could tell user that input x was changed to y so it's not a surprise if you want to edit it.

d7415:
On testing, the problem appeared to be `\x1b[0m`, but I figured it was safer to remove all of them. There may be other places where this would be useful, but I'm not sure where.

### DIFF
--- a/khal/controllers.py
+++ b/khal/controllers.py
@@ -24,6 +24,7 @@ import datetime as dt
 import logging
 import os
 import textwrap
+import re
 from collections import OrderedDict, defaultdict
 from shutil import get_terminal_size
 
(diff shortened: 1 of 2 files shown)


geier:
LGTM. I added a test and comment.

PR #1060, opened 2021-05-30
/x/at/calendar/list/search should be quiet if no events are found

### LINKED ISSUE #1059: at/list/search should return nothing if no matching events are found
at/calendar/list currently returns "No events\n" if no events are found.  search returns "\n".  I suggest at, list and search are consistent and return "" if they have nothing to say.  I don't have a strong opinion on calendar.

allanwind:
Fix #1059

### DIFF
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -247,7 +247,7 @@ def test_calendar(runner):
         assert not result.exception
         assert result.exit_code == 0
         output = '\n'.join([
-            "    Mo Tu We Th Fr Sa Su     ",
+            "    Mo Tu We Th Fr Sa Su     No events",
             "Jun  1  2  3  4  5  6  7     ",
             "     8  9 10 11 12 13 14     ",
             "    15 16 17 18 19 20 21     ",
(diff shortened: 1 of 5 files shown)


geier:
lgtm. @WhyNotHugo  what do you think?

geier:
Sleeping about this, we probably should at least add some debug logging, that no events were found.

tests/cli_test.py
```diff
@@ -104,7 +104,7 @@ def test_direct_modification(runner):
     runner = runner()
 
     result = runner.invoke(main_khal, ['list'])
-    assert result.output == 'No events\n'
+    assert result.output == ''
```
WhyNotHugo, line 107: This change should make scripted usage a lot easier. :+1:

tests/cli_test.py
```diff
@@ -245,7 +245,7 @@ def test_calendar(runner):
         assert not result.exception
         assert result.exit_code == 0
         output = '\n'.join([
-            "    Mo Tu We Th Fr Sa Su     No events",
```
WhyNotHugo, line 248: Not sure I'm convinced about rendering removing this one -- we're rendering the calendar anyway, so it looks kinda odd that there's nothing on the side.

I can't imagine this output being parsed by scripts either; it's far from ideal for being parsed.

PR #1061, opened 2021-05-30
dstart and until should be same type (floating or localized)

### LINKED ISSUE #810: khal new --repeat --until produces invalid VEVENTs
DTSTART and UNTIL are sometimes not of the same type (floating / localized)

allanwind:
Fix #810

### DIFF
--- a/tests/parse_datetime_test.py
+++ b/tests/parse_datetime_test.py
@@ -10,8 +10,8 @@ from khal.parse_datetime import (construct_daynames, eventinfofstr,
                                  weekdaypstr)
 from khal.icalendar import new_event
 
-from .utils import (LOCALE_BERLIN, LOCALE_NEW_YORK, _replace_uid,
-                    normalize_component)
+from .utils import (LOCALE_BERLIN, LOCALE_FLOATING, LOCALE_NEW_YORK,
+                    _replace_uid, normalize_component)
 
 
 def _create_testcases(*cases):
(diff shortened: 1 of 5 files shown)


allanwind:
Ticket was written as a cli issue.  That said, I also looked very closely at the class LocalizedEvent which localizes start and end as needed which comes into play via Event.fromVEvents().  Both import_event and pritnics call Event.fromVEvents().

tests/parse_datetime_test.py
```diff
@@ -497,7 +497,7 @@ def test_description():
          'DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20140216T080000',
          'DTEND;TZID=Europe/Berlin;VALUE=DATE-TIME:20140216T090000',
          'DESCRIPTION:please describe the event',
-         'RRULE:FREQ=DAILY;UNTIL=20150605T000000')))
+         'RRULE:FREQ=DAILY;UNTIL=20150604T230000Z')))
```
WhyNotHugo, line 500: Do you think it makes sense to create a copy of this test with non-localised times?

WhyNotHugo (approved):
Makes sense to me.

tests/parse_datetime_test.py
allanwind: Done.

WhyNotHugo:
Thanks!

PR #1066, opened 2021-07-01
Add new permutations for testing print_ics and condense into re. Fixes #1065

d7415:
This adds the permutations from #1065 into the valid responses. 8 options seemed excessive, so I've condensed it into a regular expression. Happy to hear any objections to that.

PR #1067, opened 2021-07-02
Install type stubs and dependencies for mypy

WhyNotHugo:
mypy no longer ships type stubs and they must now be installed
separately.

Configure tox to install these for the `mypy` environment, as well as
regular dependencies (since many of these include typing hints
themselves).

I've moved `ignore_missing_imports` to a mypy-specific section, so when
mypy is run via other tools (e.g.: code editors) this settings is
applied too.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,4 @@
+[mypy]
+ignore_missing_imports = True
+# See https://github.com/python/mypy/issues/7511:
+warn_no_return = False
--- a/tox.ini
+++ b/tox.ini
@@ -59,8 +59,3 @@ commands =
 max-line-length = 100
 ignore = E252,W504,E121,B008
 exclude = .tox,examples,doc
-
-[mypy]
-ignore_missing_imports = True
-# See https://github.com/python/mypy/issues/7511:
-warn_no_return = False
--- a/tox.ini
+++ b/tox.ini
@@ -48,13 +48,19 @@ commands =
   make -C doc man
 
 [testenv:mypy]
-skip_install=True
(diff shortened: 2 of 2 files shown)


d7415:
Alternative would be `mypy --install-types --non-interactive --ignore-missing-imports khal`. Probably less correct, but it means we're less likely to miss a dependency later on. I added it to #1066 (which does need splitting up) (https://github.com/pimutils/khal/pull/1066/commits/e68c050f71c877e55efb976f12d5c5eadf600c59)

WhyNotHugo:
> Note that using the --install-types is problematic. Mutating the pre-commit environment at runtime breaks cache and will break parallel builds.

https://github.com/pre-commit/mirrors-mypy

WhyNotHugo:
(ignore that last message, mixup)

WhyNotHugo:
I think `--install-types --non-interactive` is better than what I'm doing here.

`--ignore-missing-imports`, however, should be specified in `setup.cfg` or `tox.ini`, so that other tools can pick that value up (e.g.: mypy when run via `nvim`).

d7415:
> `--ignore-missing-imports`, however, should be specified in `setup.cfg` or `tox.ini`, so that other tools can pick that value up (e.g.: mypy when run via `nvim`).

That makes sense. I feel like I've seen that elsewhere.

WhyNotHugo:
`todoman` and `vdirsyncer` follow that pattern, and it's pretty common in general.

d7415 (approved):
Makes sense to me.

PR #1068, opened 2021-07-02
Run all Python versions on MS Actions

WhyNotHugo:
This is a rebase of  #1013

### DIFF
--- a/.github/workflows/basic_test.yml
+++ b/.github/workflows/basic_test.yml
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.9"]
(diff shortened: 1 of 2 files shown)


d7415:
I've managed to get past that point (I think it was https://github.com/pimutils/khal/commit/647d41ab14f2efe1057ac5f3cd49319a06076f14) in `d7415_gha`, then needed to use #1066.

Now I think I'm down to the tests that some package maintainers were finding temperamental already (e.g. last run it was `test_event_different_timezones`, `test_default_calendar` and `test_birthdays`, none consistently failing).

I'm going to try to take a closer look at those, but failing that it may be worth xfailing them (and opening an issue against it) so that we have the CI in the meantime.

PR #1074, opened 2021-07-24
Imported events with no DTEND or DURATION will now last one hour (for DATETIME) or one day (for DATE)

### LINKED ISSUE #1071: How to display the hour in ikhal (+ is there a dicussion forum for basic question)
Hello again. I'm sorry to bother the issue with a rather basic question. I want to output event with their hour time with `ikhal`. I could do it with the command:

```
khal list --format "{title} {start-time-full}"
```
I tried to modify the conf file in several ways without success:

```
[calendars]

  [[private]]
    path = /path/to/calendar/
    type = calendar
    color = dark green

[locale]
timeformat = %H:%M
dateformat = %d/%m/%Y
longdateformat = %d/%m/%Y
datetimeformat = %d/%m/%Y %H:%M
longdatetimeformat = %d/%m/%Y %H:%M

[default]
default_calendar = private
```
I tried to read the doc [here](https://khal.readthedocs.io/en/latest/usage.html#) and [there](https://khal.readthedocs.io/en/latest/configure.html) but unfortunately could not find how to implement the answer

More generally is there a dicussion forum for basic questions as this one?

Many thanks in advance

d7415:
Fixes #1071

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,8 @@ not released
 * CHANGE ikhal: tab (and shift tab) jump from the events back to the calendar
 * NEW Add symbol for events with at least one alarm
 * FIX URL can now be set/updated from ikhal
+* FIX Imported events without an end or duration will now last one day if
+   `DTSTART` is a date (as per RFC) or one hour if it is a datetime.
 
 0.10.3
 ======
--- a/khal/icalendar.py
+++ b/khal/icalendar.py
@@ -413,8 +413,9 @@ def sanitize_timerange(dtstart, dtend, duration=None):
 
     if dtend is None and duration is None:
         if isinstance(dtstart, dt.datetime):
-            dtstart = dtstart.date()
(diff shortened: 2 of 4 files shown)


d7415:
I put it back to one hour to match expected behaviour elsewhere. I think one minute (or maybe one second) might be more appropriate considering instantaneous events are allowed in the spec, but that seems like a job for another PR.

Dropping the start time entirely and making them all-day presumably had a reason, but is against the RFC and effectively lost data.

WhyNotHugo (approved):
I think one hour is very reasonable, though maybe using the setting `default_dayevent_duration` would not be unreasonable.

I'd leave it as-is, and wait for feedback from anyone who dislikes this approach.

d7415:
> I'd leave it as-is, and wait for feedback from anyone who dislikes this approach.

Yeah, I'll merge this PR to deal with #1071 and if someone wants a different duration (or if it bugs me) I'll look at raising another one to use `default_dayevent_duration` or some such.

PR #1075, opened 2021-07-25
Another attempt at GitHub Actions

d7415:
Some inspiration from #1068, plus extra locales, #1066 and a few other tweaks managed to get an almost-pass ([everything except pytz-latest](https://github.com/pimutils/khal/runs/3154445486?check_suite_focus=true#step:6:165)) before opening it as a PR. Let's see if it lasts...

Closes #1068
Closes #1013 
Closes #1066

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,13 +10,19 @@ on:
 
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
+    name: "Python ${{ matrix.python-version }} ${{ matrix.tox-test }}"
     runs-on: "ubuntu-latest"
 
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.6", "3.7", "3.8", "3.9"]
+        tox-test: ["default"]
+        include:
+          - python-version: "3.9"
+            tox-test: "mypy"
+          - python-version: "3.9"
+            tox-test: "style"
 
     steps:
       - uses: "actions/checkout@v2"
@@ -39,6 +45,19 @@ jobs:
           python -m pip install --upgrade coverage[toml] virtualenv tox tox-gh-actions          
 
       - name: "Run tox targets for ${{ matrix.python-version }}"
+        if: ${{ matrix.tox-test == 'default' }}
         # Fake a TTY
         shell: 'script -q -e -c "bash --noprofile --norc -eo pipefail {0}"'
         run: "python -m tox"
+
+      - name: "Run tox targets for mypy"
+        if: ${{ matrix.tox-test == 'mypy' }}
+        # Fake a TTY
+        shell: 'script -q -e -c "bash --noprofile --norc -eo pipefail {0}"'
+        run: "python -m tox -e mypy"
+
+      - name: "Run tox targets for style"
+        if: ${{ matrix.tox-test == 'style' }}
+        # Fake a TTY
+        shell: 'script -q -e -c "bash --noprofile --norc -eo pipefail {0}"'
+        run: "python -m tox -e style"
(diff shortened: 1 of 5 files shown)


d7415:
Different failures each time. When they do fail they seem to be the same tests that have had issues previously (e.g. #860).

d7415:
All of the failures switch between the 5 tests that use 
```py
vdirs[cal1].upload(DumbItem(
```

Those are:

- test_multi_uid_vdir
- test_event_different_timezones
- test_birthdays
- test_birthdays_29feb
- test_birthdays_no_year

d7415:
It's a race condition.

.github/workflows/ci.yml
```diff
@@ -33,4 +39,5 @@ jobs:
           python -m pip install --upgrade coverage[toml] virtualenv tox tox-gh-actions          
 
       - name: "Run tox targets for ${{ matrix.python-version }}"
+        shell: 'script -q -e -c "bash --noprofile --norc -eo pipefail {0}"'
```
WhyNotHugo, line 42: Do you mind leaving a comment as to what this does / why it's necessary?

I can imagine us looking at it in a year and having no idea :sweat_smile:

tests/khalendar_test.py
```diff
@@ -339,9 +339,10 @@ def test_invalid_timezones(self, coll_vdirs):
         assert events[0].format('{start} {end} {title}', dt.date.today()) == \
             '02.12. 08:00 02.12. 09:30 Some event\x1b[0m'
 
-    def test_multi_uid_vdir(self, coll_vdirs, caplog, fix_caplog):
+    def test_multi_uid_vdir(self, coll_vdirs, caplog, fix_caplog, sleep_time):
         coll, vdirs = coll_vdirs
         caplog.set_level(logging.WARNING)
+        sleep(sleep_time)
```
WhyNotHugo, line 345: Are these required due to poor clock resolution on the CI environment or something?

WhyNotHugo (commented):
Neat! Thanks for getting this working.

Just some questions on in, but LG.

.github/workflows/ci.yml
d7415: Good thinking!

tests/khalendar_test.py
d7415: I'm not sure about the details, but when I traced the failures I found that `needs_update()` was false, and on further investigation the ctag in the DB and the filesystem were returning the [same timestamp](https://github.com/pimutils/khal/runs/3156981790?check_suite_focus=true#step:6:183) (to what *looks* like high resolution).

I assume that the DB ctag is being created/reset at the start of the test, and if the upload step is faster than the clock resolution the test fails. The sleep used elsewhere in `khalendar_test.py` (~0.011s) seems to be enough to avoid this.

I guess I should comment these too...

d7415:
I've also change it so that the mypy and style tests run as separate jobs, so in parallel and easier to see vs the normal py39 tests.

I reduced some duplication in tox.ini in https://github.com/pimutils/khal/commit/53b3fd5bbfbaa7d1a3ea68240f3502589daae55d but I think that sacrifices readability.

tests/khalendar_test.py
WhyNotHugo: Note that they all end in `.000000000`, so it seems they're not high resolution. I wonder what filesystem is being used under the hood that has this issue. AFAIK, we've never seen anything this low resolution.

Some background:

`needs_update` will only return `True` if the file has changed since we last read it. If the clock is **too** low precision, then when we edit the file a second time, it's mtime hasn't changed (since the clock hasn't moved forward due to its lack of precision).

In simpler terms: if the clock only has milisecond precision, then we must wait at least 1.1 miliseconds between edits for the timestamp to change. `sleep_time` determines how much time we need to sleep (see its docs for details).

WhyNotHugo (approved):
LGMT.

Issue #1078, opened 2021-08-11
khal new adds start date as end date and end date to summary


Konfekt:
With

```
khal --version
khal, version 0.10.4

khal printformats
longdatetimeformat: 2013-12-21T21:45
datetimeformat: 2013-12-21
longdateformat: 2013-12-21
dateformat: 2013-12-21
timeformat: 21:45
```

the commands

```
khal new 2021-8-11 11:00 2021-8-11 12:00 test
khal list today
```

give output

```
00:00-11:00 2021-8-11 12:00 test
```

Expected would be

```
11:00-12:00 2021-8-11 test
```

d7415:
> longdatetimeformat: 2013-12-21T21:45

If you change the `T` to a space this should work as you are expecting it to.

Konfekt:
Yes that is the work around I use [here](https://github.com/pimutils/khal/pull/1079?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDIyNzYzNTYwNzU6NzE5NTU1Mw%3D%3D#pullrequestreview-727043834) but it breaks default configuration. 

But then neither does

```
khal new '2021-8-11 11:00' '2021-8-11 12:00' test 

khal list

00:00-11:00 2021-8-11 12:00 test

```

nor as [officially documented](https://khal.readthedocs.io/en/latest/usage.html) :

```
khal new 2021-8-11 18:00 24:00 Another Event :: with Alice and Bob

khal list

00:00-18:00 24:00 Another Event :: with Alice and Bob
```

Therefore this issue is not fixed.

Konfekt:
Oh, I see, khal only accepts one single date format?

d7415:
> Oh, I see, khal only accepts one single date format?

It accepts one of each of the formats in `khal printformats`

Konfekt:
Yes, but only one longdatetimeformat (otherwise it complains about a doubled entry in the config file). I overread

```
The exact format of longdatetimeformat, datetimeformat, timeformat, longdateformat and dateformat can be configured in the configuration file. Valid keywords for dates are today, tomorrow, the English name of all seven weekdays and their three letter abbreviations (their next occurrence is used).
```

in the [documentation](https://khal.readthedocs.io/en/latest/usage.html) and forgot that at some point I changed the default time format to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601).

PR #1079, opened 2021-08-11
Mutt2khal: pass more robust date formats to khal and shell escape characters

Konfekt:
pass more robust date formats to khal and shell escape characters

see also https://github.com/pimutils/khal/issues/1078

### DIFF
--- a/misc/mutt2khal
+++ b/misc/mutt2khal
@@ -40,7 +40,7 @@
 }
 
 END {
-  printf("khal new --location %s %sT%s %sT%s %s :: %s", loc, d_st, t_st, d_end, t_end, summ, desc) | "sh"
+  printf("khal new --location %s %s %s %s %s %s :: %s", loc, d_st, t_st, d_end, t_end, summ, desc) | "sh"
 }
 
 ## IMPORTANT ##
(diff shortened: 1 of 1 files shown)


misc/mutt2khal
```diff
@@ -37,7 +40,7 @@
 }
 
 END {
-  print "khal new", loc, d_st, t_st, d_end, t_end, summ, desc | "sh"
+  printf("khal new --location %s %sT%s %sT%s %s :: %s", loc, d_st, t_st, d_end, t_end, summ, desc) | "sh"
```
d7415, line 43: The `T`s here will break the default configuration.

Konfekt: Yes, but according to https://github.com/pimutils/khal/issues/1078#issuecomment-896695860 it is the only way to currently pass dates to khal.

Konfekt: Okay, got it, khal only accepts the one date and time format specified in the config; I somehow thought it always tries to interpret all formats thrown at it (independent of the configuration).

Konfekt: Corrected it  to assume the default format again.

geier:
closing and opening to force tests to re-run

PR #1082, opened 2021-08-20
Add support for processing and importing ics without UID

viiru-:
Files like this seem to be produced by all kinds of proprietary reservation
systems and such. They are clearly out of spec, but since they exist I'd
prefer to be able to import them without needing to manually add an UID
every time.

Another alternative would be to report an error if there is no UID and
random_uid=False, and then doing this only when random_uid=True. I'm not
sure it would be an improvement over this simpler solution, so I'll offer
this first.


WhyNotHugo:
The general idea here makes sense to me, I mostly have one suggestion.

I would rather make the UID the hash (sha256?) of the input event itself, so that if we handle the exact same event twice, it gets the same UID, and we avoid creating a duplicate entry.

What do you think?

d7415:
I'd also suggest adding a warning, so that a user importing it knows it's a bad file.

viiru-:
Both suggestions sound good to me. I'll rework and get back to you (sadly not today, I'm out of time for the moment). Thanks for the review @WhyNotHugo and @d7415.

viiru-:
I have pushed a new version that implements the suggested improvements, can you review again?

khal/icalendar.py
```diff
@@ -62,6 +63,12 @@ def split_ics(ics, random_uid=False, default_timezone=None):
             tzs[key] = item
 
         if item.name == 'VEVENT':
+            if 'UID' not in item:
+                logger.warning(
+                    f"Event with summary '{items['SUMMARY']' doesn't have a unique ID,"
```
d7415, line 68:
```suggestion
                    f"Event with summary '{item['SUMMARY']}' doesn't have a unique ID. "
```

viiru-: Indeed. Dunno how I missed that. In my hurry I must have only run the tests on an earlier version where the summary wasn't yet logged... Thanks.

WhyNotHugo (approved):
LGTM

WhyNotHugo:
Merged, thanks!

Issue #1092, opened 2021-10-21
Pytz warnings when using khal


rodriguegaspard:
Hello, I did a full system upgrade yesterday, and I now have those warnings when I'm trying to display events or calendars (`khal at`, `khal list`, `khal calendar`)

```
/usr/lib/python3.9/site-packages/khal/khalendar/khalendar.py:153: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  localized_events = self.get_localized(localize(start), localize(end))
/usr/lib/python3.9/site-packages/khal/khalendar/event.py:541: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  start_local_datetime = self._locale['local_timezone'].localize(
/usr/lib/python3.9/site-packages/khal/khalendar/event.py:543: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  end_local_datetime = self._locale['local_timezone'].localize(
/usr/lib/python3.9/site-packages/khal/khalendar/event.py:546: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  day_start = self._locale['local_timezone'].localize(
/usr/lib/python3.9/site-packages/khal/khalendar/event.py:549: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  day_end = self._locale['local_timezone'].localize(
[...]

PR #1093, opened 2021-10-22
Update khal to Follow PEP495 Removing Deprecation Warnings

pjkaufman:
When running `khal ...` or `ikhal ...` an issue would arise where warnings about deprecated functions being used (see https://github.com/pimutils/khal/issues/1092). This was not a critical change that was needed, but it was more to remove the warnings as an eye sore for myself and those that would not be able to do so on their own.

Changes Made:
- Updated AUTHORS.txt as per the documentation for contributing
- Imported and used ZoneInfo in order to assign a timezone where the warnings were being thrown (2 locations in `khalendar.py` and `event.py` though I decided to add two more locale format updates in one of the methods here)

I had to convert the timezone info to a string just in case it was something else because when I did not do so, I got errors.
Also, I did initially try to change all of the `localize` methods to use ZoneInfo, but that started causing test cases to fail for certain files, so I opted to just fix the visible issue.

I am newish to python, but have used it to write programs or scripts before. So please let me know if there is something I missed, a better way to do things, or if this PR should be more extensive.

Here is the source I am using for determine how to convert code to the PEP495: https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html

**Note: there is no longer a built in list of transition dates for daylight savings times, so it will need to be decided how to handle this especial since the standards for this have changed several times**.


pjkaufman:
I think I have something for DST times that are newer than 2007 (I think I can add other date ranges as well, it is just that the algorithm for doing so will need to be modified. I am having issues with DST transition times. Is there a way outside of pytz to do determine its transition times?

Here is what I am thinking for the new create_timezone method:
```
def create_timezone(tz, first_date=None, last_date=None):
...
    # TODO last_date = None, recurring to infinity

    first_date = dt.datetime.today() if not first_date else to_naive_utc(first_date)
    # TODO: determine if there is a way to do this for BST pjk
    # check to see if timezone is a part of dst by checking if before or after the transition it has an offset
    if first_date.dst() == dt.timedelta(0) and first_date.replace(fold= 1).dst() == dt.timedelta(0):
        return _create_timezone_static(tz)
    
    last_date = dt.datetime.today() if not last_date else to_naive_utc(last_date)
    timezone = icalendar.Timezone()
    timezone.add('TZID', tz)

    transition_times = _get_transition_dates_dst_for_daterange(tz, first_date.replace(tzinfo=tz), last_date.replace(tzinfo=tz))
    timezones = {}
    for index in range(1, len(transition_times)):
        name = transition_times[index].tzname()
        if name in timezones:
            ttime = transition_times[index].replace(tzinfo=None)
            if 'RDATE' in timezones[name]:
                timezones[name]['RDATE'].dts.append(
                    icalendar.prop.vDDDTypes(ttime))
            else:
                timezones[name].add('RDATE', ttime)
            continue

        if transition_times[index].dst() != dt.timedelta(0):
            subcomp = icalendar.TimezoneDaylight()
        else:
            subcomp = icalendar.TimezoneStandard()
        
        subcomp.add('TZNAME', name)
        subcomp.add(
            'DTSTART',
            tz.fromutc(transition_times[index]).replace(tzinfo=None) # Note: for some reason this is using EDT instead of the zone info name for the value here, I need to figure out why as this is incorrect
        )
        
        subcomp.add('TZOFFSETTO', transition_times[index].utcoffset())
        subcomp.add('TZOFFSETFROM', transition_times[index - 1].utcoffset())
        timezones[name] = subcomp

    for subcomp in timezones.values():
        timezone.add_component(subcomp)

    return timezone
```

Here is the DST algorithm (this is more a proof of concept than an actual final algorithm):
```
def _get_transition_dates_dst_for_daterange(tz, start_date, end_date):
    #TODO: add date range based algorithms for determining the correct dst time zones where rules exist pjk
    if start_date.year < 2007:
        msg = (
                f"Cannot work with start dates prior to 2007 at this time. Date supplied was" + str(start_date) + "."
            )
        logger.fatal(msg)
        raise FatalError(  # because in ikhal you won't see the logger's output
            msg
        )

    year = start_date.year-1
    years = []
    while year <= end_date.year+1:
        years.append(year) 
        year +=1

    index_of_first_transition_before_start_date = 0
    index = 0
    transition_times = []
    for year in years:
        marchCal = calendar.monthcalendar(year, 3)
        first_week  = marchCal[0]
        second_week = marchCal[1]
        third_week  = marchCal[2]

        if first_week[calendar.SUNDAY]:
            dst_begins = second_week[calendar.SUNDAY]
        else:
            dst_begins = third_week[calendar.SUNDAY]

        dst_start = dt.datetime(year, 3, dst_begins, 2, tzinfo= tz)
        transition_times.append(dst_start)
        if dst_start < start_date:
            index_of_first_transition_before_start_date = index
        
        if end_date < dst_start:
            break
        
        novemberCal = calendar.monthcalendar(year, 11)
        first_week  = novemberCal[0]
        second_week = novemberCal[1]

        if first_week[calendar.SUNDAY]:
            dst_ends = first_week[calendar.SUNDAY]
        else:
            dst_ends = second_week[calendar.SUNDAY]
        
        dst_end = dt.datetime(year, 11, dst_ends, 2, tzinfo= tz)
        transition_times.append(dst_end)
        if end_date < dst_end:
            break
        
        index +=1
    
    return transition_times[index_of_first_transition_before_start_date:]
```

WhyNotHugo:
@bancsorin10 Agreed. A separate PR pinning a working version of pytz is welcome, at least until we have a definitive version of PEP495 support.

pjkaufman:
I made some changes to remove a lot of the pytz method calls and it seems to work for some things. The unit tests are failing (41 of them to be exact) and some of that is because I have not implemented logic for DST dates prior to 2007, I have no logic in place for BST dates, and the TZID is being listed at the timezone name instead of the ZoneInfo key.

However, I tried to commit my changes, but I got several errors from the precommit checks. They currently are preventing me from actually commiting to my own repo, so I will just list out some of the errors and ask for ideas on how to fix them.

The first issue it is really complaining about is the following:
```
try:
    from zoneinfo import ZoneInfo
except ImportError:
    from backports import zoneinfo as ZoneInfo
```
The precheck says the following about it: 
```
khal/utils.py:29: error: Incompatible import of "ZoneInfo" (imported name has type Module, local name has type "Type[ZoneInfo]")
tests/utils.py:7: error: Incompatible import of "ZoneInfo" (imported name has type Module, local name has type "Type[ZoneInfo]")
```
Any ideas on how to fix this?

pjkaufman:
I figured out how to get past the the precommit check failures for `flake8`, but I am not sure what I need to do to fix the actual errors. Please let me know what I should do to fix them. I am also unaware of what we should do to correctly replace transition time checking and BST time checking. Any ideas are welcome.

khal/icalendar.py
```diff
@@ -22,10 +22,13 @@
 """collection of icalendar helper functions"""
 
 import datetime as dt
+try:
+    from zoneinfo import ZoneInfo
+except ImportError:
+    from backports import zoneinfo as ZoneInfo
```
pjkaufman, line 28: For some reason flake8 does not like this. Any ideas on how to fix it?

khal/khalendar/event.py
```diff
@@ -923,3 +910,62 @@ def _create_timezone_static(tz):
...
+            dst_ends = second_week[calendar.SUNDAY]
+
+        dst_end = dt.datetime(year, 11, dst_ends, 2, tzinfo=tz)
+        transition_times.append(dst_end)
+        if end_date < dst_end:
+            break
+
+        index += 1
+
+    return transition_times[index_of_first_transition_before_start_date:]
```
pjkaufman, line 971: Here is an idea for how to get transition dates without pytz.

khal/khalendar/event.py
```diff
@@ -843,60 +844,46 @@ def create_timezone(tz, first_date=None, last_date=None):
...
             subcomp = icalendar.TimezoneStandard()
 
-        subcomp.add('TZNAME', tz._transition_info[num][2])
+        subcomp.add('TZNAME', name)
         subcomp.add(
             'DTSTART',
-            tz.fromutc(tz._utc_transition_times[num]).replace(tzinfo=None))
-        subcomp.add('TZOFFSETTO', tz._transition_info[num][0])
-        subcomp.add('TZOFFSETFROM', tz._transition_info[num - 1][0])
+            tz.fromutc(transition_times[index]).replace(tzinfo=None)
```
pjkaufman, line 882: For some reason the `TZID` is being listed as the timezone name (i.e. EDT, CET, etc.) instead of the passed in name for the ZoneInfo (i.e. "UTC", "America/New_York", etc.).

khal/settings/utils.py
```diff
@@ -25,7 +25,10 @@
 import os
 from os.path import expanduser, expandvars, join
 
-import pytz
+try:
+    import zoneinfo as ZoneInfo
+except ImportError:  # I am not sure if this is correct for the backport
+    from backports import zoneinfo as ZoneInfo
```
pjkaufman, line 31: I am not sure if this is correct for the backport as I did not change it while I imported the module instead of a module inside zoneinfo for the try portion of the try except.

khal/controllers.py
```diff
@@ -27,7 +27,10 @@
 from collections import OrderedDict, defaultdict
 from shutil import get_terminal_size
 
-import pytz
+try:
+    import zoneinfo as ZoneInfo
+except ImportError:  # I am not sure if this is correct for the backport
+    from backports import zoneinfo as ZoneInfo
```
RonnyPfannschmidt, line 33:
```suggestion
    from backports.zoneinfo import ZoneInfo
```

bancsorin10:
> @bancsorin10 Agreed. A separate PR pinning a working version of pytz is welcome, at least until we have a definitive version of PEP495 support.

from my testing I could not pinpoint the problem. I used python 3.9 with pytz version as old as 2013, and python 3.7.7 with the newest versions of pytz. It might be a combination of both or my system may just be too corrupted even tho I made the builds with a virtual python environment. I will try a few more versions and if I find something before this pr goes live I'll let you know.

Terrance:
@bancsorin10 The issue is likely with tzlocal rather than pytz directly, see my PR linked above.

khal/icalendar.py
bancsorin10: is this problem still up? Is there a full log of the error. I will look into it and come back

khal/controllers.py
WhyNotHugo: When flake8 runs on Python3.9, this library won't be installed. It's safe to ignore this specific error.

pjkaufman:
I have been able to determine that the TZID issue is probably an icalendar issue where it is not handling ZoneInfo and date util timezone names correctly: https://github.com/collective/icalendar/issues/291

pjkaufman:
I can definitely continue working on these changes, but I will probably need some help with fixing the unit tests as I am not the most familiar with timezone information especially when stored using ical.

khal/controllers.py
```diff
@@ -320,10 +323,10 @@ def new_interactive(collection, calendar_name, conf, info, location=None,
         tz = info.get('timezone') or conf['locale']['default_timezone']
         timezone = prompt("timezone", default=str(tz))
         try:
-            tz = pytz.timezone(timezone)
+            tz = ZoneInfo.ZoneInfo(str(timezone))
             info['timezone'] = tz
             break
-        except pytz.UnknownTimeZoneError:
+        except ZoneInfo._common.ZoneInfoNotFoundError:
```
WhyNotHugo, line 329: Use the "public" path for this: `from zoneinfo import ZoneInfoNotFoundError`

setup.py
```diff
@@ -16,7 +16,7 @@
     'icalendar>=4.0.3',
     'urwid>=1.3.0',
     'pyxdg',
-    'pytz',
+    "backports.zoneinfo;python_version<'3.9'",
```
WhyNotHugo, line 19: Please add `types-backports` to mypy's [additional dependencies](https://github.com/pimutils/khal/blob/16e9540f3bc51cebfae3015898cbee645d4bfd92/.pre-commit-config.yaml#L39) so that mypy checks pass with the new dependency.

WhyNotHugo (changes requested):

WhyNotHugo:
Close-opened so CI will run, since GitHub seems to have dropped the last logs.

geier:
Having a look at this now.

It's a bit unfortunate, that the only way to localize an existing datetime seems to be to use `.replace(tzinfo=newtz)`, as this doesn't except like pytz's localize() did, if the existing datetime is already timezone aware.

geier:
We might want to remove pytz from icalendar first, so we can get rid of all the ugly `ZoneInfo(str(timezone))`s.

geier:
Thank you, @pjkaufman! This really helps!

I rebased this PR and fixed the first outstanding issue to #1125 for now.

Especially the timezone `create_timezone()` is a really ugly piece of code I never had the chance to properly clean up. I'll try to see if I can find the time to fix this outstanding issues (no promises).

geier:
Regarding creating the VTIMEZONEs:

we can probably create something similar as before from `ZoneInfo._ttinfos`, `Zoneinfo._trans_utc` and `Zoneinfo._fixed_offsets`. But we could use the chance to read the timezone information values and extract the rrules from those (would lead to smaller VTIMEZONEs).

pjkaufman:
> Regarding creating the VTIMEZONEs:
> 
> we can probably create something similar as before from `ZoneInfo._ttinfos`, `Zoneinfo._trans_utc` and `Zoneinfo._fixed_offsets`. But we could use the chance to read the timezone information values and extract the rrules from those (would lead to smaller VTIMEZONEs).

That is quite interesting. I was unaware that this was a thing. I can see about adding it if you would like.

pjkaufman:
> We might want to remove pytz from icalendar first, so we can get rid of all the ugly `ZoneInfo(str(timezone))`s.

This would be great if it could be updated since it looks like pytz is being removed/deprecated.

geier:
> > Regarding creating the VTIMEZONEs:
> > we can probably create something similar as before from `ZoneInfo._ttinfos`, `Zoneinfo._trans_utc` and `Zoneinfo._fixed_offsets`. But we could use the chance to read the timezone information values and extract the rrules from those (would lead to smaller VTIMEZONEs).
> 
> That is quite interesting. I was unaware that this was a thing. I can see about adding it if you would like.

Yes, please go ahead.

geier:
@pjkaufman you can also pull the rebased branch into this.

geier:
Also the ics project seems to be able to create VTIMEZONEs, perhaps it's valuable to have a look at their approach: https://github.com/ics-py/ics-py/blob/main/src/ics/timezone/__init__.py

pjkaufman:
> > > Regarding creating the VTIMEZONEs:
> > > we can probably create something similar as before from `ZoneInfo._ttinfos`, `Zoneinfo._trans_utc` and `Zoneinfo._fixed_offsets`. But we could use the chance to read the timezone information values and extract the rrules from those (would lead to smaller VTIMEZONEs).
> > 
> > 
> > That is quite interesting. I was unaware that this was a thing. I can see about adding it if you would like.
> 
> Yes, please go ahead.

Could you please link out to the documentation for `ZoneInfo._ttinfos`, `Zoneinfo._trans_utc` and `Zoneinfo._fixed_offsets` as I cannot seem to find it? Without the documentation and the fact that python 3.9.7 keeps saying that 
```ZoneInfo.__getattr__('_ttinfos')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/zoneinfo/__init__.py", line 27, in __getattr__
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
AttributeError: module 'zoneinfo' has no attribute '_ttinfos'
```

geier:
I don't think there is any documentation for those _ variables apart from the [source code](https://github.com/python/cpython/blob/3.10/Lib/zoneinfo/_zoneinfo.py)

geier:
It looks like if we import ZoneInfo like this

```
In [1]: from zoneinfo._zoneinfo import ZoneInfo

In [2]: berlin = ZoneInfo('Europe/Berlin')

In [3]: berlin._trans_local

```
we can force to import the python version.

Also, we would need to first upgrade collectivie/icalendar to use tzinfo to completely switch to tzinfo.

But the current vtimezone implementation definitively is a mess and needs an upgrade. 

Perhaps it's more valuable, to have a look at how the ZoneInfo implementation loads the timezone, but I'm not sure what the best way forward here is at the moment.

pjkaufman:
I plan to make a couple of changes yet on the master branch of my fork, but I would like to ask that someone that uses this app takeover the changes from there since I no longer use this app which means that I don't have a good way to verify that it is working correctly beyond the UTs which may not be extensive enough to be considered proof that the app is working, besides the fact that my changes modify them as well.

I would also recommend updating icalendar to allow for use without pytz as that was a major blocker that has drawn out the updating of this repo.

Beyond that, I plan to see about swapping out the logic for timezone info and finishing the rebase of the branch before stopping making changes to my fork.

I do hope that this change gets implemented as it seems necessary. I also hope that the code that has been added here will aid in the transition from pytz.

Please let me know if there would be a problem with me finishing the rebase and making the timezone transition times change before closing out the PR to allow another dev to hopefully take over.

pjkaufman:
@WhyNotHugo , @geier , I have made several of the changes requested, but I am not able to build it locally and do not have the expertise to figure out the python errors. However since I plan to relinquish the remaining work and allow someone that is familiar with python and is using khal to complete this PR, I do not see resolving the local build problems due to bad dependency installations as an issue.

Hopefully this will PR can serve as a base for making the change over to no longer use pytz. I hope that the dependencies get updated soon to be compliant with no longer using pytz to help this project keep moving forward. Thank you for letting me try my hand at creating this fix.

PR #1098, opened 2021-11-03
Pin to tzlocal 2.x to (temporarily) avoid zoneinfo upgrade

Terrance:
This prevents use of newer versions of tzlocal -- 3.0 replaced its own use of pytz with zoneinfo; 4.0 introduced pytz_deprecation_shim but that doesn't cover everything, as seen in #1093.  This will keep the current code working until khal switches to PEP-495.


WhyNotHugo (approved):

WhyNotHugo:
Thanks!

PR #1113, opened 2022-02-12
xfail the tag_sync test as well (Fixes #1112)

### LINKED ISSUE #1112: `test_etag_sync` fails on khal 0.10.4
`test_etag_sync` fails on khal 0.10.4 on openSUSE Tumbleweed aarch64 and armv7.

```
[   82s] =================================== FAILURES ===================================
[   82s] ________________________________ test_etag_sync ________________________________
[   82s] 
[   82s] tmpdir = local('/tmp/pytest-of-abuild/pytest-1/test_etag_sync0')
[   82s] 
[   82s]     def test_etag_sync(tmpdir):
[   82s]         fpath = os.path.join(str(tmpdir), 'foo')
[   82s]     
[   82s]         file_ = open(fpath, 'w')
[   82s]         file_.write('foo')
[   82s]         file_.close()
[   82s]         os.sync()
[   82s]         old_etag = vdir.get_etag_from_file(fpath)
[   82s]     
[   82s]         file_ = open(fpath, 'w')
[   82s]         file_.write('foo')
[   82s]         file_.close()
[   82s]     
[   82s]         new_etag = vdir.get_etag_from_file(fpath)
[   82s]     
[   82s] >       assert old_etag != new_etag
[   82s] E       AssertionError: assert '1642549344080000000.000000000' != '1642549344080000000.000000000'
[   82s] 
[   82s] tests/vdir_test.py:68: AssertionError
```

dirkmueller:
In virtualized enviroments, kvmclock provides a slightly lower time resolution than the host default tsc and therefore this test can fail on fast hosts where sync() is not adding enough wait time.

### DIFF
--- a/tests/vdir_test.py
+++ b/tests/vdir_test.py
@@ -66,7 +66,12 @@ def test_etag_sync(tmpdir):
 
     new_etag = vdir.get_etag_from_file(fpath)
 
-    assert old_etag != new_etag
+    try:
+        assert old_etag != new_etag
+    except AssertionError:
+        pytest.xfail(
+            "Do we need to sleep?"
+        )
 
 
 def test_etag_sleep(tmpdir, sleep_time):
--- a/tests/vdir_test.py
+++ b/tests/vdir_test.py
@@ -27,8 +27,10 @@ import pytest
 from khal.khalendar import vdir
 
 
-@pytest.mark.xfail
-def test_etag(tmpdir):
+def test_etag(tmpdir, sleep_time):
+    if sleep_time > 0.01:
+        pytest.xfail("This environment requires sleeping")
+
     fpath = os.path.join(str(tmpdir), 'foo')
 
     file_ = open(fpath, 'w')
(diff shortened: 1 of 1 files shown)


d7415 (approved):

WhyNotHugo:
This changes makes the whole tests meaningless: if it fails, it's just treated as an xfail.


Add a `sleep(sleep_time)` instead: that will sleep the right amount based on the measured granularity of the filesystem's timestamps.

d7415:
> Add a `sleep(sleep_time)` instead: that will sleep the right amount based on the measured granularity of the filesystem's timestamps.

If we do that, I can't see how this test can possibly fail if the one below passes, as the `os.sync()` will be the only difference between them. In that case surely we may as well remove this test (and the one above) altogether.

dirkmueller:
> This changes makes the whole tests meaningless: if it fails, it's just treated as an xfail.

its a copy of the test before which is also doing xfail. I agree it is not a good solution. a better solution could be to make it only xfail when there is no nanosec time precision available because it is virtualized. 

if thats preferred I can try to come up with a condition to detect that.

d7415:
> if thats preferred I can try to come up with a condition to detect that.

I guess you could probably use `sleep_time` for that? So if sleep_time is above a certain threshold, xfail (with a note about sleep), otherwise fail?

I'm still ok with this version (the sleep is tested separately anyway with a hard fail) but see what @WhyNotHugo thinks.

dirkmueller:
I've added a patch in that regard.

WhyNotHugo (approved):
This approach LGTM.

Issue #1062, opened 2021-05-30
zsh completion is mistaken


WhyNotHugo:
```
$ khal list --<tab>
--days              -- specify how many days to include
--events            -- specify how many events to include
--exclude-calendar  -- don't use this calendar
--help              -- show help information
--include-calendar  -- specify calendar to use
$ khal list --days 20
Usage: khal list [OPTIONS] [DATETIME [DATETIME | RANGE]]
Try 'khal list --help' for help.

Error: No such option: --days

sdx23:
I began taking on this by manually editing `misc/__khal`, realized it's a quite boring task, considered using https://github.com/RobSis/zsh-completion-generator and finally thought one should fix this "upstream" i.e. in click.

Turns out click can automatically generate completions: https://github.com/pallets/click/tree/main/examples/completion
The result somewhat works. It can complete subcommands and options, but not both; neither does it do parameter for subcommands. Possibly this may be improved somewhat by adding decorators and/or extending existing ones.

However, I'm not sure whether that's the way to go or we should rather write an up-to-date __khal. The optimal solution imho would be if click could generate a _khal definition that doesn't rely on repeated calls to khal.

WhyNotHugo:
I'd prefer to do as much as possible using what upstream has to offer, and eventually submit improvements there.

Having custom made completion adds a bit of maintenance burden. Manually made completion needs to be updated every time something changes, whereas something generated by click should update itself.

sdx23:
I agree. However, I can estimate neither how difficult the required changes upstream to click would be nor whether they're possible at all.

To clarify what I meant by "that doesn't rely on repeated calls to khal", the attachement "khal1.txt" is what the first invocation `_KHAL_COMPLETE=zsh_source khal` outputs. The contained `env COMP_WORDS="${words[*]}" COMP_CWORD=$((CURRENT-1)) _KHAL_COMPLETE=zsh_complete khal` is another call to khal which outputs "khal2.txt". Due to the structuring of khal in commands and subcommands with their very own options this may become rather suboptimal from an efficiency point of view.

[khal2.txt](https://github.com/pimutils/khal/files/6786792/khal2.txt)
[khal1.txt](https://github.com/pimutils/khal/files/6786794/khal1.txt)

WhyNotHugo:
I understand your point, but don't have enough understanding of autocompletion mechanics to comment further.

Can we get away with very few calls to `khal` using click's completion framework?

sdx23:
[...]

Issue #937, opened 2020-05-16
Add auto-completion for other shells


d7415:
Thanks for responding. I have one more suggestion to add auto-completion as a feature. I develop a python cli tool as my day job and there's an argcomplete module that, when combined with argparse, allows for easy and complete argument completion. Is this issue board a good place for feature requests? I don't want to spam.

_Originally posted by @ciminobo in https://github.com/pimutils/khal/issues/935#issuecomment-629563251_

@ephase [responded](https://github.com/pimutils/khal/issues/935#issuecomment-629632311):
> ... khal use click instead of argparse. I know there is auto-completion for zsh because I use it everyday.

d7415:
The zsh completion is [here](https://github.com/pimutils/khal/blob/master/misc/__khal)

geier:
closed with #1141

PR #1141, opened 2022-06-24
Docs on generating and installing shell completion

### LINKED ISSUE #1034: Install shell completion from khal
Shell completion files for khal are not shipped by default with any shell (as far as I know). We could bring those files and install from within khal. 

For ideas on how to do it, see [typer's implementation](https://github.com/tiangolo/typer/blob/c3a4c72a4f1a18b7f16addb508862ec63411fa63/tests/test_completion/test_completion_install.py) (license is the same, so even code could be copied).

geier:
This isn't perfect as the click generated completion files cannot
complete subcommands and options at the same time and also not
parameters for subcommands, but it's still better that what we have now.

fixes #1034 #1062 #937


geier:
Perhaps this also should be moved to 0.11 as we break compatibility for packagers with this PR.

WhyNotHugo:
> @WhyNotHugo we might want to add similar information to vdirsyncer.

I've created an issue for it, it would definitely be very useful. https://github.com/pimutils/khal/pull/1141

PR #1148, opened 2022-06-25
Update .gitignore

abhinav1912:
Added common virtual env dirnames to .gitignore, in case the env is creating inside the root directory.


geier (approved):

geier:
Thanks!

Issue #1111, opened 2022-02-09
`readonly` in `discover` type not working


mstmob:
I have the following minimal config:
```
[calendars]

[[calendar_local]]
path = ~/.local/share/calendars/*
type = discover
readonly = True
```

`khal -v DEBUG -l ./khal.log` gives me the following debug log:
```
Using config:
[calendars]
  [[cal1]]
    path: /home/me/.local/share/calendars/cal1
    color: None
    type: calendar
    readonly: False
    priority: 10
  [[cal2]]
    path: /home/me/.local/share/calendars/cal2
    color: None
    type: calendar
    readonly: False
    priority: 10
```

I would expect `readonly: True` for both discovered calendars.
The `readonly = True` option seems to be ignored in the config for `type = discover`.

PR #1152, opened 2022-06-25
Fix/1111

geier:


geier:
We really need to test for this. Unfortunately our test setup currently doesn't actually set up some vdirs to be discovered. If those are set up, we can also fix #1111 (there is already a PR #1152 which I believe fixes the issue, but breaks the tests).

PR #859, opened 2019-03-30
Fix event creation for future events

geier:
Events too far into the future (after the second next DST transition)
used to be created with invalid timezones.

fix #817

Issue #817, opened 2018-10-20
Events that start after beginning of DST contain no STANDARD or DAYLIGHT information in VTIMEZONE field


Pieter1024:
I am using khal 0.9.10. When I create an event that starts after the start of DST in 2019, I run into problems when syncing it to CalDAVd via vdirsyncer. CalDAVd reports that the calendar data has unfixable problems: "[VTIMEZONE] At least one component must be present: STANDARD or DAYLIGHT".

I can reproduce this by creating one event right before the start of DST (say, the evening of 30 March 2019) and one right after the start of DST (say, the evening of 31 March 2019).

These are the .ics files that ikhal creates (also occurs with khal):

before start of DST:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//PIMUTILS.ORG//NONSGML khal / icalendar //EN
BEGIN:VTIMEZONE
TZID:Europe/Amsterdam
BEGIN:STANDARD
DTSTART;VALUE=DATE-TIME:20181028T020000
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
END:STANDARD
END:VTIMEZONE
BEGIN:VEVENT
SUMMARY:abc123
DTSTART;TZID=Europe/Amsterdam;VALUE=DATE-TIME:20190330T190000
DTEND;TZID=Europe/Amsterdam;VALUE=DATE-TIME:20190330T200000
DTSTAMP;VALUE=DATE-TIME:20181020T194957Z
UID:S1RCXZ00XWW6YED07S7S9Z1LM2Q63DD88SQE
SEQUENCE:0
END:VEVENT
END:VCALENDAR
```

after start of DST:

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//PIMUTILS.ORG//NONSGML khal / icalendar //EN
BEGIN:VTIMEZONE
TZID:Europe/Amsterdam
END:VTIMEZONE
BEGIN:VEVENT
SUMMARY:abc123
DTSTART;TZID=Europe/Amsterdam;VALUE=DATE-TIME:20190331T190000
DTEND;TZID=Europe/Amsterdam;VALUE=DATE-TIME:20190331T200000
DTSTAMP;VALUE=DATE-TIME:20181020T195008Z
UID:07H39ZOVC4AIOGGY0VSE71Q7SMKT5MDALSX0
SEQUENCE:0
END:VEVENT
END:VCALENDAR
```

geier:
[...]

PR #1154, opened 2022-06-26
Tests for VTIMEZONE creation

geier:
More tests for #817 / #859.

As this introduces a new test dependency, we should only merge this after the 0.11.0 release.

Events that are too far into the future (after the second next DST
transition) are created with invalid VTIMEZONES.

Introduces a new test dependency: hypothesis

Reproduces #817

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ doc/source/configspec.rst
 .venv
 env/
 venv/
+.hypothesis/
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ requirements = [
 
 test_requirements = [
     'freezegun',
+    'hypothesis',
     'vdirsyncer',
 ]
 
--- a/tests/event_test.py
+++ b/tests/event_test.py
@@ -3,6 +3,8 @@ import datetime as dt
 import pytest
 import pytz
 from freezegun import freeze_time
+from hypothesis import event, given
+from hypothesis.strategies import datetimes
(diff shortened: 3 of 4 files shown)


WhyNotHugo:
@gebner Will these non-deterministic tests be a problem for packaging on Nix?

The resulting build remains deterministic; only test output changes.

geier:
> @gebner Will these non-deterministic tests be a problem for packaging on Nix?
> 
> The resulting build remains deterministic; only test output changes.

I don't think so, isn't vdirsyncer using hypothesis as well? I also believe we are generating random uuids somewhere.

WhyNotHugo:
We do, and we have a flag to make tests deterministic:

https://github.com/pimutils/vdirsyncer/blob/87f3a594c64c84635c7e566e6d3573403379680d/docs/packaging.rst#testing

https://github.com/pimutils/vdirsyncer/blob/ec101b20d6abdcf8c1aa4adb41d4aebe6867d5ed/tests/conftest.py#L40-L50

https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/vdirsyncer/default.nix#L60

gebner:
It's only a problem insofar as that the tests become flaky.  If a test fails randomly (and non-deterministic tests have that tendency), then we'll have to disable it.

WhyNotHugo:
Thanks for the feedback!

WhyNotHugo (approved):

geier:
Those tests really shouldn't be inconsistent, but if you find some failing tests, please open an issue.

PR #1156, opened 2022-06-26
Remove reference to broken file

WhyNotHugo:
This reference makes builds fail. How did this not come up on CI?


WhyNotHugo:
Tox should be installing the package on CI. But this fails when building the AUR's `-git` package.

d7415 (approved):

WhyNotHugo:
Actually, the failure on the AUR package is because the package referred to this file too.

The manifest is not used by the AUR or CI, apparently...?

PR #1157, opened 2022-06-26
Drop unused MANIFEST

WhyNotHugo:
We're using `setuptools_scm`, so git-tracked files are included
automatically. The manifest is unused.

See: https://github.com/pimutils/khal/pull/1156


WhyNotHugo:
Leaving in draft; I need to double-check that the generated package contains the same files.

Issue #1180, opened 2022-08-27
Test failure on recent pytz version


icp1994:
`pytz_version()` defined under [conftest](https://github.com/pimutils/khal/blob/v0.10.5/tests/conftest.py#L109-L113) results in ValueError with `pytz 2022.2.1` since the split now returns 3 values.

geier:
@icp1994  thank you for reporting this!

PR #1181, opened 2022-08-30
Support for latest version of pytz

geier:
We had assumed that pytz versions would always follow the schema
year.month, but pytz 2022.2.1 broke that assumption and our code. We
should now be able to deal with pytz versions in the
year.month.anything.

fix https://github.com/pimutils/khal/issues/1180

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ not released
 
 * DROPPED support for python versions < 3.8
 * UPDATED REQUIREMENT pytz is now required >= 2018.7
+* NEW test REQUIREMENT: packaging
+* FIX support in tests for pytz version numbers of the format year.month.minor
 * FIX deleting of instances of recurring events in ikhal
 * FIX if a `discover` collection is set to "readonly", discovered collections
   will now inherit the readonly property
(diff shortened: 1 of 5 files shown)


geier:
For now just trying to reproduce #1180

WhyNotHugo:
LGTM.

You can also use [`packaging.version.parse`](https://packaging.pypa.io/en/latest/version.html#packaging.version.parse). `packaging` is already a [transitive] test dependency.

WhyNotHugo (approved):

geier:
Didn't know about packaging, thanks!

But I'm hesitant to include without a major version bump, so I guess I'll put this release on a new v0.10.x branch and create a new PR using packaging for master.

CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ not released
 
 * DROPPED support for python versions < 3.8
 * UPDATED REQUIREMENT pytz is now required >= 2018.7
+* NEW test REQUIREMENT: packaging
```
WhyNotHugo, line 16: This is already required by a test dependency, so it's not strictly a "new" requirement.

WhyNotHugo (approved):

PR #1185, opened 2022-09-11
Make test_etag_sync more reliable

bmwiedemann:
Make `test_etag_sync` more reliable
with filesystems that only have millisecond-accuracy

### DIFF
--- a/tests/vdir_test.py
+++ b/tests/vdir_test.py
@@ -59,6 +59,7 @@ def test_etag_sync(tmpdir, sleep_time):
     file_.close()
     os.sync()
     old_etag = vdir.get_etag_from_file(fpath)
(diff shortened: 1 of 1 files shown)


d7415:
Your example is effectively tested a few lines down in `test_etag_sleep`. If you look a few lines above, there's a reference to `sleep_time`, which is calculated [here](https://github.com/pimutils/khal/blob/master/tests/conftest.py#L109).

If your system is failing (rather than `xfail`ing) the test you've modified, this threshold may need to be revisited.

bmwiedemann:
Ah, it might be that after 2038-01-19, floating point numbers lose one bit of accuracy that breaks that fine balance.

WhyNotHugo:
Indeed, use `sleep_time` instead, which sleep the minimum amount of time required for the current filesystem.

PR #1198, opened 2022-10-22
Do not evaluate etag when deleting instances

t2b3:
If multiple instances of an event are deleted one after the other, the event will be updated each time and thus its etag will change dynamically each time. This will lead to a mismatch with the static etag recorded at the time when the instances are queued for deletion.

Instead of re-adjusting the deletion queue each time, do not compare the dynamic etag with the static etag. The etag for the event will be updated anyway.


geier:
Thank you for sending this PR, and sorry for not having a look until now. I believe removing those etag checks is a mistake, or at least we should then decide that editing events in khal always wins and overwrites whatever currently is on disk. I have instead created #1273, which checks etags for subsequent deletions.

PR #1211, opened 2022-12-08
implementation  of issues 535 and 809

ZakariaSoukrat:
we explained our changes in the file CHANGELOG.rst


geier:
Thank you for you pr and your interest in khal. Sorry about not getting back to you any sooner. 

I believe this functionality would be a great fit for a plugin, where we could let the code mature and see if there is enough interest in than later move it into khal proper. 

Have a look https://github.com/pimutils/khal/pull/1309#issuecomment-1783855447 on how to possibly migrating this to a plugin. Please get back to us, if you need any help with that.

PR #1212, opened 2022-12-10
Implementation for issue #809 and general display improvements

GabrielDum:
We used an awful lot of try/excepts in this PR, so if you have a better way of doing it you're welcome ! Anyway this PR might not be perfect but proposes a new display for both khal and ikhal where months are visually separated. The ideas in came from @aleprovencio and @WhyNotHugo. I did it with a coworker, so i added him as well to the AUTHORS file. There might be some useless commits, just look at the final files. We won't be working on it anymore, so if it isn't receivable but you like the idea, feel free to adapt and improve it !


geier:
Thanks for your PR, and sorry for not taking a look at this earlier.

Personally, I'm not a fan of visual separation between different months (the first `khal` script I wrote because I didn't like the output of `cal` or `ncal`). If you make this configurable and not the default behavior, I'm not against merging it though.

geier:
Apart from that, there seem to be some problem with the current state:

image

image

geier:
Perhaps we can also find a way to make the separation work as a [plugin](https://github.com/pimutils/khal/issues/1258) first and let it mature outside of the main code base.

geier (changes requested):
see comments above

geier:
closed for now, please re-open if you are still interested in working on this.

PR #1213, opened 2022-12-10
Import only xdg from settings.py

Andonome:
The line `import xdg.BaseDirectory` can fail if there is no such module.

I'm using khal, version 0.10.5, from the Arch community repos.

The change doesn't seem notable enough for extra changes in `AUTHORS.txt`, or `CHANGELOG.rst`.

### DIFF
--- a/khal/settings/settings.py
+++ b/khal/settings/settings.py
@@ -23,7 +23,7 @@
 import logging
 import os
 
-import xdg.BaseDirectory
+import xdg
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
What issue does this fix? `khal list` with khal from the Arch repos works fine for me as-is.

Andonome:
My mistake. My problem was caused by a local pip installation of xdg. Fixed with `pip uninstall xdg`.

PR #1220, opened 2023-01-26
Added Support for Microsoft Busystatus

sebu06:
I added support for the ics Attribute X-MICROSOFT-BUSYSTATUS. Since this is my fist contribution to the ikhal project, I would greatly appreciate feedback on this one.

Kind regards,
Sebastian

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -300,6 +300,14 @@ class Event:
             }
 
     @property
+    def allowed_busystatus(self) -> Dict[str, str]:
+        return {'BUSY': 'Busy',
+                'FREE': 'Free',
+                'TENTATIVE': 'Tentative',
+                'WORKINGELSEWHERE': 'Work elsewhere',
+                'OOF': 'Out of office'}
+
+    @property
     def start_local(self) -> dt.datetime:
         """self.start() localized to local timezone"""
         return self.start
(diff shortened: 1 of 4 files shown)


sebu06:
this should close pull request #989 as well...

sebu06:
I'm using this with davmail. However, I just found out, that it only displays the status correctly, and the update only sometimes works - due to davmail doing some processing on their end as well. So if this should be fully supported, a patched version of davmail is required as well... For displaying it works well though

khal/khalendar/event.py
```diff
@@ -299,6 +299,14 @@ def symbol_strings(self) -> Dict[str, str]:
                 'right_arrow': '->'
             }
 
+    @property
+    def allowed_busystatus(self) -> Dict[str, str]:
+        return {'BUSY': 'Busy',
+                'FREE': 'Free',
+                'TENTATIVE': 'Tentative',
+                'WORKINGELSEWHERE': 'Work elsewhere',
+                'OOF': 'Out of office'}
```
WhyNotHugo, line 308: We don't need to define this dictionary inside a function, just make it a global dictionary. Otherwise it gets recreated each time the function is called, which is kinda pointless.

khal/khalendar/event.py
```diff
@@ -343,6 +351,27 @@ def organizer(self) -> str:
         else:
             return email
 
+    @property
+    def busystatus(self) -> str:
+        if 'X-MICROSOFT-CDO-BUSYSTATUS' not in self._vevents[self.ref]:
+            return 'BUSY'
```
WhyNotHugo, line 357: If this property is not set, I don't think we should fall back to `BUSY`, we should return `None` instead. Especially since this property is non-standard.

khal/khalendar/event.py
```diff
@@ -343,6 +351,27 @@ def organizer(self) -> str:
...
+        return self._vevents[self.ref]['X-MICROSOFT-CDO-BUSYSTATUS']
+
+    @property
+    def busystatus_formatted(self) -> str:
+        if 'X-MICROSOFT-CDO-BUSYSTATUS' not in self._vevents[self.ref]:
+            return ''
+        busystatus = self._vevents[self.ref]['X-MICROSOFT-CDO-BUSYSTATUS']
+        if busystatus not in self.allowed_busystatus.keys():
+            return ''
+        return self.allowed_busystatus[busystatus]
```
WhyNotHugo, line 367: **Non blocking hint**: You can use `self._vevents[self.ref].get('X-MICROSOFT-CDO-BUSYSTATUS', None)`, which returns the value (if present) or `None` (if absent).

```suggestion
        busystatus = self._vevents[self.ref].get('X-MICROSOFT-CDO-BUSYSTATUS', None)
        if busystatus not in self.allowed_busystatus.keys():
            return ''
        return self.allowed_busystatus[busystatus]
```

Which can be further simplified into:


```suggestion
        busystatus = self._vevents[self.ref].get('X-MICROSOFT-CDO-BUSYSTATUS', None)
        return self.allowed_busystatus.get(busystatus, ""):
```

WhyNotHugo (commented):
I haven't had a chance to properly test this yet, but overall seems to look okay.

sebu06:
> Any idea what they're doing?

Yes, they seem to do the following:

on update:
set the busystatus to STATUS entry in VEVENT

on event creation:
1. set status to Tentative if STATUS:TENTATIVE
2. otherwise set status to Busy is X-MICROSOFT-CDO-BUSYSTATUS:BUSY
3. if both is not true, set status to Free

I'm currently thinking there are two options: only supprort setting the status to Busy / Free / Tentative (as it seems to be the standard) and only display the MS-Status for events that have been added externally or modify davmail. I'm leaning towards the first option, since the latter would most likely lead to problems with other clients (like Dav5x) when they do not follow the MS standard.

sebu06:
Thanks for all the feedback, I will update it =)

PR #1221, opened 2023-02-10
Use Ruff for code linting and checking

WhyNotHugo:
[Ruff](https://github.com/charliermarsh/ruff) does the same as flake8(+plugins)+pyupgrade+isort, but much faster and can also automatically fix many of these quirks too.

It ignores all files that are ignored by git, so all the exclude rules are not necessary.

I have left isort-related rules out for now since they results in some changes which might add noise at this point.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,19 +13,6 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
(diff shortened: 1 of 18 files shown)


WhyNotHugo:
Darn, `isort` seems to be broken when run via pre-commit. It might be a good time to replace it too TBH.

PR #1225, opened 2023-02-14
Fix recently broken tests

WhyNotHugo:


d7415:
I think we discussed previously, but if you're adding `sleep` to those tests, you may as well remove `test_etag_sleep` as it becomes entirely redundant.

WhyNotHugo:
@d7415 Do you want to send a patch for that?

Issue #1243, opened 2023-04-28
khal at ignores event_format


heull001:
khal at is only formatable via --format, the entry view/event_format in the config-file is not used.

The documentation doesn't clearly state this is used here, but it sounds like.

> Default formatting for events used when the start- and end-date are not clear through context, e.g. for search, used almost everywhere but list and calendar.                                                                                    

Version: 0.11.1

heull001:
Oh, now I see, it uses agenda_format instead. I thought, it would not use event_format nor agenda_format. I still think, at should use event_format, because unlike to list no start and end dates are displayed, but perhaps a discussion is needed before merging this.

geier:
Agreed.

geier:
fixed with #1244

PR #1244, opened 2023-04-28
Fix for #1243

heull001:
#1243 the at-commant ignores view/event_format in the config-file


WhyNotHugo (approved):
I think this is good.

geier:
Thanks!

PR #1256, opened 2023-05-23
more warnings on failing to import events

geier:


khal/icalendar.py
```diff
@@ -72,8 +72,18 @@ def split_ics(ics: str, random_uid: bool=False, default_timezone=None):
...
+    saved_exception = None
+    for uid, events in sorted(events_grouped.items()):
+        try:
+            ics = ics_from_list(events, tzs, random_uid, default_timezone)
+        except Exception as exception:
+            logger.warn(f'Error when trying to import the event {uid}')
+            saved_exception = exception
+        out.append(ics)
+    if saved_exception:
+        raise saved_exception
```
WhyNotHugo, line 85: This will only raise the last exception (which I guess is fine?).

Maybe we can use an [`ExceptionGroup`](https://docs.python.org/3/library/exceptions.html#exception-groups) in a few years when Python 3.11 has trickled down to most distributions.

d7415: Yeah, the last is probably enough. If we wanted more detail, we could include `{exception}` or `{exception!r}` in the warning?

khal/icalendar.py
```diff
@@ -72,8 +72,18 @@ def split_ics(ics: str, random_uid: bool=False, default_timezone=None):
...
-            sorted(events_grouped.items())]
+    out = []
+    saved_exception = None
+    for uid, events in sorted(events_grouped.items()):
+        try:
+            ics = ics_from_list(events, tzs, random_uid, default_timezone)
+        except Exception as exception:
+            logger.warn(f'Error when trying to import the event {uid}')
+            saved_exception = exception
+        out.append(ics)
```
d7415, line 83:
```suggestion
        else:
            out.append(ics)
```

d7415:
> I think you can include `exc_info=True` and the warning will include a stacktrace. It should be useful, but maybe TOO verbose.

I'd forgotten about that one, but yeah, multiple stack traces might be a bit much. That's why I was thinking of a one-liner. That said, at least the event ID is there so it can always be attempted again to get more details.

Issue #1217, opened 2023-01-22
Crash when importing event with invalid VALARM/TRIGGER


WhyNotHugo:
I tried importing this file, which is generated by [NS International](https://www.nsinternational.com/en/traintickets-v3/).

```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//NS Internationaal B.V.//sales-v2//en
BEGIN:VTIMEZONE
TZID:Europe/Brussels
LAST-MODIFIED:20221105T024525Z
TZURL:http://www.tzurl.org/zoneinfo/Europe/Brussels
X-LIC-LOCATION:Europe/Brussels
X-PROLEPTIC-TZNAME:LMT
BEGIN:STANDARD
TZNAME:BMT
TZOFFSETFROM:+0017
TZOFFSETTO:+0017
DTSTART:18800101T000000
END:STANDARD
BEGIN:STANDARD
TZNAME:WET
TZOFFSETFROM:+0017
TZOFFSETTO:+0000
DTSTART:18920501T001730
END:STANDARD
BEGIN:STANDARD
TZNAME:CET
TZOFFSETFROM:+0000
TZOFFSETTO:+0100
DTSTART:19141108T000000
END:STANDARD
BEGIN:STANDARD
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
DTSTART:19161001T010000
RDATE:19421102T030000
RDATE:19431004T030000
RDATE:19440917T030000
RDATE:19450916T030000
RDATE:19461007T030000
RDATE:19770925T030000
RDATE:19781001T030000
END:STANDARD
BEGIN:STANDARD
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
DTSTART:19170917T030000
RRULE:FREQ=YEARLY;UNTIL=19180916T010000Z;BYDAY=3MO;BYMONTH=9
END:STANDARD
BEGIN:STANDARD
TZNAME:WET
TZOFFSETFROM:+0100
TZOFFSETTO:+0000
DTSTART:19181111T120000
RDATE:19191005T000000
RDATE:19201024T000000
RDATE:19211026T000000
RDATE:19391119T030000
END:STANDARD
BEGIN:STANDARD
TZNAME:WET
TZOFFSETFROM:+0100
TZOFFSETTO:+0000
DTSTART:19221008T000000
[...]

PR #1259, opened 2023-05-25
Raise error properly on failure to parse ics

magarnicle:
If cal_from_ics in icalendar.py raises a ValueError that is not to do with timezone issue, it will raise it instead of ignoring it and then generating "UnboundLocalError: local variable 'cal' referenced before assignment"


geier:
Good catch, thanks! Could you please add yourself to AUTHORS.txt, and -- if you have -- provide an example .ics file that triggers the issue? (I'd like to write a test to cover that case.)

magarnicle:
Sorry, should have read the contribution guildelines. I've added a test and the .ics from issue #1217

geier:
Awesome, thanks!

PR #1260, opened 2023-05-26
Fix duplicate event from ikhal

tretter:
Commit f7fd70bb ("CalendarCollection.new() -> CalendarCollection.insert()") broke duplicating an event from ikhal using "p". Follow the function rename to fix it.


geier:
Thanks for fixing this! Could you add yourself to the AUTHORS.txt file? Ideally also create a new entry in the CHANGELOG, but I can do that, too.

geier:
Thanks!

geier (approved):

PR #1261, opened 2023-05-27
Implement minimum month display configuration

### LINKED ISSUE #1226: month count setting for calendar command
I have been trying to use calendar command to display the calendar of the current month but the default is 3 months [hardcoded](https://github.com/pimutils/khal/blob/395a396d1396fa08cdb346d190a9a8a8a5ee9885/khal/controllers.py#L119).

I was thinking about making it accessible through configuration. A new option called min_calendar_display under view section.

It could be useful for #800 althought I don't really know how much of an impact it could have on ikhal.

Any thoughts about this?

r4ulill0:
fix #1226


geier:
Thanks!

Issue #580, opened 2017-02-17
Ability to mark/hide declined events


hobarrera:
I currently have some work events which I've declined, and are indistinguishable from accepted events.

I'd like to somehow mark (an icon similar to recurring events), hide, stike-through (this is my first choice, but very few terminals support it, including my own) or dim (eg: slightly darker colour) events which I've declined.

As far as I understand, this also requires khal to know my email[s] (because that's how attendees are identified).

IMHO, dimming or some special character to recognize these are the best choices.

geier:
Good idea, once khal know about your emails, it can also answer(RSVP) and send them (invite).

I have actually yesterday implemented something similar for `STATUS=CANCELLED` events, which for now are displayed with a big, fat CANCELLED in front, which is indeed very ugly.



Your suggestions

* strikethrough: I like this best, and my *terminal* supports this, but tmux which I often use to run khal etc in does not (and it looks like [they are not interested to](https://github.com/tmux/tmux/issues/612)), so this is probably out (to test run ``echo -e "\e[9mstrikethrough\e[0m" ``)

* dimmed: I find dimmed (dark gray for dark background, light gray for light background) problematic, because it often leads to me not being able to read that part, than I need more attention to actually read it (which kind of defeats the purpose). But perhaps I would just get used to ignoring unreadable stuff in khal.

* symbol: 
  1. do you have a suggestion for a symbol?
  2. is that noticable enough? (also for cancelled events)

geier:
An alternative to ANSI codes for strikethrough would be [unicode strikethrough](https://en.wikipedia.org/wiki/Strikethrough#Combining_characters).

Seems to work for me:

```python
In [1]: STRIKE = '\u0336'

In [2]: def strikethrough(string):
   ...:     new = list()
   ...:     for char in string:
   ...:         new.append(STRIKE)
   ...:         new.append(char)
   ...:     return ''.join(new)
   ...: 

In [3]: print(strikethrough('strikken'))
s̶t̶r̶i̶k̶k̶e̶n
```

geier:
font support for that method is only so and so:

DejaVu Sans Mono:
![dejavusans](https://cloud.githubusercontent.com/assets/275330/23093329/16cd40de-f5d8-11e6-96ab-6034cbfaa6ed.png)

M+1mn:
![m 1mn](https://cloud.githubusercontent.com/assets/275330/23093332/1e34840e-f5d8-11e6-9d62-cb74f5565ed0.png)

Iosevka:
[...]

PR #1271, opened 2023-06-15
Better support for status and initial support for partstat

geier:


madduck:
I can confirm that this works as intended. Will take discussion back to #580 though.

khal/khalendar/event.py
```diff
@@ -556,6 +568,33 @@ def _alarm_str(self) -> str:
             alarmstr = ''
         return alarmstr
 
+    @property
+    def _status_str(self) -> str:
+        if self.status == 'CANCELLED':
+            statusstr = ' ' + self.symbol_strings['cancelled']
```
madduck, line 574: I'd like to make a case against adding a leading space, or at least to make this configurable. Such a leading space should really be part of the format specification, though I realise it's currently not possible to work with conditionals there.

However, `{status: <1}` will probably suit anyone better when the status symbol is to be used mid-field, whereas adding all the statuses to the end of the line won't make a difference anyway.

geier: fixed

geier:
config file now takes several addresses per account, but we expect that only one will match.

geier:
@WhyNotHugo sorry for bombing you with review requests, feel free to ignore them.

geier:
This PR doesn't cover yet formatting events by their status as it's a bit more difficult to get this looking the same in khal and ikhal. I'd prefer to merge this without that now, because rebasing against master is becoming tiresome (not especially this branch, but keeping several non-merged features branches around is getting tiring).

WhyNotHugo (commented):
If I understand correctly, [the `CALDAV:calendar-user-address-set` property](https://datatracker.ietf.org/doc/html/rfc6638#section-2.4.1) allows querying the address of the owner of a calendar.

This should enable fetching the email address of the calendar owner via vdirsyncer.

Perhaps treating this as calendar metadata might be more future compatible?

geier:
@WhyNotHugo interesting. We should definitively support that, but probably only as a default. At least one of my caldav servers has an email address set that nobody sends email to.

PR #1275, opened 2023-06-20
Use vermin to check Python-compat

WhyNotHugo:
Vermin checks source code for anything that might be incompatible with older Python versions.

This prevents accidentally using a feature from Python 3.10 when we're still supporting 3.8.

While tests ideally detect this, there will always be some bits not covered by tests.

Using a non-release revision since we need some unreleased changes to deal with zoneinfo backports.

See: https://github.com/netromdk/vermin/issues/189

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,8 @@ repos:
     hooks:
       - id: ruff
         args: ["--fix"]
+  - repo: https://github.com/netromdk/vermin
+    rev: 58fe25f0a3bdaeefe7a2a021bee2938f5465ccde
+    hooks:
+      - id: vermin
+        args: ['-t=3.8-', '--violations']


geier:
Nice.

Seems to work https://github.com/pimutils/khal/pull/1276

doesn't catch `def test(a: int | str)` but does catch `'test'.removeprefix()`. Agree, might be helpful.

netromdk:
Hi there. Vermin author here.

> doesn't catch `def test(a: int | str)` but does catch `'test'.removeprefix()`. Agree, might be helpful.

If you are using union types, then you must explicitly enable the detection via `--feature union-types`. As you can read in the [caveats section](https://github.com/netromdk/vermin#caveats):

> Detecting union types (X | Y [PEP 604](https://www.python.org/dev/peps/pep-0604/)) can be tricky because Vermin doesn't know all underlying details of constants and types since it parses and traverses the AST. For this reason, heuristics are employed and this can sometimes yield incorrect results (https://github.com/netromdk/vermin/issues/103). To enable (unstable) union types detection, use --feature union-types.

If you have:
```py
def test(a: int | str):
  pass
```

Then you will get the following using `--feature union-types`:
```
!2, 3.10     /tmp/test.py
  L1: annotations require !2, 3.0
  L1: union types as `X | Y` require !2, 3.10
```

geier:
Thanks @netromdk for the information!

I was testing basically testing `test(a: int | str)` because it was the first feature I could think of that python 3.8 doesn't support. I don't think we need to enable that for now.

netromdk:
Sure thing! :)

PR #1279, opened 2023-06-23
ikhal: initial support for config based ikhal theming

### LINKED ISSUE #1219: Change blue background
How can I change the blue background to another color?

image

By reading the docs it looks like I have to change source files?

What is the right option I have to change?

geier:
This PR allows to override ikhal's color theme with a custom palette. This is useful to style certain elements of ikhal individually.

Palette entries take the form of `key = foreground, background, mono, foreground_high, background_high` where foreground and background are used in "low color mode"  and foreground_high and background_high are used in "high
 color mode" and mono if only monocolor is supported. If you don't want to set a value for a certain color, use an empty string (`''`). Valid entries for low color mode are listed on the [urwid website](http://urwid.org/manual/displayattributes.html#standard-foreground-colors). For high color mode you can use any valid 24-bit color value, e.g. `'#ff0000'` (Note: 24-bit color values must be enclosed in quotes).

Most modern terminals should support high color mode. 

Example entry: `header = light red, default, default, '#ff0000', '#ff0011'`
See the default palettes in `khal/ui/colors.py` for all available keys. If you can't theme an element in ikhal, please open an issue on github.

Example config:
image

leads to this ugliness:
image

fixes #1219

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ not released yet
 * optimization in ikhal when editing events in the far future or past
 * FIX an issue in ikhal with updating the view of the event list after editing
   an event
+* NEW properties of ikhal themes (dark and light) can now be overriden from the
+  config file (via the new [palette] section, check the documenation)
 
 0.11.2
 ======
--- a/khal/settings/khal.spec
+++ b/khal/settings/khal.spec
@@ -324,3 +324,21 @@ multiple_on_overflow = boolean(default=False)
 # actually disables highlighting for events that should use the
 # default color.
 default_color = color(default='')
+
+# Override ikhal's color theme with a custom palette. This is useful to style
+# certain elements of ikhal individually.
+# Palette entries take the form of `key = foreground, background, mono,
+# foreground_high, background_high` where foreground and background are used in
+# "low color mode"  and foreground_high and background_high are used in "high
+# color mode" and mono if only monocolor is supported. If you don't want to set
+# a value for a certain color, use an empty string (`''`).
+# Valid entries for low color mode are listed on the urwid website_.
+# _http://urwid.org/manual/displayattributes.html#standard-foreground-colors For
+# high color mode you can use any valid 24-bit color value, e.g. `'#ff0000'`.
+# NOTE: 24-bit colors must be enclosed in single quotes to be parsed correctly,
+# otherwise the `#` will be interpreted as a comment.
+# Most modern terminals should support high color mode.
+# Example entry: `header = light red, default, default, #ff0000, default`
+# See the default palettes in `khal/ui/colors.py` for all available keys.
+# If you can't theme an element in ikhal, please open an issue on github.
+[palette]
--- a/khal/settings/khal.spec
+++ b/khal/settings/khal.spec
@@ -257,7 +257,7 @@ blank_line_before_day = boolean(default=False)
 # `khal/settings/khal.spec` in the section `[default]` of the property `theme`.
 #
 # __ http://urwid.org/manual/displayattributes.html
-# .. _github: # https://github.com/pimutils/khal/issues
+# .. _github: https://github.com/pimutils/khal/issues
 theme = option('dark', 'light', default='dark')
 
 # Whether to show a visible frame (with *box drawing* characters) around some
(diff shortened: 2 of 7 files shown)


tgy:
just wondering if it would be possible to source another config file within the config

many people use both a 'dark' and a 'light' theme (I do myself: catppuccin latte/frappe) and like being able to switch from one to another depending on the time of the day

with the proposed single file config, this requires having both themes within the config, commenting out one or the other depending on which theme should be applied

geier:
@tgy for that use case one could write a color theme plugin. I'll upload my first try tonight.

tgy:
could, we should add documentation for that! i can help

khal/api.py
tgy, line 1: what is this file?

geier:
@tgy see here https://github.com/geier/khal_gruvbox

Mostly useful for now for finding parts which are not theme-able yet 

image

khal/api.py
geier: the idea is, that this part of (plugin) api we will keep backwards compatible. The current approach of just loading any python module also allows accessing all of khal, but at least for now I would prefer not to make any big promises on backwards compatibilty.

geier:
> could, we should add documentation for that! i can help

we should, but only after we have properly worked out how this is going to work in the long run.

geier:
complication: the code highlighting days and events in the calendar's color remoes the foreground/background color.

tgy:
@geier could you share your part of the config with the color definitions you are testing?

i checked out your version of the code in my local environment and will help you test it!

geier:
@tgy it should be enough to follow the instructions here: https://github.com/geier/khal_gruvbox, if not, I forgot to check something in.

w-jablonski:
I am unsure of what is the current state of this, but I would really love the option to have weekends coloured differently than weekdays. Thanks.

khal/api.py
```diff
@@ -0,0 +1 @@
+from .ui.colors import register_color_theme  # noqa
```
WhyNotHugo, line 1: You can use the following to indicate that this module explicitly exports a function:

```py
__all__ = ["register_color_theme"]
```

Mainly, it'll prevent the "unused import" warning entirely while conveying some meaning to casual readers.

khal/cli.py
```diff
@@ -482,7 +484,7 @@ def isatty(_file):
     @multi_calendar_option
     @click.pass_context
     def interactive(ctx, include_calendar, exclude_calendar):
-        '''Interactive UI. Also launchable via `ikhal`.'''
+        '''Interactive UI. Also launchable via `khal interactive`.'''
```
WhyNotHugo, line 487: _"Also launchable via `khal interactive`"_ doesn't sound right as a description for `khal interactive`. This is currently shown as the output for `khal interactive --help`:

```console
> khal interactive --help
Usage: khal interactive [OPTIONS]

  Interactive UI. Also launchable via `ikhal`.
```

khal/cli.py
```diff
@@ -682,5 +690,18 @@ def configure(ctx):
...
+    # __init__.py file. If so, import them.
+    plugin_dir = os.path.join(xdg.BaseDirectory.xdg_data_home, 'khal', 'plugins')
+    if not os.path.isdir(plugin_dir):
+        return
+    sys.path.append(plugin_dir)
+    for plugin in os.listdir(plugin_dir):
+        if os.path.isfile(os.path.join(plugin_dir, plugin, '__init__.py')):
+            logger.debug(f'loading plugin {plugin}')
+            __import__(plugin)
+
```
WhyNotHugo, line 705: Nice and simple approach :+1:

WhyNotHugo (commented):
Only minor nitpicks from my end. Looks good.

WhyNotHugo:
Needs rebase

khal/cli.py
geier: @tgy you are right, I'm mixing several things in this PR, they only have in common that they are related to colors. I will split this into two PRs:
1. overriding (elements of) the color theme, will also make more elements themeable and slightly tweak the existing color themes. This one should be good to merge.
2. Plugin support for writing color theme plugins, this one is probably not ready yet.

khal/cli.py
geier: true, removed.

geier:
I have removed the color theme plugin part (and moved it to #1303).

khal/api.py
geier: Thanks, implemented

geier:
> I am unsure of what is the current state of this, but I would really love the option to have weekends coloured differently than weekdays. Thanks.

good point, tracked in https://github.com/pimutils/khal/issues/1304

Tokariew:
question i have.
How should be colors with spaces in name be edited? like `list focused`?

geier:
Did you try wrrapping them like this `"list focused"` or `'list focused'`?

PR #1282, opened 2023-06-26
Add .python-version to .gitignore

heull001:
.python-version is used by pyenv to specify the used python-version by directory. So this is very usefull for testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ doc/source/configspec.rst
 env/
 venv/
 .hypothesis/
+.python-version


WhyNotHugo:
I would suggest that you add this to `~/.git/config/ignore`, so that it gets ignored on any git repository you work on locally.

heull001:
That would work fine for me, but I thought, this might be helpfull for others
too, but if you don't think so, this is okay for me.

geier:
Thanks!

PR #1283, opened 2023-06-29
poc: format event html descriptions

tcuthbert:
HTML event descriptions are difficult to read, especially when originating from marketing or tech firms.

This PoC proposes transforming event description text through a filter that strips the tag soup and formats text for easier reading on a console displays.

An alternative to my PR would be for users to configure their own formatters, and have subprocess pipe the event descriptions into the process. For example, the user could configure `~/.config/khal/config` like so:

```
[view]
event_description_formatter = pandoc -f html -t plain
```

To my knowledge there's no way of knowing if an event description is going to be html, so another feature goal could be to have some kind of heuristics and only format if a number of html tags are detected.

This PoC is something I hacked together in an afternoon; as per your [suggestion](https://khal.readthedocs.io/en/latest/hacking.html#hacking), I didn't want to burn too much time on it in case you reject the feature proposal. Thank you for all the work on khal, I'd love to contribute back as its worked wonders for my smooth ADHD brain.

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -39,7 +39,7 @@ from ..exceptions import FatalError
 from ..icalendar import cal_from_ics, delete_instance, invalid_timezone
 from ..parse_datetime import timedelta2str
 from ..terminal import get_color
-from ..utils import generate_random_uid, format_text, is_aware, to_naive_utc, to_unix_time
+from ..utils import format_text, generate_random_uid, is_aware, to_naive_utc, to_unix_time
 
 logger = logging.getLogger('khal')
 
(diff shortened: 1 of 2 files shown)


geier:
Hi @tcuthbert, I agree that does sound useful!

I'm wondering if this would be a good fit for a plugin (see #1258 and #1279), mainly because I would like to see how that would work and because it seems to be rather well suited (I would guess this could be achieved with overriding `Event.description`). If it works well and other people find it useful as well, we could then integrate it into khal.

Would this be something you would be interested in?

tcuthbert:
> Hi @tcuthbert, I agree that does sound useful!
> 
> I'm wondering if this would be a good fit for a plugin (see #1258 and #1279), mainly because I would like to see how that would work and because it seems to be rather well suited (I would guess this could be achieved with overriding `Event.description`). If it works well and other people find it useful as well, we could then integrate it into khal.
> 
> Would this be something you would be interested in?

Yes! I'll take a look at plugins over the next few weeks.

PR #1293, opened 2023-09-03
Remove mime association from khal.desktop

xeruf:
ikhal cannot open calendar files natively and just errors out
 
 Originally reported at https://bugs.archlinux.org/task/79538


d7415 (approved):
> ikhal cannot open calendar files natively and just errors out

I'd say this is phrased poorly, but I agree with the change.

PR #1294, opened 2023-09-04
Add url-separator

juw:
Similar to description-separator add an url-separator.


tgy:
Thanks! Do you mind adding some tests where this is actually used?

juw:
Hi,

Am Sat, 16 Sep 2023 11:35:00 -0700
schrieb Valentin Iovene ***@***.***>:

> Thanks! Do you mind adding some tests where this is actually used?
>

yes, I do, it is one simple one-line if clause. Either a URL is present
and it makes sense to have the possibility to split it of from the rest
of the fields or there is none and hence no need to have a separator
printed, as is done with the 'description' field, pretty straight
forward.


Cheers.

--
 ()  ascii ribbon campaign - against html e-mail
 /\                        - against proprietary attachments

geier:
Thanks! Looks sensible. Please add yourself to AUTHORS.txt

PR #1295, opened 2023-09-16
[ikhal] add option to disable mouse, fixes #1289

### LINKED ISSUE #1289: [feature request] allow disabling mouse input
from the `man` page:

>    interactive
> invokes  the  interactive version of khal, can also be invoked by calling ikhal. While ikhal can be used entirely with the keyboard, some elements respond if clicked on with a mouse (mostly by being selected).

I don't use the mouse with any application which can be used entirely with the keyboard. For instance `htop` and `fzf` have an option (a flag) to disable mouse support.

```
fzf --no-mouse
htop --no-mouse
```

tgy:
This is done by passing the `handle_mouse` Boolean argument to urwid's main loop (currently unused). By default, the mouse remains enabled (see `khal.spec`). Mouse can be disabled either by setting `enable_mouse = False` in the `[default]` section of khal's config file, or by passing the `--no-mouse` flag to the `khal interactive` or `ikhal` click commands.

fixes #1289

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ env/
 venv/
 .hypothesis/
 .python-version
+.dmypy.json
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -97,7 +97,7 @@ def mouse_option(f):
     o = click.option(
         '--mouse/--no-mouse',
         is_flag=True,
-        default=True,
+        default=None,
         help='Disable mouse in interactive UI'
     )
     return o(f)
@@ -494,7 +494,8 @@ def _get_cli():
     @click.pass_context
     def interactive(ctx, include_calendar, exclude_calendar, mouse):
         '''Interactive UI. Also launchable via `ikhal`.'''
-        ctx.obj['conf']['default']['enable_mouse'] = mouse
+        if mouse is not None:
+            ctx.obj['conf']['default']['enable_mouse'] = mouse
         controllers.interactive(
             build_collection(
                 ctx.obj['conf'],
@@ -511,7 +512,8 @@ def _get_cli():
     def interactive_cli(ctx, config, include_calendar, exclude_calendar, mouse):
         '''Interactive UI. Also launchable via `khal interactive`.'''
         prepare_context(ctx, config)
-        ctx.obj['conf']['default']['enable_mouse'] = mouse
+        if mouse is not None:
+            ctx.obj['conf']['default']['enable_mouse'] = mouse
         controllers.interactive(
             build_collection(
                 ctx.obj['conf'],
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -94,7 +94,7 @@ def multi_calendar_option(f):
 
 
 def no_mouse_option(f):
-    o = click.option('--no-mouse', is_flag=True,
+    o = click.option('--mouse/--no-mouse', is_flag=True, default=True
                      help='Disable mouse in interactive UI')
     return o(f)
 
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -93,9 +93,13 @@ def multi_calendar_option(f):
     return d(a(f))
 
 
-def no_mouse_option(f):
-    o = click.option('--mouse/--no-mouse', is_flag=True, default=True
-                     help='Disable mouse in interactive UI')
+def mouse_option(f):
+    o = click.option(
+        '--mouse/--no-mouse',
+        is_flag=True,
+        default=True,
+        help='Disable mouse in interactive UI'
+    )
     return o(f)
 
 
@@ -486,12 +490,11 @@ def _get_cli():
 
     @cli.command()
     @multi_calendar_option
-    @no_mouse_option
+    @mouse_option
     @click.pass_context
-    def interactive(ctx, include_calendar, exclude_calendar, no_mouse):
+    def interactive(ctx, include_calendar, exclude_calendar, mouse):
         '''Interactive UI. Also launchable via `ikhal`.'''
-        if no_mouse:
-            ctx.obj['conf']['default']['enable_mouse'] = False
+        ctx.obj['conf']['default']['enable_mouse'] = mouse
         controllers.interactive(
             build_collection(
                 ctx.obj['conf'],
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -19,3 +19,4 @@ env/
 venv/
 .hypothesis/
 .python-version
+.dmypy.json
```
tgy, line 22: I've added this because I'm using neovim with pylsp-mypy (with the dmypy daemon, which can create this file).

khal/cli.py
```diff
@@ -93,6 +93,12 @@ def multi_calendar_option(f):
     return d(a(f))
 
 
+def no_mouse_option(f):
```
tgy, line 96: Since I had to use this both for `ikhal` and `khal interactive`, I figured it made sense to abstract this option into its own function decorator.

khal/cli.py
```diff
@@ -480,9 +486,12 @@ def isatty(_file):
 
     @cli.command()
     @multi_calendar_option
+    @no_mouse_option
     @click.pass_context
-    def interactive(ctx, include_calendar, exclude_calendar):
+    def interactive(ctx, include_calendar, exclude_calendar, no_mouse):
         '''Interactive UI. Also launchable via `ikhal`.'''
+        if no_mouse:
+            ctx.obj['conf']['default']['enable_mouse'] = False
```
tgy, line 494: I'm overriding the config value when the `--no-mouse` flag is used. Not sure this is the best way, but seems to work.

khal/ui/__init__.py
```diff
@@ -1344,7 +1344,12 @@ def emit(self, record):
     palette = _add_calendar_colors(
         getattr(colors, pane._conf['view']['theme']), pane.collection)
     loop = urwid.MainLoop(
-        frame, palette, unhandled_input=frame.on_key_press, pop_ups=True)
+        widget=frame,
+        palette=palette,
```
tgy, line 1348: I took the liberty to use kwargs instead of args, to make the urwid API call clearer

.gitignore
WhyNotHugo: No objections on this.

Note that you can include this in `~/.config/git/ignore` and git will ignore the file for all your local repositories.

WhyNotHugo (approved):
Looks okay to me.

.gitignore
tgy: Ah thanks, did not know that. Good to know :)

tgy:
@WhyNotHugo i fixed the tests, can you trigger the CI build (sorry was busy with work and forgot this PR)

khal/cli.py
```diff
@@ -93,6 +93,12 @@ def multi_calendar_option(f):
     return d(a(f))
 
 
+def no_mouse_option(f):
+    o = click.option('--no-mouse', is_flag=True,
```
WhyNotHugo, line 97: You can use a flag that accepts both positive and negative here:

```suggestion
    o = click.option('--mouse/--no-mouse', is_flag=True, default=True
```

This changes the name of this value to `mouse` and avoids a double negation when handling it below. E.g.: `--mouse` is `True` and `--no-mouse` is `False`.

tgy: nice! learning something new everyday :)

khal/cli.py
```diff
@@ -93,6 +93,12 @@ def multi_calendar_option(f):
     return d(a(f))
 
 
+def no_mouse_option(f):
+    o = click.option('--mouse/--no-mouse', is_flag=True, default=True
```
WhyNotHugo, line 97: Missing comma after `default=True`. Doesn't run.

khal/cli.py
```diff
@@ -480,9 +486,12 @@ def isatty(_file):
 
     @cli.command()
     @multi_calendar_option
+    @no_mouse_option
     @click.pass_context
-    def interactive(ctx, include_calendar, exclude_calendar):
+    def interactive(ctx, include_calendar, exclude_calendar, no_mouse):
```
WhyNotHugo, line 491: I believe this flag would now be `mouse`.

khal/cli.py
```diff
@@ -494,10 +503,13 @@ def interactive(ctx, include_calendar, exclude_calendar):
     @click.command()
     @global_options
     @multi_calendar_option
+    @no_mouse_option
     @click.pass_context
-    def interactive_cli(ctx, config, include_calendar, exclude_calendar):
+    def interactive_cli(ctx, config, include_calendar, exclude_calendar, no_mouse):
```
WhyNotHugo, line 508: Ditto

WhyNotHugo (changes requested):

tgy:
@WhyNotHugo addressed your comments (sorry for not running the tests before), ran the tests locally and everything passes. i think you can trigger the CI again!

khal/cli.py
```diff
@@ -93,6 +93,16 @@ def multi_calendar_option(f):
     return d(a(f))
 
 
+def mouse_option(f):
+    o = click.option(
+        '--mouse/--no-mouse',
+        is_flag=True,
+        default=True,
```
geier, line 100: I believe we need to set the default to `None` and then check explicity in `interactive_cli()` and `interactive()` if `mouse` is set to something other than `None` and only override ctx if that is the case.

khal/cli.py
```diff
@@ -494,10 +506,12 @@ def interactive(ctx, include_calendar, exclude_calendar):
     @click.command()
     @global_options
     @multi_calendar_option
+    @mouse_option
     @click.pass_context
-    def interactive_cli(ctx, config, include_calendar, exclude_calendar):
+    def interactive_cli(ctx, config, include_calendar, exclude_calendar, mouse):
         '''Interactive UI. Also launchable via `khal interactive`.'''
         prepare_context(ctx, config)
+        ctx.obj['conf']['default']['enable_mouse'] = mouse
```
geier, line 514: Something like
```python
if mouse is not None:
        ctx.obj['conf']['default']['enable_mouse'] = mouse
```

geier (changes requested):
Thanks for the PR, looks good. But I believe setting the default of the click flag to True messes with overriding the config variable. See comments.

tgy:
@geier done! thanks for catching that

geier:
@tgy I don't see any changes, did you push?

geier:
Works, thanks!

Issue #1300, opened 2023-10-14
Khal generates empty CATEGORIES: property in ics files


thibaut-ient:
**Describe the bug**
When editing an existing event, or when creating it in ikhal, the resulting .ics file contains an empty "CATEGORIES:" line.
This does not happen when using `khal new`.
This prevent the incriminated events to get synchronized to some servers using vdirsyncer

**To Reproduce**
Simply create or edit an event in ikhal and observe a "CATEGORIES:" line in the resulting file.

**Expected behavior**
ikhal shoudn't write out that line if there is no categories information?

**OS, version, khal version and how you installed it:**
khal, version 0.11.2
from Arch OS repos
Python 3.11.3
OS Archlinux

WhyNotHugo:
As I mentioned on IRC: I did a quick read of the icalendar RFC, and it seems that CATEGORIES: must have at least one category; an empty label doesn’t seem to be valid.

I guess that servers could be lenient and allow this, but strictly speaking, the ics file is not valid.

WhyNotHugo:
I thought this sounded familiar, we already had this issue in the past: https://github.com/pimutils/khal/issues/453

titibandit:
Is there anything that speaks against merging that pull request?

PR #1301, opened 2023-10-18
doesn't add empty CATEGORIES if field empty in interactive mode

titibandit:
fixes issue #1300

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -524,6 +524,7 @@ class Event:
 
     def update_categories(self, categories: List[str]) -> None:
         assert isinstance(categories, list)
+        categories = [c.strip() for c in categories if c != ""]
         self._vevents[self.ref].pop('CATEGORIES', False)
         if categories:
             self._vevents[self.ref].add('CATEGORIES', categories)


khal/khalendar/event.py
```diff
@@ -524,8 +524,9 @@ def categories(self) -> str:
 
     def update_categories(self, categories: List[str]) -> None:
         assert isinstance(categories, list)
+        categories = [c.strip().lower() for c in categories if c != ""]
```
WhyNotHugo, line 527: `.lower()` is not correct here; categories may be uppercase or mixed case.

khal/khalendar/event.py
```diff
@@ -524,8 +524,9 @@ def categories(self) -> str:
...
         self._vevents[self.ref].pop('CATEGORIES', False)
-        if categories:
+        if len(categories) > 0:
```
WhyNotHugo, line 529: Does this make any difference?

WhyNotHugo (changes requested):

khal/khalendar/event.py
titibandit: That's right. I removed it.

khal/khalendar/event.py
titibandit: It doesn't :) changed accordingly

WhyNotHugo:
I tested this with the following sequence (with `ikhal`):

- Edit an event and add a category to it.
- Check the file (it has `CATEGORIES:test`).
- Edit the event again and remove the category.
- `CATEGORIES:` is gone.

WhyNotHugo (approved):
LGTM

PR #1302, opened 2023-10-24
ReadTheDocs now requires a build.os config key

geier:


geier:
Seems to work: https://khal--1302.org.readthedocs.build/en/1302/

PR #1305, opened 2023-10-27
Feature/permissive delta parsing

### LINKED ISSUE #1285: Permissive parsing of DELTA
I often enter the DELTA using `khal add` with a leading plus, e.g. because it's how stuff works in TaskWarrior, among others. So: `khal add tomorrow 10:00 +90m Test`, but this generates an event with title "+90m Test".

My suggestion would be to swallow a leading `+` in what might be a DELTA value, and try both.

Happy to try to submit a patch.

geier:
fix #1285


WhyNotHugo (approved):
LGTM

PR #1309, opened 2023-10-28
[WIP] support plugins that implement new commands

geier:
This is rebased onto color theme plugins and the json import branch.

Initial support for plugins that implement commands.

Example (put this file at `~/.local/share/khal/plugins/overview/__init__.py`):

```python
from khal.api import register_command
from khal.cli import multi_calendar_option

import click

@click.command('overview', short_help='Show an overview of the calendar')
@multi_calendar_option
def overview(include_calendar, exclude_calendar, args=None):
    print('hello from overview')
    print('Got the following options:')
    print(include_calendar)
    print(exclude_calendar)
    print(args)



register_command('overview', overview)
```

Major issue: you cannot import `khal.cli` in your command plugin, that would lead to a circular import. We might want to restructure a bit.

Supporting command plugins would make it easy to ship new commands to users without including them directly into khal (which we could always decide to do if they are matured and there are people who maintain them). An example of this would be https://github.com/pimutils/khal/pull/1211.

Note to self:
documentation/inspiration: https://click.palletsprojects.com/en/8.1.x/complex/

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -54,3 +54,4 @@ Raúl Medina - raulmgcontact [at] gmail (dot] com
 Matthew Rademaker - matthew.rademaker [at] gmail [dot] com
 Valentin Iovene - val [at] too [dot] gy
 Julian Wollrath
+Mattori Birnbaum - me [at] mattori [dot] com - https://mattori.com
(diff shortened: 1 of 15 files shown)


geier:
Here is a longer example (the converted command from #1211):

PR #1317, opened 2023-11-03
run tests with python 3.12

geier:


geier:
blocked by https://github.com/aio-libs/aiohttp/issues/7646

tox.ini
```diff
@@ -15,15 +15,10 @@ passenv =
...
-deps =
-    pytest
-    freezegun
-    !py36: vdirsyncer
-    py36: vdirsyncer==0.16.8
-    python-dateutil
+extras =
+    tests
     pytz20187: pytz==2018.7
     pytz_latest: pytz
```
WhyNotHugo, line 21: These should remain in `deps =`:

```suggestion
deps =
    pytz20187: pytz==2018.7
    pytz_latest: pytz
```

tox.ini
```diff
@@ -15,15 +15,10 @@ passenv =
   TRAVIS_JOB_NUMBER
   TRAVIS_PULL_REQUEST
   TRAVIS_REPO_SLUG
-deps =
-    pytest
```
WhyNotHugo, line 19: `pytest` is missing from `project.optional-dependencies.test`. This is why CI is failing.

tox.ini
```diff
@@ -15,15 +15,10 @@ passenv =
...
-    freezegun
-    !py36: vdirsyncer
-    py36: vdirsyncer==0.16.8
-    python-dateutil
```
WhyNotHugo, line 23: This one is also missing, but is not required because it is a non-test dependency already.

WhyNotHugo (changes requested):

tox.ini
WhyNotHugo: (nothing to do on this item, just FYI)

tox.ini
geier: Thanks. It looks like we still need to define all dependencies in tox.ini or make it run pip install independently. Not sure, which way is worse. For now I optioned for using commands_pre to run pip install .[test].

WhyNotHugo: Here you're overriding dependencies, so they need to be in the `deps =` array (e.g.: these two are not extras).

Fixing this tests work.

PR #1328, opened 2024-01-13
make ikhal work with urwid 2.4.2

geier:
There seem to be no issues with older versions of urwid.


geier:
I'm not really sure what the reason for doing a +1 here were originally, doesn't seem to have any effect nowadays.

WhyNotHugo (approved):

PR #1333, opened 2024-01-28
In case alpha value exists remove it for urwid.

kraasch:
I don't know what is in the iCal spec, but I expect an extra check for if the color includes the alpha and then removing it cannot be harmful. 

As far as I know, there is clients like SOGO that set and alpha value and this would crash urwid.


kraasch:
In some cases it might makes sense to not demand the `addresses` field be configured, because khal can work just fine without it as for now.

geier:
Looks sensible, Thanks!

Issue #1274, opened 2023-06-18
Registering an event with [Alt][Enter] on a Date or Time widget does not save the date or time


pdav:
**Describe the bug**
When pressing [Alt][Enter] in a date-time widget, date or time is not updated in event.

**If applicable: Stack Trace**
N/A

**To Reproduce**
Using ikhal, add an event, then modify a date (start or end) or a time (start or end) without leaving the widget, then use [Alt][Enter] to register the event.

**Expected behavior**
Event should include the modified date or time.

**Screenshots**
N/A

**OS, version, khal version and how you installed it:**
 - The output of khal --version: khal, version 0.11.3.dev1+gb8a987a
 - Installation method : git + pip
 - python version python 3.11
 - OS : linux x86_64
 - 
**Additional context**
N/A

Issue #1345, opened 2024-04-20
Ikhal doesn't save changes in start or end times if you exit the widget with Page Down


Pi2048:
**Describe the bug**
If I open an event for editing in ikhal, and I change the start or end time, this change is sometimes not saved. This occurs if I exit the widget for the start or end time by using Page Down instead of the arrow keys.

**If applicable: Stack Trace**
n/a

**To Reproduce**

These are the steps for the From: time, but the bug also applies to the To: time.

* Start ikhal
* Move the cursor to an existing event with a start and end time (i.e. not an all-day event) and press enter to view details.
* Press enter again to open the event editor
* Use the arrow keys to navigate to the From: time field.
* Change its value slightly, e.g. change 18:10 to 18:12
* Navigate away from the field, but use Page Down instead of the arrow keys. The cursor ends up on the Export button.
* Press the up arrow key once to land on the Save button.
* Press enter to press the Save button.

**Expected behavior**
I expected the change I made to the From: time to be saved and displayed in the calendar overview. Instead, the change I made is silently dropped.

**Screenshots**

**OS, version, khal version and how you installed it:**
 - The output of khal --version: `khal, version 0.11.3`
 - Installation method: PyPi
 - python version: Python 3.10
 - OS: Ubuntu 22.04
 - Your khal config file: I use a minimal config file with two calendars. The only fields that are relevant to this bug are:
 ```
[locale]

timeformat="%H:%M"
```
 - The versions of your other python packages [e.g. the output of `pip freeze`]
<details>
<summary>Output of `pip freeze`</summary>
androguard==3.4.0a1
appdirs==1.4.4
apsw==3.36.0.post1
apturl==0.5.2
arandr==0.1.10
archive-hocr-tools==1.1.50
archive-pdf-tools==1.5.4
argcomplete==1.8.1
asn1crypto==1.4.0
atomicwrites==1.4.0
attrs==21.2.0
autobahn==21.11.1
Automat==20.2.0
backcall==0.2.0
base58==1.0.3
bcrypt==3.2.0
beautifulsoup4==4.10.0
beniget==0.4.1
binwalk==2.3.3
blinker==1.4
Brlapi==0.8.3
Brotli==1.0.9
cbor==1.0.0
certifi==2020.6.20
cffi==1.15.0
chardet==4.0.0
chrome-gnome-shell==0.0.0
click==8.0.3
click-log==0.2.1
click-repl==0.3.0
click-threading==0.5.0
colorama==0.4.4
[...]

PR #1346, opened 2024-04-21
Update the date and time when pressing Alt+Enter or Page Up/Down in edit widget

Pi2048:
This PR aims to fix issues #1274 and #1345

The keypress methods in DateTimeWidget and ValidatedEdit did not account for leaving the field by pressing page up/down or meta+enter. I've added these.

Also, the keypress method of EventEditor did not propagate a keypress of 'meta enter' to its children (i.e. the time and date fields). Therefore, they were not able to update the event before the EventEditor closed. I've changed the order of the calls. I'm not 100% confident about my understanding of urwid keypress hierarchy, so I've made the conservative choice to save the return value of `super().keypress(size, key)` and return that after saving (i.e. where it originally was).

The reference to 'meta enter' should probably be replaced by referring to the proper keybinding, but I'm not entirely sure how those work.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -55,3 +55,4 @@ Matthew Rademaker - matthew.rademaker [at] gmail [dot] com
 Valentin Iovene - val [at] too [dot] gy
 Julian Wollrath
 Mattori Birnbaum - me [at] mattori [dot] com - https://mattori.com
+Pi R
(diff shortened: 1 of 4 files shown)


geier:
@Pi2048 Thank you for your contribution! I can confirm that this works.

Would you please
* add an entry to the CHANGELOG (in the project root)
* add yourself to the AUTHORS.txt file (also in the project root)

Pi2048:
I've added the requested information.

I'm confused about the CHANGELOG file, though. It says version 0.11.3 is not released yet, but that version is actually listed under Releases, and it's also the one I installed through pip.

I think perhaps there needs to be a new listing under there for all changes since 0.11.3?

geier:
@Pi2048 you are totally right, somebody (me!?!) fucked up doing the last release. I'll merge this now anyway and sort that out afterwards.

PR #1362, opened 2024-10-08
Fix: Use date.replace instead of localize

rootandroo:
Fixed an issue with timezone handling in the `sanitize_datetime` function. It replaces the use of the `localize` method (specific to `pytz`) with the `replace` method, which is compatible with both `pytz` and `zoneinfo`.


WhyNotHugo:
This is just syntax changes, I don't see any changes in logic anywhere.

rootandroo:
Oh my editor automatically changed the syntax I'll resubmit with only the change

WhyNotHugo:
No worries, it happens.
I don’t think we have any automated solution to fix that either.

PR #1363, opened 2024-10-12
Replace tz.localize with date.replace

rootandroo:
Resubmitted PR without syntax changes https://github.com/pimutils/khal/pull/1362

### DIFF
--- a/khal/icalendar.py
+++ b/khal/icalendar.py
@@ -246,7 +246,7 @@ def expand(
         if allday and isinstance(date, dt.datetime):
             date = date.date()
         if events_tz is not None:
-            date = events_tz.localize(date)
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Tests don't pass; we still rely on other pytz-specific attributes.

iamleot:
I have tried to address that in #1364.

`icalendar` exposes a `use_pytz()` and I think it's safer and less intrusive. When running the tests (and also when running for real `khal`!) I have seen several problems due unavailability of `icalendar.windows_to_olson` and I've fixed that too.

All the tests now passes for me with #1364.

I think that PR can obsolete/close this one if it looks good to you!

Thank you!

PR #1364, opened 2024-10-31
Force usage of pytz

iamleot:
For timezone functionalities `icalendar` can use both `pytz` and `zoneinfo`, preferring `zoneinfo`.

We rely on `pytz` for several attributes, stick to it.

Fixes #1361.
Should obsolete #1363.

PR #1365, opened 2024-10-31
Replace intersphinx_mapping format removed in Sphinx 8.0.0

### LINKED ISSUE #1356: Build problem with sphinx 8
Building khal 0.11.3 with sphinx 8.0.2 fails with:
```
Running Sphinx v8.0.2
loading translations [en]... done
making output directory... done
Converting `source_suffix = '.rst'` to `source_suffix = {'.rst': 'restructuredtext'}`.
ERROR: Invalid value `None` in intersphinx_mapping['http://docs.python.org/']. Expected a two-element tuple or list.

Configuration error:
Invalid `intersphinx_mapping` configuration (1 error).
*** Error code 2
```

The following patch works around this, but I'm not sure if it's correct or if we lose some links this way:
```
--- doc/source/conf.py.orig     2024-08-06 17:18:55.985963360 +0000
+++ doc/source/conf.py
@@ -309,4 +309,4 @@ texinfo_documents = [


 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'http://docs.python.org/': None}
+#intersphinx_mapping = {'http://docs.python.org/': None}
```

d7415:
Fixes #1356

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -309,4 +309,4 @@ texinfo_documents = [
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'http://docs.python.org/': None}
+intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}


d7415:
The new format has been supported since v1.0 in 2010.

d7415:
Branch protection means I can't merge this until we fix the tests, possibly with #1364.

mathstuf:
Fedora Rawhide is looking to update to Sphinx 8…any progress here? I see concerns about `icalendar` 6 being "too new"…but we're getting to the point that distros are going to start requiring the non-deprecated behavior too…

mathstuf:
This has now become an FTBFS in Fedora 42: https://bugzilla.redhat.com/show_bug.cgi?id=2340695

antonmosich:
Is intersphinx even used anywhere? Is there a reason to not outright remove that line, as well as the import of the extension?

PR #1374, opened 2024-12-01
fix 1367: :khal crashes with zoneinfo attrib error adding new event

syasma:
1. Managing compatibility between pytz and zoneinfo.ZoneInfo
The original code was designed to work only with pytz.BaseTzInfo objects, which caused problems when zoneinfo.ZoneInfo objects (introduced with Python 3.9) were used. The solution was to dynamically check which type of timezone object was being used (pytz or zoneinfo), and treat each case differently.

Key changes :
Check timezone type:
If the timezone is a pytz instance, the code applies the methods adapted to pytz to extract time transition information (e.g.: tz._utc_transition_times).
If the timezone is an instance of zoneinfo.ZoneInfo, the code must handle the retrieval of transition information differently, as zoneinfo does not contain the same internal attributes (such as _tzinfos and _utc_transition_times).

2. Adapting the create_timezone function
The create_timezone function has been refactored to accept both pytz and zoneinfo.ZoneInfo objects as arguments and to generate a VTIMEZONE component based on the transition information of both timezone types.

Key changes :
zoneinfo.ZoneInfo support: logic that depended on pytz internal attributes (such as _tzinfos and _utc_transition_times) has been replaced by zoneinfo-compatible management. For example, for zoneinfo, the _transitions() method can be used to obtain time transitions, although this is not directly supported as with pytz.
VTIMEZONE generation for both timezone types: Whether the timezone is of the pytz or zoneinfo type, a VTIMEZONE component is created, including the necessary information such as TZNAME, DTSTART, and offsets (start and end) for each time transition.

3. Dynamic start_date management
For each timezone, a VTIMEZONE must include time transitions relating to a specific period of time (often associated with the first and last date of an event). The code has been adjusted to dynamically retrieve the event start_date if available. This date is then used to correctly generate the relevant time transitions.

Key change:
Dynamic management of start_date: If self.start (the event start date) is defined, it is used to calculate the period during which time transitions should be generated. Otherwise, the current date is used as the default value.

4. Refactoring the raw method
The raw method generates the complete VCALENDAR file, including the VTIMEZONE components for each event. It has been updated to include timezone checks (pytz and zoneinfo), as well as adding VTIMEZONE components only if the timezone is not UTC.

Key change:
Timezone collection: The method scans events to collect all unique timezones (those associated with DTSTART and DTEND). It then adds the corresponding VTIMEZONE components to the calendar, by calling the create_timezone function.
Adding events (VEVENT): After adding the VTIMEZONE components, the events themselves are added to the calendar.

### DIFF
--- a/khal/khalendar/event.py
+++ b/khal/khalendar/event.py
@@ -40,6 +40,7 @@ from ..icalendar import cal_from_ics, delete_instance, invalid_timezone
 from ..parse_datetime import timedelta2str
 from ..plugins import FORMATTERS
 from ..utils import generate_random_uid, is_aware, to_naive_utc, to_unix_time
+from zoneinfo import ZoneInfo
 
 logger = logging.getLogger('khal')
 
@@ -377,29 +378,43 @@ class Event:
         )
         return calendar
 
+
     @property
     def raw(self) -> str:
-        """Creates a VCALENDAR containing VTIMEZONEs
-        """
+        """Creates a VCALENDAR containing VTIMEZONEs."""
         calendar = self._create_calendar()
         tzs = []
+
+        # Collect timezones from the events
         for vevent in self._vevents.values():
+            # Handle DTSTART timezone
             if hasattr(vevent['DTSTART'].dt, 'tzinfo') and vevent['DTSTART'].dt.tzinfo is not None:
                 tzs.append(vevent['DTSTART'].dt.tzinfo)
-            if 'DTEND' in vevent and hasattr(vevent['DTEND'].dt, 'tzinfo') and \
-                    vevent['DTEND'].dt.tzinfo is not None and \
-                    vevent['DTEND'].dt.tzinfo not in tzs:
+            # Handle DTEND timezone
+            if (
+                'DTEND' in vevent
+                and hasattr(vevent['DTEND'].dt, 'tzinfo')
+                and vevent['DTEND'].dt.tzinfo is not None
+                and vevent['DTEND'].dt.tzinfo not in tzs
+            ):
                 tzs.append(vevent['DTEND'].dt.tzinfo)
 
+        # Add VTIMEZONE components for collected timezones
         for tzinfo in tzs:
-            if tzinfo == pytz.UTC:
-                continue
-            timezone = create_timezone(tzinfo, self.start)
+            if tzinfo in [pytz.UTC, ZoneInfo("UTC")]:
+                continue  # Skip UTC as it doesn't need a VTIMEZONE
+            # Determine start date for timezone creation
+            start_date = getattr(self, 'start', None)
+            timezone = create_timezone(tzinfo, start_date)  # Ensure create_timezone supports both pytz and ZoneInfo
             calendar.add_component(timezone)
 
+        # Add VEVENT components to the calendar
         for vevent in self._vevents.values():
             calendar.add_component(vevent)
+
+        # Return the VCALENDAR as a string
         return calendar.to_ical().decode('utf-8')
+
 
     def export_ics(self, path: str) -> None:
         """export event as ICS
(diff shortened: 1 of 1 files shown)


khal/khalendar/event.py
```diff
@@ -377,30 +377,44 @@ def _create_calendar() -> icalendar.Calendar:
...
         return calendar
 
+
     @property
     def raw(self) -> str:
-        """Creates a VCALENDAR containing VTIMEZONEs
-        """
+        """Creates a VCALENDAR containing VTIMEZONEs."""
         calendar = self._create_calendar()
         tzs = []
```
WhyNotHugo, line 385: **non-blocking nitpick**: I would use a set here, so we can just insert new entries without having any duplicates (inserting an existing element into a set is a no-op):

```suggestion
        tzs = {}
```

khal/khalendar/event.py
```diff
@@ -919,110 +933,90 @@ def duration(self) -> dt.timedelta:
 
 
 def create_timezone(
-    tz: pytz.BaseTzInfo,
-    first_date: Optional[dt.datetime]=None,
-    last_date: Optional[dt.datetime]=None
+    tz,
```
WhyNotHugo, line 936: Please leave type hints here:

```suggestion
    tz: pytz.tzinfo | zoneinfo.ZoneInfo,
```

khal/khalendar/event.py
```diff
@@ -919,110 +933,90 @@ def duration(self) -> dt.timedelta:
...
-
+    Create an icalendar vtimezone from a pytz.tzinfo or zoneinfo.ZoneInfo object.
     """
-    if isinstance(tz, StaticTzInfo):
+    if isinstance(tz, pytz.tzinfo.StaticTzInfo):
         return _create_timezone_static(tz)
 
-    # TODO last_date = None, recurring to infinity
+    first_date = dt.datetime.today() if not first_date else first_date
+    last_date = first_date + dt.timedelta(days=1) if not last_date else last_date
```
WhyNotHugo, line 947: I'm not entirely familiar with this code. Why did you remove `to_naive_utc` here?

khal/khalendar/event.py
```diff
@@ -919,110 +933,90 @@ def duration(self) -> dt.timedelta:
...
-        subcomp.add('TZOFFSETFROM', tz._transition_info[num - 1][0])  # type: ignore
-        timezones[name] = subcomp
+    if isinstance(tz, ZoneInfo):
+        # Handle ZoneInfo transitions manually since it doesn't have _transitions
+        transitions = _get_zoneinfo_transitions(tz, first_date, last_date)
 
-    for subcomp in timezones.values():
+    elif isinstance(tz, pytz.BaseTzInfo):
+        # Use pytz's transitions
+        transitions = tz._utc_transition_times  # Avoid using internal attributes directly.
```
WhyNotHugo, line 961:
```suggestion
        transitions = tz._utc_transition_times  # TODO: Avoid using internal attributes directly.
```

WhyNotHugo (commented):
I created an event without this patch and with this patch. The serialised timezones are not the same.

Without this patch:

```icalendar

BEGIN:VTIMEZONE
TZID:Europe/Amsterdam
BEGIN:STANDARD
DTSTART:20241027T020000
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:20250330T030000
TZNAME:CEST
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
END:DAYLIGHT
END:VTIMEZONE
```

With this patch:

```icalendar
BEGIN:VTIMEZONE
TZID:Europe/Amsterdam
BEGIN:STANDARD
DTSTART;TZID=Europe/Amsterdam:20241203T100000
TZNAME:1:00:00
TZOFFSETFROM:+0100
TZOFFSETTO:+0100
END:STANDARD
END:VTIMEZONE
```


diff:

```diff
@@ -1,9 +1,15 @@
 BEGIN:VTIMEZONE
 TZID:Europe/Amsterdam
 BEGIN:STANDARD
-DTSTART;TZID=Europe/Amsterdam:20241203T100000
-TZNAME:1:00:00
-TZOFFSETFROM:+0100
+DTSTART:20241027T020000
+TZNAME:CET
+TZOFFSETFROM:+0200
 TZOFFSETTO:+0100
 END:STANDARD
+BEGIN:DAYLIGHT
+DTSTART:20250330T030000
+TZNAME:CEST
+TZOFFSETFROM:+0100
+TZOFFSETTO:+0200
+END:DAYLIGHT
 END:VTIMEZONE
```

From the definition for `VTIMEZONE` in  https://www.rfc-editor.org/rfc/rfc5545#section-3.6.5:

> The mandatory "DTSTART" property gives the effective onset date
> and local time for the time zone sub-component definition.
> "DTSTART" in this usage MUST be specified as a date with a local
> time value.

If I understand correctly, the new output seems to be valid. It's accepted and rendered properly with Fastmail, no idea about other clients.

The `TZNAME`s are lost, but I'm not sure how important they are of if any client actually renders these anyway.

Issue #1388, opened 2025-01-26
Compability with Python 3.13


cl-parsons:
Adapt the code to be able to run on Python 3.13

WhyNotHugo:
Context? What need to be adapted?

cl-parsons:
In fact only the pyproject.toml was to be changed.

WhyNotHugo:
Please explain what is failing for you.

cl-parsons:
I think the issue is useless... But I did a pull request that should be enough.

WhyNotHugo:
Sorry, didn't see your PR. Will look at https://github.com/pimutils/khal/pull/1389 now.

Issue #1361, opened 2024-10-01
Support icalendar v6


dotlambda:
**Describe the bug**
Icalendar v6 was released: https://github.com/collective/icalendar/blob/main/README.rst#version-6-with-zoneinfo
Now running pytest yields test failures.

**If applicable: Stack Trace**

```
============================= test session starts ==============================
platform linux -- Python 3.12.5, pytest-8.3.2, pluggy-1.5.0
rootdir: /build/source
configfile: pyproject.toml
plugins: hypothesis-6.108.5
collected 315 items / 5 deselected / 310 selected                              

tests/backend_test.py .F.FF...FFFF....F.............                     [  9%]
tests/cal_display_test.py .........                                      [ 12%]
tests/cli_test.py .FFFF...F.xX.x..FF....F...........F.F.FF.              [ 25%]
tests/configwizard_test.py ..                                            [ 26%]
tests/controller_test.py .....FF.....                                    [ 30%]
tests/event_test.py ..F.FF..FFFFFFF...F..F.........F..F.......F......... [ 47%]
...                                                                      [ 48%]
tests/icalendar_test.py ..FF.F.                                          [ 50%]
tests/khalendar_test.py .......F..FFF..F..FFF.........                   [ 60%]
tests/khalendar_utils_test.py FFF....F.....F...F...FFFF.F..FF.........   [ 72%]
tests/parse_datetime_test.py ........................................... [ 86%]
.....                                                                    [ 88%]
tests/settings_test.py ............                                      [ 92%]
tests/terminal_test.py ...                                               [ 93%]
tests/ui/test_calendarwidget.py .....                                    [ 94%]
tests/ui/test_editor.py ....                                             [ 96%]
tests/ui/test_widgets.py .                                               [ 96%]
tests/utils_test.py ........                                             [ 99%]
tests/vdir_test.py .                                                     [ 99%]
tests/vtimezone_test.py ..                                               [100%]

=================================== FAILURES ===================================
...

=========================== short test summary info ============================
FAILED tests/backend_test.py::test_event_rrule_recurrence_id - AttributeError: 'zoneinfo.ZoneInfo' object has no attribute 'localize'
FAILED tests/backend_test.py::test_event_rrule_recurrence_id_reverse - AttributeError: 'zoneinfo.ZoneInfo' object has no attribute 'localize'
[...]

PR #1389, opened 2025-01-26
support python 3.13 (#1388), limit icalendar version < 6 (#1361)

cl-parsons:
Bump Python version, tested on Archlinux with Python 3.13.1 .

I had to limit the icalendar version to version 5.0.13 because of unresolved issue #1361 .

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,11 +23,11 @@ classifiers = [
   "Topic :: Communications",
   "Topic :: Utilities",
 ]
-requires-python = ">=3.8,<3.13"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -23,11 +23,11 @@ classifiers = [
   "Topic :: Communications",
```
WhyNotHugo, line 23: Should also add `3.13` as a classifier on the line above this one.

cl-parsons: I just added it

PR #1414, opened 2025-06-14
Add location-separator field

markk:
Analogous to `description-separator` and `url-separator` this can be used in a format string when the `location` information is desired.


WhyNotHugo (approved):

sotpapathe:
I was about to create a PR implementing this exact thing. Is there anything blocking it from being merged? I'd love to have this feature.

WhyNotHugo:
Am I missing something? This `location-separator` field is not used anywhere. Did GitHub screw up the rebase?

WhyNotHugo:
No, looks like the original commit only sets this field too: 7350d2145454a64b9d329704eb07bc3fd52fbd69

But I don't see it used anywhere.

sotpapathe:
I think the author's (and my) idea would be to use it in a format string passed to the `--format` option to allow further customization of the output.

In my case, it would allow changing this
``` text
--format '{title}{description-separator}{description} :: {location}{url-separator}{url}'
```
into this
``` text
--format '{title}{description-separator}{description}{location-separator}{location}{url-separator}{url}'
```

WhyNotHugo:
What's the difference between using `location-separator` in your template instead of putting just the separator itself in there?

sotpapathe:
The difference is that events without a location will have unnecessary separators and look weird with the hardcoded separator, e.g.:

``` text
# only title
Some event :: 

# title and description
Some other event :: do foo :: 

# title and URL
Another event :: :: https://example.com
```
whereas with this PR the above would look like:
``` text
Some event
Some other event :: do foo
Another event :: https://example.com
```

markk:
Bump on this. Let me know if there is a problem with merging this please.

WhyNotHugo:
Thanks

PR #1419, opened 2025-07-18
Fix default cache database location

### LINKED ISSUE #1296: default khal.db location
If I'm not mistaken, khal.db is just a cache file and as such it would make more sense to me to store it under XDG_CACHE_HOME by default instead of XDG_DATA_HOME. This is of course a minor thing and perhaps slighly subjective, on the other hand for example todoman follows this scheme. It can also be useful for instance when creating backups where the entire ~/.cache directory might be skipped as non-essential, but not so ~/.local (not that the usual size of the database would make that much of a difference, but I imagine the binary format could make it harder for deduplication algorithms to be efficient and small changes in the database could cause unnecessary amount of data to get backed up on incremental backups).

Please note that I'm aware that the path can be configured (which makes this even a more minor thing) - I'm talking about the *default* location.

The change required to make this switch should be simple, however I cannot think of a good simple way of moving the existing file to the new location, so the least evil would probably be just leaving it there for the user to delete it manually (a notice could be put in the changelog) and generating a new one. No changes would occur if the user has specified a custom path for the file in his configuration file.

Of course, if you disagree with this or don't think it's worth changing it anymore, feel free to close this :-)

Jorenar:
Fixes #1296

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -58,3 +58,4 @@ Mattori Birnbaum - me [at] mattori [dot] com - https://mattori.com
 Pi R
 Alnoman Kamil - noman [at] kamil [dot] gr - https://kamil.gr
 Leonardo Taccari - iamleot [at] gmail [dot] com
+Jorenar - dev [at] jorenar [dot] com - https://jorenar.com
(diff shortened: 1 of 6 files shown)


WhyNotHugo:
Makes sense to me :+1:

Jorenar:
> Makes sense to me 👍

@WhyNotHugo Yay! :rocket:

PR #1442, opened 2025-11-15
Improve usability of `khal list -d absent`

mgumz:
If a user provides an absent calendar name to be filtered out with `khal list -d`, an error is raised and the call stack is thrown at the user (24 lines of internals with a final "KeyError: 'absent').

To improve the usability, this improvement catches this situation and properly tells the user what happend and how to remedy the situation, just like `-a absent` would do.


WhyNotHugo:
Nice, thanks!

PR #1443, opened 2025-11-27
Add main category and icon to khal.desktop

renanglr:
This pull request adds the "Office" category and the "x-office-calendar" icon to the khal.desktop file.

* Adding a category fixes the entry being placed in the "Lost and Found" category in KDE Plasma, maybe other desktops?
* Adding an icon makes it look good in the menus :-) I've chosen an icon that is available in most icon themes. It is an icon for a file type instead of an application, but it should look good enough.

### DIFF
--- a/misc/khal.desktop
+++ b/misc/khal.desktop
@@ -1,8 +1,9 @@
 [Desktop Entry]
 Name=ikhal
-Categories=Calendar;ConsoleOnly;
+Categories=Office;Calendar;ConsoleOnly;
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
x-office-calendar is a well-known icon as per the [Icon Naming Specification](https://specifications.freedesktop.org/icon-naming/latest/) :+1:

WhyNotHugo (approved):

PR #1454, opened 2026-02-19
Add support for icalendar 7.0.0

WhyNotHugo:
- Contentlines moved to another module.
- Exception raised is now a subclass of ValueError


WhyNotHugo:
Some test fail giving wrong dates.

Issue #1195, opened 2022-10-12
Making changes to an entry without editing for use in a script


fredvos:
Hi,

I'ld like to have the option to change or add attributes to entries using commands.

Currently I can make changes using the edit command, but editing is interactive. I'm using Taskwarrior too. Taskwarrior supports the use in scripts. You can add, remove and change attributes for a task in a non-interactive manner. This is a very neat feature of Taskwarrior.

If this could be possible too in Khal, it would make my work much easier.

An example. I want to add references to notes in my file system to both calendar items in Khal and tasks in Taskwarrior. I then need at least a marker that notes are available, possibly with a reference to the name of the file, so a command like
$ khal modify [uid] --add-category "HasNotes.xyz"
could be used in a script to add a reference to a file with notes. Currently I can only add a category using the interactive edit command.

Thanks, Fred

Tonus1:
Hi
In a script, since that's all text files, doesn't sed command gets job done ? 


Le 12 octobre 2022 10:49:37 GMT+02:00, Fred Vos ***@***.***> a écrit :

Issue #603, opened 2017-03-09
Non-interactive way to delete events?


oschrenk:
I saw that there was actually an issue #49 to be able to delete events. It got closed since the new `edit` feature came out. Is there a way to make it non-interactive and have it delete all/first occurrence?

geier:
No, not at the moment and I'm not sure how that should work. What do exactly mean with non-interactive? `khal delete UID` and then get queried `delete [A]ll, [N]ext instance, or [S]pecific instance`?

hobarrera:
I think the simplest thing is something like:
```bash
khal delete 12:00  # Deletes event for today at midday
khal delete 2017-04-05 13:00
```

The problem is if you have multiple events at the same time.

FWIW, we deal with this in `todoman` by printing todos with an internal id (an auto-numeric that we assign on-cache), but I think that's an ugly UX for khal.

oschrenk:
To delete an event I currently have to

```
$ khal edit "search term"
10.03. 09:30-10.03. 11:00 Event with search term
Edit?  [n]o  [s]ummary  [d]escription  da[t]etime range  re[p]eat  [l]ocation  [c]ategories  [a]larm  [D]elete  [q]uit: D
Delete all occurences of event? [y/N]: y
```

So currently I have to be at the keyboard, press `D` and then press `y` to delete an event. For me that is interactive. A non-interactive way would be where I can issue a command and the event(s) are deleted, without waiting me to give input. The goal for me here is to automate things 

I see these options:

1. What you suggested: Having UIDs (or some other form of id) as part of search results, which in turn I can target with a delete command
2. Or having a flag/option on the `edit` command to delete results eg `khal edit --delete "search term". 
3. Having a separate command like `khal delete "search term". It could automatically delete all the results it would return and just spit out the result. But it could also offer an interactive version as default, asking for confirmation for each deletion, and then flags like `--force` for a non-interactive version

bbugyi200:
I remember seeing this issue and wanted to provide a quick work around for anyone else who wants this feature as bad as I do. If your instance of khal is using sqlite, you can just write a shell function or two with a oneliner interfacing the sqlite database. For example, the function below will delete all khal events with an event summary. Once defined, you can use `kdel EVENT` from the command-line to delete all khal events with summaries that start with `EVENT`. Be warned that this query will also match events with fields *below* summary that match `EVENT`. I'm not sure how to fix that.

[...]

PR #1462, opened 2026-03-22
Implement non-interactive edits

lukasl-dev:
Previously, khal only supported interactive edits. My commits adds the feature to specify edit flags, such as `--summary` and `--description`, to edit an event non-interactively (interesting for automated environments, such as cli, scripts, agents, etc.).

If none of the new edit flags is specified, khal runs in interactive mode as expected before my changes.

If there are multiple matches, the changes error by default. To change all matches, specify `--all`.

**Updated command description:**

```
Edit (or delete) events matching the search string.

When run without any of the edit flags (--summary, --description, etc.),
this command runs in interactive mode, allowing you to edit events one at a
time.

When run with edit flags, this command runs in non-interactive mode. If only
one event matches the search string, it is edited directly. If multiple
events match, the command will error out unless --all is used.
```

### DIFF
--- a/khal/cli.py
+++ b/khal/cli.py
@@ -493,10 +493,43 @@ def search(ctx, format, json, search_string, include_calendar, exclude_calendar)
               help=('The format of the events.'))
 @click.option('--show-past', help=('Show events that have already occurred as options'),
               is_flag=True)
+@click.option('--summary', '-s',
+              help=('Update the event summary.'))
+@click.option('--description', help=('Update the event description.'))
+@click.option('--location', '-l',
+              help=('Update the event location.'))
+@click.option('--categories', '-g',
+              help=('Update the event categories, comma separated.'))
+@click.option('--url', '-u',
+              help=('Update the event url.'))
+@click.option('--start',
+              help=('Update the event start datetime.'))
+@click.option('--end',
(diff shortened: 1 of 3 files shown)


nfvelten:
I hit this from the other side. I am writing a khal-backed calendar UI for Omarchy, and since `khal edit` is interactive-only I ended up bypassing khal entirely: walking the vdir to find the .ics whose VEVENT carries the UID, rewriting it through a temp file in the same directory, and unlinking it for delete.

It works, but it means every consumer re-derives things khal already knows. Which calendars are readonly. That the filename is conventionally the UID but nothing in the format requires it, so matching on the filename is wrong. That a non-atomic write can leave khal indexing a truncated event on its next reindex. That is a fair amount of surface for something that is one flag away in khal itself, and this PR covers both #603 and #1195.

One note from that use case: addressing by search string is the right default for someone at a shell, but a program usually already holds the UID, and search is ambiguous for it (`--all` is not the answer there, since the intent is exactly one event). Would you be open to a `--uid` selector alongside the search string, or is that better kept as a separate change on top of this one?

I am happy to test the branch against my setup if that is useful. vdirsyncer with filesystem calendars, including a couple of readonly ones.

PR #1460, opened 2026-03-15
Add timezone when assigning start/end time

mschilli87:
See/fixes https://github.com/pimutils/khal/issues/1459.

PR #1465, opened 2026-03-24
Add changelog entry for #1460

mschilli87:
See https://github.com/pimutils/khal/pull/1460#issuecomment-4119943094.


WhyNotHugo:
Thanks

PR #1467, opened 2026-05-08
fix: replace bare except with except Exception in backend.py

KeloYuan:
## Summary

Replaces bare `except:` with `except Exception:` in `khal/khalendar/backend.py` (line 101).

## Why

Bare `except:` catches `SystemExit` and `KeyboardInterrupt`, which can mask Ctrl+C handling and make debugging harder.

### DIFF
--- a/khal/khalendar/backend.py
+++ b/khal/khalendar/backend.py
@@ -98,7 +98,7 @@ class SQLiteDb:
         self._at_once = True
         try:
             yield self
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Sounds correct to me. Thanks.

KeloYuan:
> Sounds correct to me. Thanks.

You' are welcome

PR #1470, opened 2026-06-07
pyproject: support Python 3.15

mathstuf:
Fedora 45 is already starting to build with it.

---
Can be deferred to whenever it is wanted; I'm mostly making sure that it's actually OK here before blindly updating constraints in the Fedora package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14", "3.15.0-beta.2"]
         tox-test: ["default"]
 
     steps:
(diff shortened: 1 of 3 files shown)


mathstuf:
Ok, well then. Seems something is broken on 3.15 then.

rosset:
[khal-0.14.0-python3.15-strptime.patch](https://github.com/user-attachments/files/28732693/khal-0.14.0-python3.15-strptime.patch)
@mathstuf  @WhyNotHugo  could you please check if this patch address the fix for py3.15 ?

mathstuf:
Looks reasonable for Fedora. Not sure what makes sense to actually merge as a long-term fix. Testing it here in the meantime.

I do have edge-case worries about things like `%e %%yolo` in the format string being a false-positive, but…highly unlikely.

rosset:
> Looks reasonable for Fedora. Not sure what makes sense to actually merge as a long-term fix. Testing it here in the meantime.
> 
> I do have edge-case worries about things like `%e %%yolo` in the format string being a false-positive, but…highly unlikely.

thanks @mathstuf 

@WhyNotHugo  can you please review if this patch make sense to be merge to the master branch ?

For now we will use this patch to fix 0.14.0 build on Fedora rawhide (development)

thanks guys.

PR #1480, opened 2026-08-28
docs: fix broken links in installation instructions

wjstoal3959-lgtm:
Three links in the installation guide are dead. All verified with curl (old URLs return 404, new ones return 200).

| Location | Problem | Fix |
| --- | --- | --- |
| `install.rst` Arch Linux link | `[community]` repo was merged into `[extra]` in 2023; the old URL now 404s | point to `archlinux.org/packages/extra/any/khal/` |
| `install.rst` Fedora link | `admin.fedoraproject.org/pkgdb` was retired years ago | point to `packages.fedoraproject.org/pkgs/khal/khal/` |
| `install.rst` dev install command | GitHub disabled the `git://` protocol in 2022, so `pip install git+git://…` fails | use `git+https://…` |

Docs only, no behaviour change. Happy to add an `AUTHORS.txt` / `CHANGELOG.rst` entry if you'd like one for this.

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -7,10 +7,10 @@ khal has been packaged for, among others: Arch Linux (stable_ and development_
 versions), Debian_, Fedora_, FreeBSD_, Guix_, and pkgsrc_ (or see repology_ for
 a more complete list).
 
-.. _stable: https://www.archlinux.org/packages/community/any/khal/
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Thanks, looks good! Changelog entry is not necessary for this. AUTHORS entry is at your choosing.

PR #1481, opened 2026-09-02
ikhal: view event details in the external pager

### LINKED ISSUE #534: in event preview use space key to view long/overflowing event descriptions
see #413

### LINKED ISSUE #1385: ikhal scroll event information
I know this is an edge case, but didn't find info in the main page is there a way to scroll the event in ikhal ?

I am asking because of work, I  have events that have a huge list of Attendees, that literally fill the entire space, is the only approach enter the edit event mode see the info and exit it  ?

Don't  know if I made myself clear but here is a picture that shows the event literally filled with Attendees and can't see the description.

Image

MarioMoura:
The event view window clips content it cannot fit, so events with long descriptions or attendee lists are partially unreadable. Add a 'pager' keybinding (default: V) that suspends the urwid loop and shows the focused event's details via click.echo_via_pager, which respects $PAGER and falls back to less.

Notes for review: all tests pass locally (334 passed); no new test since this is in the urwid part. I considered the plugin route first, but the plugin API (commands/formatters/color themes) has no hook for ikhal key actions, so core seemed the only place for this (I may be wrong). Changelog, AUTHORS and usage docs are updated.

Closes #534
Closes #1385

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -59,3 +59,4 @@ Pi R
 Alnoman Kamil - noman [at] kamil [dot] gr - https://kamil.gr
 Leonardo Taccari - iamleot [at] gmail [dot] com
 Jorenar - dev [at] jorenar [dot] com - https://jorenar.com
+Mario Moura - mm [at] mariomoura [dot] com - https://mariomoura.com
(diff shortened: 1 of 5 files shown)


WhyNotHugo:
LGTM. We should probably wrap content in the regular view, or make it scrollable, but this is sensible regardless.

PR #1485, opened 2026-09-18
import: honor --format (#1165)

### LINKED ISSUE #1165: Feature Request: Formatter for the recurrence of a recurring meeting
I would like to be able to see what the recurrence is for a recurring meeting, particularly on import.
As the event display is controlled by a `--format` or equivalent configuration, I would propose having a `{repeat}` or `{repeat-details}` format option, even if that means it is just a dump of the "raw" RRULE as the first pass.

John-Wassilak:
Closes #1165.

Most of what was asked for seems to already exist : `{repeat-pattern}` expands to the raw RRULE, in _list_, _search_, _--json_, and on _import_.

The one gap I found was `khal import` declares `--format` but never processed the value.

Now, for example, you can run:
```
$ khal import -f '{title} {repeat-pattern}' birthday.ics
Dummy's Birthday (1965) FREQ=YEARLY
```


WhyNotHugo (approved):
LGTM

